PR #3, opened 2017-08-30
Enhancements of Vec and CircularBuffer

XOSplicer:
# Changelog
- new file structure with one file per type and pub use in lib.rs
- rustdoc and examples for public methods
- add doctests and test modules
- increased version number to 0.2.0
- implement some of the methods from std::Vec for Vec
- implement iter for CircularBuffer
- remove deref for CircularBuffer


src/circular_buffer.rs
```diff
@@ -0,0 +1,313 @@
+
+#![deny(missing_docs)]
```
japaric: nit: if you have these `#![deny]`s in `src/lib.rs`, the root of the crate, you don't need to write them here as the `#![deny]` in the root of the crate will apply to the whole crate.

src/circular_buffer.rs
```diff
@@ -0,0 +1,313 @@
...
+            _marker: PhantomData,
+            array: array,
+            index: 0,
+            len: len,
+        }
+    }
+
+    /// Returns the capacity of this buffer.
+    ///
+    ///
```
japaric: nit: extra empty line

src/circular_buffer.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Insertions will overwrite at the start of the array.
+    ///
+    /// # Example
+    /// ```
+    /// use heapless::CircularBuffer;
+    /// let buf = CircularBuffer::from_array([1,2,3]);
+    /// assert!(buf.is_full());
+    /// ```
+    #[inline]
+    pub fn from_array(array: A) -> Self {
```
japaric: nit: this can be a `const` function

src/circular_buffer.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// This is the next place to push to.
+    ///
+    /// # Example
+    /// ```
+    /// use heapless::CircularBuffer;
+    /// let buf = CircularBuffer::from_array([1,2,3]);
+    /// assert_eq!(buf.index(), 0);
+    /// ```
+    #[inline]
+    pub fn index(&self) -> usize {
```
japaric: I'm not sure if we want to expose this and the `first_index` method in the API. Do you have an use case for it?

src/circular_buffer.rs
japaric: I'm wondering if this can be named just `from` or if this could be an implementation of the `From` trait. In the latter case, this couldn't be a `const` function.

src/circular_buffer.rs
```diff
@@ -0,0 +1,313 @@
...
+        unsafe {
+            self.set_len(len - 1);
+            self.set_index(index - 1);
+            Some(self.array.as_ref()[self.index()])
+        }
+    }
+
+    /// Returns an iterator over the elements as `&T`
+    #[inline]
+    pub fn iter(&self) -> Iter<T, A> {
```
japaric: nit: I think it's better if this explicitly showed that `Iter` will freeze `Self` by explictly annotating the lifetimes: as in `fn iter<'a>(&'a self) -> Iter<'a, T, A>`.

src/vec.rs
```diff
@@ -0,0 +1,637 @@
...
+    /// at compile time.
+    ///
+    /// # Example
+    /// ```
+    /// use heapless::Vec;
+    /// let vec = unsafe{ Vec::with_len([1,2,3], 3) };
+    /// assert!(vec.is_full());
+    /// ```
+    #[inline]
+    pub const unsafe fn with_len(array: A, len: usize) -> Self {
```
japaric: This seems rather easy to screw up. Could this become safe by adding an assertion / check into the implementation?

src/vec.rs
```diff
@@ -0,0 +1,637 @@
...
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    /// Returns the number currently available backup storage.
+    ///
+    /// this is equal to `self.capacity() - self.len()`
+    #[inline]
+    pub fn available(&self) -> usize {
```
japaric: I'm not sure if want to add this utilitiy method just yet.

XOSplicer:
ok I will look into this after work

src/circular_buffer.rs
XOSplicer: ` let len = array.as_ref().len();`
can't be evaluated at compile time, I would suggest to either remove it or use 
`pub const unsafe fn with_len(array: A, len: usize)` as in headless::Vec::with_len.

src/vec.rs
XOSplicer: could be made private for now

src/circular_buffer.rs
XOSplicer: It is used in the circular_buffer::Iter::new method to initialize the iterator.
It could be made non-public.

src/vec.rs
XOSplicer: We would need some kind of compile time assertions here. I don't know if this is in the compiler yet or if there are any crates providing this. Furthermore the length of the array is not known on compile time. Maybe we should drop this function for now.

src/circular_buffer.rs
XOSplicer: When I think about it, all constructor functions in this crate should be const fn so they can be used with static, so this function should either be removed or re-written to const

japaric:
I don't have time to continue my review of this PR today but want to mention some improvements I've been meaning to add: (a) we can use the [untagged-option] crate to avoid needing an initial value in `Vec::new`, (b) we can drop the `T: Copy` bound by using `ptr::{read,write}` and being careful about destructors and (c) we can use the `Unsize` trait instead of `AsRef` and `AsMut` to allow using arrays larger than 32 elements as backup storage. I have prototype all these ideas in [this gist].

It would be great to land these improvements in `Vec`, and probably in `CircularBuffer` as well before releasing 0.2.0. @XOSplicer would you be willing to integrate these improvements into this PR?

And in general about this PR I'd like to not provide more API surface than `std::Vec` *for now*. We can always add new API in patch releases but deprecating some API is a breaking change so I'd like to be cautious.

[untagged-option]: https://crates.io/crates/untagged-option
[this gist]: https://gist.github.com/japaric/f1ac02c216b0494213df38a65142f27a

XOSplicer:
So I see you want to be conservative about new features in this crate, which is fine.
Incorporating the functionality for non-copy types would be out of the scope for this particular PR as I would need to remodel all trait binds. However I think working with non-copy types is essential for collections.
Your gist gives a good staring point for development, which I would happily like to give a try.
I propose the following: 
1. I close this PR without it being merged to master as there are a lot of new features that might need more consideration from your side @japaric 
2. I try to re-implement the basic functionality of this crate with non-copy types.
3. Only basic features are included, no extra methods that might impose maintenance burdens.
4. I open a new PR.

What  is your opinion on that?

XOSplicer:
A short note from my side: I work on this PR mostly for the purpose to improve my skills in rust, so I'm always happy to learn something new :smile:

japaric:
@XOSplicer BTW, I took a stab at this in #4 while keeping the API minimal. (I'll need these features soon-ish so I implemented the changes). I'd love to get your feedback.

homunkulus:
:umbrella: The latest upstream changes (presumably #4) made this pull request unmergeable. Please resolve the merge conflicts.

japaric:
#4 has been merged so I'm going to close this.

@XOSplicer I hope that this served as a good exercise for you. If you are looking for more practice you could take a stab at issue #6.

PR #4, opened 2017-10-03
support arrays of any size, don't require an initialization value, ..

japaric:
single producer single consumer support for ring buffer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/*.rs.bk
+.#*
 Cargo.lock
 target/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,19 @@
 [package]
 authors = ["Jorge Aparicio <jorge@japaric.io>"]
-categories = ["data-structures", "no-std"]
+categories = [
+    "data-structures",
+    "no-std",
+]
 description = "`static` friendly data structures that don't require dynamic memory allocation"
 documentation = "https://docs.rs/heapless"
-keywords = ["static", "no-heap"]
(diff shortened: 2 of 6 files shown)


pftbest:
Sorry for interrupting, but I'm curious, why did you decide to use usize for head and tail instead of atomics. Is it for making it work on armv6 which doesn't have atomics? If I understand correctly, volatile operations are only safe on single core CPUs. On multi core CPUs you need some kind of memory fence to share the data in buffer between cores. Best option is release-acquire synchronization using atomics.

pftbest:
@japaric Not so long ago I wrote a similar spsc ring buffer and tested it using thread sanitizer

https://play.rust-lang.org/?gist=4ebd95ec79757a7d2b625cb85d58300c&version=stable

japaric:
@pftbest yes and yes. I'm also mostly interested in using this crate with single core processors (microcontrollers) but it's a good idea to note that this won't work on multicore processors. We can revisit the atomic implementation if / when rust-lang/rust#45085 otherwise it's tricky to write an implementation for ARMv6-M.

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/heapless/builds/295391877?utm_source=github_status&utm_medium=notification)

japaric:
@homunkulus retry

homunkulus:
:hourglass: Testing commit da5757a71b7dab140b30c1a7bb94d5d2394a0490 with merge 10c5542c989b49a0f78f8ac201319bff3e2e199d...

PR #10, opened 2017-11-09
What if tsan is right?

pftbest:
Is speedup from one less barrier worth the risk in this case?

### DIFF
--- a/src/ring_buffer/spsc.rs
+++ b/src/ring_buffer/spsc.rs
@@ -88,8 +88,8 @@ where
         let n = rb.capacity() + 1;
         let buffer: &mut [T] = unsafe { rb.buffer.as_mut() };
 
-        let head = rb.head.load_relaxed();
         let tail = rb.tail.load_relaxed();
+        let head = rb.head.load_acquire();
         let next_tail = (tail + 1) % n;
         if next_tail != head {
             // NOTE(ptr::write) the memory slot that we are about to write to is uninitialized. We
--- a/tests/tsan.rs
+++ b/tests/tsan.rs
@@ -87,7 +87,7 @@ fn contention() {
             scope.execute(move || {
                 let mut sum: u32 = 0;
 
-                for i in 0..N {
+                for i in 0..(2*N) {
                     let i = i as u8;
                     sum = sum.wrapping_add(i as u32);
                     while let Err(_) = p.enqueue(i) {}
@@ -99,7 +99,7 @@ fn contention() {
             scope.execute(move || {
                 let mut sum: u32 = 0;
 
-                for _ in 0..N {
+                for _ in 0..(2*N) {
                     loop {
                         match c.dequeue() {
                             Some(v) => {


pftbest:
ppf from #llvm suggested that this might be a false positive:

>‎[17:28] ‎<‎ppf‎>‎ pftbest_: you need an acquire in the first load
‎[17:29] ‎<‎ppf‎>‎ oh wait, i'm wrong, that's the producer
‎[17:30] ‎<‎ppf‎>‎ i feel like tsan is wrong here
‎[17:36] ‎<‎ppf‎>‎ pftbest_: i believe tsan simply doesn't support adhoc synchronization like this
‎[17:36] ‎<‎ppf‎>‎ which means you gotta live with the false positive

japaric:
@pftbest you asked what's the cost of `dmb` on the Cortex-M. I finally got the chance to measure it.

This program:

``` rust
// `static P: Producer<'static, u8, [u8; 8]>;`
fn exti0(_t: &mut Threshold, r: EXTI0::Resources) {
    r.P.enqueue(42).ok();

    rtfm::bkpt();
}

// `static C: Consumer<'static, u8, [u8; 8]>;`
fn exti1(_t: &mut Threshold, r: EXTI1::Resources) {
    if let Some(x) = r.C.dequeue() {
        unsafe { ptr::read_volatile(&x) };
    }

    rtfm::bkpt();
}
```

Produces the following disassembly (using this PR):

``` console
0800019c <EXTI1>:
 800019c:       f240 0010       movw    r0, #16
 80001a0:       f2c2 0000       movt    r0, #8192       ; 0x2000
 80001a4:       6800            ldr     r0, [r0, #0]
 80001a6:       6842            ldr     r2, [r0, #4]    ; bkpt
 80001a8:       f3bf 8f5f       dmb     sy
 80001ac:       6801            ldr     r1, [r0, #0]
 80001ae:       4291            cmp     r1, r2
 80001b0:       d00d            beq.n   80001ce <EXTI1+0x32>
 80001b2:       b081            sub     sp, #4
 80001b4:       1842            adds    r2, r0, r1
 80001b6:       3101            adds    r1, #1
 80001b8:       7a12            ldrb    r2, [r2, #8]
 80001ba:       f001 0107       and.w   r1, r1, #7
 80001be:       f3bf 8f5f       dmb     sy
 80001c2:       6001            str     r1, [r0, #0]
 80001c4:       f88d 2003       strb.w  r2, [sp, #3]
 80001c8:       f89d 0003       ldrb.w  r0, [sp, #3]
 80001cc:       b001            add     sp, #4
 80001ce:       be00            bkpt    0x0000
 80001d0:       4770            bx      lr

080001d2 <EXTI0>:
 80001d2:       f240 0014       movw    r0, #20
 80001d6:       f2c2 0000       movt    r0, #8192       ; 0x2000
 80001da:       6800            ldr     r0, [r0, #0]
 80001dc:       6841            ldr     r1, [r0, #4]    ; bkpt
 80001de:       6803            ldr     r3, [r0, #0]
 80001e0:       f3bf 8f5f       dmb     sy
 80001e4:       1c4a            adds    r2, r1, #1
 80001e6:       f002 0207       and.w   r2, r2, #7
 80001ea:       429a            cmp     r2, r3
 80001ec:       d005            beq.n   80001fa <EXTI0+0x28>
 80001ee:       4401            add     r1, r0
 80001f0:       232a            movs    r3, #42 ; 0x2a
 80001f2:       720b            strb    r3, [r1, #8]
 80001f4:       f3bf 8f5f       dmb     sy
 80001f8:       6042            str     r2, [r0, #4]
 80001fa:       be00            bkpt    0x0000
 80001fc:       4770            bx      lr
```

The execution time for `EXTI0` (from `bkpt` to `bkpt`) is 14 cycles and the execution time for
`EXTI1` is 20 cycles.

If I change all the atomic operations to relaxed operations and add a compiler barrier where
required all the `dmb` instructions disappear and both execution times reduce by 2 cycles. So the
cost of `dmb` is 1 cycle, at least on a single core Cortex-M3 -- maybe it's implemented as a NOP
on the device I'm using?

If this makes TSan happy then I'm OK with the change.

@homunkulus r+

homunkulus:
:pushpin: Commit e5e24f0 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit e5e24f0844801d989ad156e740388a868e898580 with merge cd538447c9cbcb9da4754581a2b6ac6acd43a104...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/heapless/builds/305955499?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing cd538447c9cbcb9da4754581a2b6ac6acd43a104 to master...

pftbest:
@japaric

[I've asked about this specific case on tsan bug tracker](https://github.com/google/sanitizers/issues/882), and it turns out this is actually UB in C++ memory model (and most probably in Rust memory model too, since they are the same atm). Most probably LLVM would never break this particular code, so we could have kept the relaxed operations, and just added a comment saying that we are relying on UB. But who knows what would happen in the future.

parched:
FWIW I suspect if you tell LLVM it's single threaded (i.e., single core) you will be able to get rid of the memory barrier instructions. You can do this now via the target spec by setting `singlethread` to `true`.

japaric:
@parched Interesting. Setting `singlethread` to `true` does remove the memory barriers. The generated machine code looks correct to me but I'm not sure if changing ThreadModel to Single is properly generating compiler fences where required or just turning atomics into plain integers. (Even with this tweak the machine code doesn't look optimal to me but that may be due to the current implementation).

Having to create a new custom target to tweak `singlethread` is a bit overkill though but probably the proper way to do this. (I don't think the built-in thumb targets should be tweaked to have singlethread = true because heterogeneous multi-core microcontrollers exist and built-in targets should be as general as possible; i.e. they should generate correct code in both the single-core and multi-core case).

parched:
@japaric Yes, `singlethread` just turns it into a compiler barrier.

> Having to create a new custom target to tweak `singlethread` is a bit overkill though but probably the proper way to do this.

We could probably add it as a flag in `rustc`.

PR #12, opened 2017-12-18
Documented that the current implementation will only work correctly on single-core devices

fluffysquirrels:
As mentioned in <https://github.com/japaric/heapless/issues/5#issuecomment-352502450>


fluffysquirrels:
Oops, looks like multi-core was implemented. Ignore this.

PR #15, opened 2017-12-31
Fix buffer capacity off by one error

robert-w-gries:
I was looking to implement a UART input buffer as a `RingBuffer`, and I ran into a surprising `panic` while stress testing the driver.

#### Here's a minimal reproduction of the `panic`:

```rust
use heapless::RingBuffer;

let mut test: RingBuffer<u8, [u8; 2]> = RingBuffer::new();

println!("Capacity = {}", test.capacity());

test.enqueue(1);
test.enqueue(2);

test.dequeue().expect("Expected data");
test.dequeue().expect("Expected data");
test.dequeue().expect("This is expected to panic");
```

The problem is that the `capacity()` method returns one less than it should.  The fix is pretty simple.

```diff
-        buffer.len() - 1
+        buffer.len()
```

#### Error

```
Capacity = 1

PANIC in [snip]
    Expected data
```

#### Here's the output with fix

```
Capacity = 2

PANIC in [snip]
    This is expected to panic
```

### DIFF
--- a/src/ring_buffer/mod.rs
+++ b/src/ring_buffer/mod.rs
@@ -76,7 +76,7 @@ where
     /// Returns the maximum number of elements the ring buffer can hold
     pub fn capacity(&self) -> usize {
         let buffer: &[T] = unsafe { self.buffer.as_ref() };
-        buffer.len() - 1
+        buffer.len()
     }
 
     /// Returns the item in the front of the queue, or `None` if the queue is empty


pftbest:
Ring buffer should always have one empty slot to function properly. So there is no error here.

robert-w-gries:
I'll close the PR since there's no error.

Is this behavior an implementation detail?  It'd be more intuitive if the API abstracted away the implementation details.

From a user perspective, it was a surprise when I tried to create a `RingBuffer<u8>` of size `n` and the buffer only supported `n-1` elements.

I see that the documentation for `new()` says the following

> Creates an empty ring buffer with capacity equals to the length of the array A minus one.

But it doesn't really explain why the buffer is created with lower capacity.  It'd also be helpful to have more documentation regarding the usage of the `RingBuffer`.

I can create a separate tracking issue if these are valid requirements.

pftbest:
Yes, it is implementation detail, but I don't know a good way to abstract it, unless Rust gets const generics.

It is easy to explain why the buffer needs one empty slot, just consider the following sequence of events:

1) First we start with an empty buffer:
```
0  1  2  3
+--+--+--+--+    head = 2
|__|__|__|__|    tail = 2
+--+--+--+--+
      ^
      H
      T
```
`head` is index where we have a first cell with data, `tail` is index where we have first empty slot. On `enqueue()` we increment `tail`, on `dequeue()` we increment `head`. 

2) Then we start pushing some data into the buffer:
```
+--+--+--+--+    head = 2
|__|__|XX|__|    tail = 3
+--+--+--+--+
      ^  ^
      H  T
```
```
+--+--+--+--+    head = 2
|__|__|XX|XX|    tail = 0
+--+--+--+--+
^     ^
T     H
```
```
+--+--+--+--+    head = 2
|XX|__|XX|XX|    tail = 1
+--+--+--+--+
   ^  ^
   T  H
```
3) Now we have only one empty cell and we continue pushing data:
```
+--+--+--+--+    head = 2
|XX|XX|XX|XX|    tail = 2
+--+--+--+--+
      ^
      H
      T
```
As you can see we arrived at the same spot where we started (`head = 2`, `tail = 2`). So there is no way to distinguish an empty buffer from a completely full buffer. To solve this ambiguity we forbid filling the buffer completely, and decide that if `head == tail` the buffer is empty.

It is possible to solve this problem by introducing a third variable, but such design would not be concurrent, since both `enqueue()` and `dequeue()` would need to write to this common variable, causing data races, unless you introduce locks (which is expensive).

robert-w-gries:
Thanks for the explanation!  The problem is now more clear.

> It is possible to solve this problem by introducing a third variable, but such design would not be concurrent, because both enqueue() and dequeue() would need to write to this common variable, causing data races, unless you introduce locks (which is expensive).

This was the piece I was missing.  I wondered why there wasn't a third counting variable used. It sounds like a tradeoff between usability and performance, and this library prioritizes performance due to its use cases.  Until const generics are implemented, this problem will remain.

A note in the documentation describing what you posted above would be helpful for newbies like me 😃

pftbest:
I have found an interesting article on this topic

https://fgiesen.wordpress.com/2010/12/14/ring-buffers-and-queues/

PR #26, opened 2018-04-19
Ringbuffer iterators macro

### LINKED ISSUE #6: Improve / refactor `RingBuffer` iterators
The implementation of `ring_buffer::{Iter,IterMut}` is pretty much a copy of the other. The duplication can be eliminated by using a macro (see implementation of iterators in libcore). Also both `Iter` and `IterMut` use the default methods provided by the `Iterator` trait; the `Iterator` implementation for those two could override the default methods for better performance.

XOSplicer:
Resolves #6 

Refactor the ring_buffer::{Iter,IterMut} implementations to use a macro, so there is less duplicate code.
Inspired by the[ slice iterator macro in lib core](https://github.com/rust-lang/rust/blob/master/src/libcore/slice/mod.rs).

I have never worked with macros to generate const / mut implementations.
So this might be a little rough.

`#![allow(stable_features)]` is needed for current rust nightly, since otherwise `conservative_impl_trait` would throw a warning. Removing the feature flag would not be backwards compatible with earlier nightly versions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,6 +111,7 @@
 
 #![deny(missing_docs)]
 #![deny(warnings)]
+#![allow(stable_features)]
 #![feature(const_fn)]
 #![feature(core_intrinsics)]
 #![feature(conservative_impl_trait)]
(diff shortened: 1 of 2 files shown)


japaric:
Thanks @XOSplicer!

bors r+

PR #38, opened 2018-05-06
Implement IntoIter for Vec

XOSplicer:
~~This implementation uses vec::reverse and vec::pop to move the content of the vector out of it.~~

It might give more performance to avoid the `reverse` and read each element individually. In the end  the unmoved items need to be dropped.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -310,6 +310,42 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


japaric:
Thanks @XOSplicer. This LGTM.

bors r+

Issue #36, opened 2018-05-04
FromIterator for Vec


burrbull:
Is it possible to implement `IntoIterator` for `Vec` and maybe other?
It is needed for `.collect()`.

XOSplicer:
`IntoIterator<Item = &'a T>` and `IntoIterator<Item = &'a mut T>` are already implemented for `Vec<T, ..>`. However it might indeed be useful to implement `IntoIterator<Item = T>` which would have move semantics while iterating over the vector. This would gradually move all elements out of the vector and could drop the underlaying buffer once the collection is empty.

The trait that needs to be implemented to allow the usage of `.collect<heapless::Vec<T, ...>>` is `FromIterator`. It is currently not implemented for `Vec<T, ...>`. It can however in general by unsafe to collect an iterator with arbitrary many items into an Vec, which has a fixed size, statically allocated buffer for storage. As a user of the `collect` method you would need to specify how large this buffer is. But since `from_iter()` does not return a result, we could `panic!` if the buffer runs full, or simply drop al extra items.

burrbull:
Yes, you are right. I mean `FromIterator`

XOSplicer:
closed in #38 and #39

PR #39, opened 2018-05-06
Implement FromIterator for Vec

XOSplicer:
As discussed in #36 

Panics when the vector can not hold all values supplied by the iterator.
Calling `unwrap` on the result might not be the best choice here.
It might be better to match the result and have a custom panic message.
Is this the best possible behavior?

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -323,7 +323,7 @@ where
     {
         let mut vec = Vec::new();
         for i in iter {
-            vec.push(i).expect("Vec::from_iter overflow");
+            vec.push(i).ok().expect("Vec::from_iter overflow");
         }
         vec
     }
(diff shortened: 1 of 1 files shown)


src/vec.rs
```diff
@@ -310,6 +312,25 @@ where
...
+where
+    N: ArrayLength<T>,
+{
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = T>,
+    {
+        let mut vec = Vec::new();
+        for i in iter {
+            match vec.push(i) {
```
japaric: nit: this can be rewritten as `vec.push(i).expect("Vec::from_iter overflow")`.

src/vec.rs
```diff
@@ -323,7 +323,7 @@ where
     {
         let mut vec = Vec::new();
         for i in iter {
-            vec.push(i).expect("Vec::from_iter overflow");
+            vec.push(i).ok().expect("Vec::from_iter overflow");
```
XOSplicer: In the afterthought, this could also be implemented in terms of `Vec::extend` which would reduce code duplication and needed space for the statically allocated except message.

Issue #40, opened 2018-05-15
compile on stable


japaric:
Off the top of my head these things need to be done to get this compiling on stable:

- Put all `const` functions behind a `const-fn` feature.

- Put the generic atomic type (`Atomic<T>`) behind a Cargo feature. `AtomicUsize` should be used everywhere when that feature is disabled.

- Remove the `nonzero` feature ( `NonZeroU32`) in favor of manually checking for some magic value that represents `None`.

XOSplicer:
I might look into the `const-fn` feature gate, when I find the time this week.

XOSplicer:
Removing `NonZeroU32` is ready in #50.
When no default features are enables, heapless now compiles on stable rust.

PR #43, opened 2018-05-21
Put all const functions behind a const-fn feature

### LINKED ISSUE #41: Put all const functions behind a const-fn feature.
Put all `const` functions behind a `const-fn` feature.
As an idea to target stable from #40

XOSplicer:
## Purpose
This PR introduces the `const-fn` feature gate, which when enabled makes most `new` methods `const` 
and therefore usable for initializing `static` variables.
The idea was introduced in #40 with the purpose to come closer to targeting stable rust.
`const` functions are currently only available on rust nightly (tracking issue: [const fn tracking issue (RFC 911)](https://github.com/rust-lang/rust/issues/24111)).
In order to target stable rust this feature is made opt-in.

This feature is a **breaking change** as users of the library now need to explicitly enable it by changing their `Cargo.toml`:
```
...
[dependencies]
heapless = { version = "0.4.0", features = ["const-fn"] }
...
```

## Approach
The implementation of the feature mainly consist of the `const_fn!` macro, which takes a function with `const` modifier
and removes the modifier if the feature gate is not activated.
For the `const` functions a test is intoduced that checks if `static` variables can be initialized.
These tests are only active if the feature is active.
I have not found a way to make doc-test depend on a feature. Therefore some doc-tests are adapted, so that no static initialization is necessary.
The `ci/script.sh` is adapted to also tests with the `--all-feature` flag

## Future
When in the future the `const_fn` rust feature becomes stable, this feature gate **might become active by default**.

Closes #41 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ repository = "https://github.com/japaric/heapless"
 version = "0.3.6"
 
 [features]
+default = ["const-fn"]
 const-fn = []
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,9 @@ name = "heapless"
 repository = "https://github.com/japaric/heapless"
 version = "0.3.6"
 
+[features]
+const-fn = []
+
 [dev-dependencies]
 scoped_threadpool = "0.1.8"
 
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -5,18 +5,18 @@ main() {
 
     if [ $TARGET = x86_64-unknown-linux-gnu ]; then
         cargo test --target $TARGET
-        cargo test --target $TARGET --all-features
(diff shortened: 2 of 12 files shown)


XOSplicer:
The feature could be a **non-breaking change** if it was turned on by default. However then to disable it on stable, the `default-features = false` flag needs to be set in the manifest of the consuming crates.

japaric:
Thanks for the PR, @XOSplicer. Could you leave the const-fn feature enabled by default for now? I'd like to hold off on making a breaking change for now.

XOSplicer:
Sure, the const-fn feature is now on by default, therefore the ci script will additionally run tests with `--no-default-features`

japaric:
> A question arising:

It's fine to do both in the same builder.

Thanks again.

bors r+

PR #44, opened 2018-06-10
Added Consumer::ready and Producer::ready indicating if it is ready t…

Frans-Willem:
…o enqueue/dequeue.

I needed this to use it with a serial library. e.g. without it:
```
if let Some(x) = serial.read() {
if let Err(x) = producer.enqueue(x) {
// Now what ?
}
}
```
With the ready flag:
```
if producer.ready() {
if let Some(x) = serial.read() {
producer.enqueue(x).unwrap();
}
}
```

### DIFF
--- a/src/ring_buffer/mod.rs
+++ b/src/ring_buffer/mod.rs
@@ -556,4 +556,32 @@ mod tests {
 
         assert_eq!(rb.len(), 2);
     }
+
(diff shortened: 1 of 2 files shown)


japaric:
Thanks for the PR, @Frans-Willem. It LGTM.

bors r+

PR #45, opened 2018-06-20
Fix incorrect length calculation for RingBuffer when head > tail

ReeceStevens:
When the head index becomes greater than the tail, the `len()` function returns the incorrect length. This causes iterations over ring buffers to stop early as well, since they use the same calculation.

This PR adds two (very minimal) tests that demonstrate the error.

### DIFF
--- a/src/ring_buffer/mod.rs
+++ b/src/ring_buffer/mod.rs
@@ -335,7 +335,7 @@ macro_rules! impl_ {
                 let tail = self.tail.load_relaxed();
 
                 if head > tail {
-                    head - tail
+                    self.capacity() + 1 - head + tail
                 } else {
                     tail - head
                 }
--- a/src/ring_buffer/mod.rs
+++ b/src/ring_buffer/mod.rs
@@ -207,7 +207,7 @@ where
         let tail = self.tail.load_relaxed().into();
 
         if head > tail {
-            head - tail
+            self.capacity().into() + 1 - head + tail
         } else {
             tail - head
         }
--- a/tests/tsan.rs
+++ b/tests/tsan.rs
@@ -155,3 +155,20 @@ fn unchecked() {
 
     assert_eq!(rb.len(), N::to_usize() / 2);
 }
+
+#[test]
+fn len_properly_wraps() {
+    type N = U3;
+    let mut rb: RingBuffer<u8, N> = RingBuffer::new();
+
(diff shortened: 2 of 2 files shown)


ReeceStevens:
I am happy to make formatting and convention changes-- wasn't sure if there was a particular style I needed to adhere to.

ReeceStevens:
@japaric I wanted to follow up on this PR, since I believe this issue is causing incorrect behavior for any use of a RingBuffer where the head index becomes greater than the tail index (i.e. the buffer "wraps around"). Let me know if there is any further testing or documentation you would like me to add.

And I should have mentioned this earlier, but thanks for your great work on this library. I'm using it for a `no-std` FIR filter library: https://github.com/reecestevens/digital-filter

XOSplicer:
Hi @ReeceStevens,
Your tests seem reasonable to check the two issues you gave.
I have checked your tests against the current master, which fails them (see below).
Your PR does fix them, so I belive it would be good to get the changes into the codebase.


```
    Running target\debug\deps\tsan-2b58e509499adc95.exe

running 7 tests
test iterator_properly_wraps ... FAILED
test len_properly_wraps ... FAILED
test once ... ok
test twice ... ok
consumer: 261120
producer: 2096128
test contention ... ok
test scoped ... ok
test unchecked ... ok

failures:

---- iterator_properly_wraps stdout ----
thread 'iterator_properly_wraps' panicked at 'assertion failed: `(left == right)                                                                            `
  left: `[2, 3, 4]`,
 right: `[2, 0, 0]`', tests\tsan.rs:191:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- len_properly_wraps stdout ----
thread 'len_properly_wraps' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `3`', tests\tsan.rs:173:5


failures:
    iterator_properly_wraps
    len_properly_wraps

test result: FAILED. 5 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

ReeceStevens:
Thanks for the assessment, @XOSplicer. Is there any action I should take to move forward the merge?

XOSplicer:
I don't see any blockers, but its up to @japaric to go ahead

XOSplicer:
bors try

japaric:
Thanks for the PR, @ReeceStevens.

bors r+

PR #50, opened 2018-07-15
Remove NonZeroU32 and test on stable

XOSplicer:
# Purpose
This PR removes the need to use `NonZeroU32` in `indexmap::CoreMap` and the resulting `IndexMap` and `IndexSet`.
The idea was proposed in #40 and now the library can be **compiled on stable rust**   🎉

# Approach
Instead of using `Option<Pos>` and `Pos` with a `NonZeroU32` combined value of `index` and `hash`,
now `OptionalPos` and `Pos` are used, where a zero value in `OptionalPos` represents `None`.
This way `mem::zeroed` can still be used to initialized the empty array of positions.

So far all test pass, I hope this is sufficient to prove the correctness of the implementation.
Otherwise, let me now if there are any mistakes in the changes.

There are some `FIXME`s in the changes, wherever an `OptionalPos` is created and the case that the position woud result in a `None` is just unwrapped. 
This should however by very rare, only when the index reaches 0xffff and the hash happens to be 0xffff aswell.

I believe this is a **non-breaking** change since I did not need to alter any tests.
But looking at the available test now, I see that actually the interface of the InsexMap is not practically tested. 
Maybe this is another issue to address.

However `Pos` and `OptionalPos` are exported publically, so it counts towards the public interface which would be changed by this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: rust
 matrix:
   include:
     - env: TARGET=x86_64-unknown-linux-gnu
+      rust: stable
+      if: branch != master
+
+    - env: TARGET=x86_64-unknown-linux-gnu
       rust: beta
       if: branch != master
 
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -1,6 +1,5 @@
 use core::borrow::Borrow;
 use core::iter::FromIterator;
-use core::num::NonZeroU32;
 use core::{fmt, ops, slice};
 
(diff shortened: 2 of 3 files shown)


XOSplicer:
ping @japaric

japaric:
Thanks for the PR, @XOSplicer, and sorry for the delay.

NonZeroU32 is available on the latest stable release so I'd prefer to keep the code as it is to keep things simple.

PR #54, opened 2018-08-19
CI: run tests on stable

### LINKED ISSUE #40: compile on stable

japaric:
closes #40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,15 @@ language: rust
 matrix:
   include:
     - env: TARGET=x86_64-unknown-linux-gnu
+      rust: stable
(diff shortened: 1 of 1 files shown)


japaric:
The nightly builds will be fixed in a follow up PR.

PR #62, opened 2018-11-22
[WIP] Implement serde::Deserialize for Vec

chrysn:
This is a first working branch that implements serde's Deserialize for `heapless::Vec`, and would (if completed) close https://github.com/japaric/heapless/issues/58.

I'm unsure when I'll find the time to complete this (a full solution would at least include tests, implementations of Serialize and Deserialize for Vec and String; in-place would be nice too). It is still published at this early stage in the hope that anyone joining this effort (@sanmai-NL?) might find it a useful starting point.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ version = "0.4.0"
 [features]
 const-fn = []
 smaller-atomics = []
+with-serde = ["serde"]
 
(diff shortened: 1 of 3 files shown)


japaric:
Thanks for this PR, @chrysn. #66 is more complete so I'm going to close this PR in favor of that one.

PR #63, opened 2018-12-15
spsc::Queue: add single core variant

japaric:


japaric:
published: https://crates.io/crates/heapless/0.4.1

PR #69, opened 2019-01-24
Bump generic-array to 0.12

vorot93:


japaric:
Thanks for the PR, @vorot93. The public API of `generic-array` is re-exported in this crate (`ArrayLength`). Due to this bumping the minor version of `GenericArray` is considered a breaking change so this PR must wait until we are ready to publish v0.5.0 of this crate.

PR #75, opened 2019-02-05
Impl conversion from various number types for String

vorot93:


src/string.rs
```diff
@@ -538,6 +541,31 @@ impl_eq! { String<N>, &'a str }
...
 
+macro_rules! impl_from_num {
+    ($num:ty, $size:ty) => {
+        impl<N> From<$num> for String<N>
+        where
+            N: ArrayLength<u8> + IsGreaterOrEqual<$size, Output = True>,
+        {
+            fn from(s: $num) -> Self {
+                let mut new = String::new();
+                core::fmt::Write::write_fmt(&mut new, format_args!("{}", s)).unwrap();
```
japaric: nit: could you change this to make use of the write! macro?

japaric:
Thanks!

bors r+

PR #76, opened 2019-02-05
CapacityError

vorot93:


src/errors.rs
```diff
@@ -0,0 +1,31 @@
...
+
+const CAPERROR: &'static str = "insufficient capacity";
+
+/// Capacity error returned when container's capacity is not enough to complete the operation
+#[derive(Debug, PartialEq)]
+pub struct CapacityError {
+    /// Maximum characters allowed
+    pub maximum: usize,
+    /// Encountered number of characters
+    pub encountered: usize,
```
XOSplicer: It would be great if there was a field to return the data that was left over in the operation that overflowed. Like so:
```
struct CapacityError<T = ()> {
    pub maximum: usize,
    pub encountered: usize,
    pub rest: T 
}
```

What are your opinions on that one?

japaric:
@vorot93 could you say more about the motivation for this change?

I'm not really in favor of using std::error::Error since there were some plans to deprecate it in favor of failure::Fail.

This PR doesn't seem to replace *all* the instances of `Result<_, ()>` with `Result<_, CapacityError>` which seems a bit strange. Is there a reason for that?

vorot93:
@japaric CapacityError is a lot more informative than an empty tuple to the end user, showing by how much the client has overshot the container.

Places that still return empty tuple as the error are there by oversight.

jonas-schievink:
> I'm not really in favor of using std::error::Error since there were some plans to deprecate it in favor of failure::Fail.

FWIW this is no longer the plan. `std::error::Error` is getting various improvements that make it a good replacement for `failure::Fail`. I believe it will stay std-only though.

vorot93:
@japaric what do you think of this? Can this be merged?

vorot93:
ping @korken89 @japaric

Issue #56, opened 2018-09-01
impl Debug, Default, Clone, comparators and Hash


TeXitoi:
Having the structures implementing all these standard traits will improve the ergonomics of this crate.

XOSplicer:
I agree with Debug, Default, Eq, PartialEq and Hash. However most of them are available through `Deref<Target=[T]>` on Vec<T>. Clone on the other hand would not work as the typical heap allocated clone but would yield a new stack allocated instance. If @japaric doesn't have the time, I can look into this

TeXitoi:
The big blocker is that I can't derive when my strict contain a heapless object.

For clone, the semantic is valid, and don't forget that any Copy is Clone. It would be possible to derive Copy on these objects, bit I don't think it would be great semantically.

XOSplicer:
Sure go ahead, you might want to check https://rust-lang-nursery.github.io/api-guidelines/interoperability.html for a list of Traits that should / could be implemented for every of the public data types

XOSplicer:
I tried to implement some of the traits:
https://github.com/XOSplicer/heapless/pull/8/files
however I were hitting some things to decide upon: 
- BinaryHeap has no `Eq` implementation in the standard lib, shoult it have one?
- Implementing `Ord` for Vec is not straight forward
What are your thoughts on that?

TeXitoi:
`Eq` is not implementable for `BinaryHeap` as the same elementx can be in different order depending on the insertion order.

For `Ord` on `Vec`, you can just forward to `impl<T> Ord for [T] where T: Ord`.

XOSplicer:
i guess when `Eq`, is not implemented for `BinaryHeap`, so is `Hash` aswell not.

Well the problem I was facing with `Ord` for `Vec` is that one should be able to compare a `Vec<T, N>` against a `Vec<T,M>` where N and M are static capacities. However I failed at expressing that successfully in the trait implementation. Maybe you come up with an implementation for that one. A similar implementation for `String<N>` would be the next step, as the same problem is present there.

TeXitoi:
For `Ord` between `Vec<T, N>` and `Vec<T, M>`, the `Ord` trait doesn't allow it as lhs and rhs must be the same type.

sanmai-NL:
[...]

PR #77, opened 2019-02-05
impl Hash for Vec and String

vorot93:
ref #56


japaric:
@vorot93 thanks!

bors r+

PR #78, opened 2019-02-10
Implement Default for Vec, String, LinearMap, BinaryHeap

XOSplicer:
As pointed out in #56 

The implementation for Queue is missing, as it would be opinionated about the atomic size to use and whether it is a single or multi core variant.


japaric:
Thanks!

bors r+

PR #81, opened 2019-02-10
Implement Clone for Queue

XOSplicer:
ref #56

### DIFF
--- a/src/spsc/mod.rs
+++ b/src/spsc/mod.rs
@@ -668,4 +668,22 @@ mod tests {
         assert_eq!(c.ready(), false);
         assert_eq!(p.ready(), true);
     }
+
+    #[test]
+    fn clone() {
+        let mut rb1: Queue<i32, U4> = Queue::new();
+        rb1.enqueue(0).unwrap();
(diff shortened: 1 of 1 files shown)


XOSplicer:
bors retry

XOSplicer:
Sorry for the merge conflicts

src/spsc/mod.rs
```diff
@@ -394,6 +394,30 @@ macro_rules! impl_ {
...
         }
+
+        impl<T, N, C> Clone for Queue<T, N, $uxx, C>
+        where
+            T: Clone,
+            N: ArrayLength<T>,
+            C: sealed::XCore,
+        {
+            fn clone(&self) -> Self {
+                let mut new = Queue {
```
japaric: I think you'll need to specify the full type here (`let mut new: Queue<..> = ..`), otherwise method lookup will fail below.

japaric:
bors delegate+

You can r+ this after tests are fixed.

src/spsc/mod.rs
XOSplicer: Yes, this does indeed solve the problem 👍

PR #84, opened 2019-02-14
Implement Eq and PartialEq for Indexmap

XOSplicer:
ref #56


japaric:
@XOSplicer thanks!

bors r+

PR #85, opened 2019-02-15
Draft: CapacityError with failure

XOSplicer:
Lets have a discussion about how Errors might be handeled in 0.5.x

I think going with failure is more resonable, especially since it supports `no-std`.

The biggest open question I have is: How do we handle operations that failed inserting an element?
It might be desirable to get the element back, in order to do something with it, (e.g., place it in another buffer). However putting this element into the CapacityError is not going to work because the element would need to implement `Send + Sync`.

What are your opinions on this?
@vorot93 @japaric 

ref #76

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,8 @@ hash32 = "0.1.0"
 version = "1"
 optional = true
 default-features = false
+
+[dependencies.failure]
+version = "0.1.5"
+default-features = false
+features = ["derive"]
--- a/src/binary_heap.rs
+++ b/src/binary_heap.rs
@@ -16,6 +16,7 @@ use core::{mem, ptr, slice, fmt};
 use generic_array::ArrayLength;
 
 use Vec;
+use errors::CapacityError;
 
 /// Min-heap
 pub enum Min {}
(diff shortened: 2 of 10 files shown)


japaric:
I personally have no need for improving error messages or making errors
inter-operate with `failure` / `std::error`. For my embedded / `no_std` work
whenever I encounter a capacity error I always do one of these three things.

- Panic. I do this when the error is a bug: i.e. I selected a buffer size that's
  too small for the application and there's no way to handle error at runtime.
  The fix is to increase the collection capacity and re-compile the program. In
  this case something like `vec.push(item).unwrap_or_else(|| panic!())` gives me
  the best panic location information so that's what I use.

- Ignore the error. And maybe log an error message. For example if I'm coding a
  network application and I run out of memory (queue space) to handle incoming
  packets then I simply drop new packets. The client is likely to retry the
  request.

- Handle the error *immediately*. For example if I have to enter an entry in my
  ARP / neighbor cache and it's full then I do whatever the RFC / spec says or
  seems reasonable (e.g. drop the oldest, stale entry).

I don't think I ever had the need for bubbling up this kind of error using `?`.
I can see myself eventually doing it for things like `vec.push(a)?; (..)
vec.push(b)?; (..)` but only when *no* error conversion is involved.

I would be interested in hearing how people intend to use a `failure` /
`std::error` compatible API so we can design an API that fits their needs. Is
the use case CLI applications? (It seems kind of wrong to bubble up such error
and present it to the user though since the user can't fix the problem by
invoking the tool differently -- the only fix would be to increase capacity and
recompile the tool. A panic indicating a software bug seems more appropriate).

vorot93:
My use case is a little bit different.

While interacting with third party APIs over the network one can never be sure that the data that they send is ever correct and of expected size. Putting storage size constraint into API contract creates an additional layer of security against malicious inputs that could otherwise cause Denial-of-Service.
```
#[derive(Deserialize)]
struct Input {
    strings: Box<heapless::Vec<String, U100>>,
}
```

As you can understand, outright panicking is quite undesirable here. Of course, you can always force panic by calling `unwrap` on the result. But why force it?

PR #86, opened 2019-02-17
CapacityError and CapacityResult with Failure

XOSplicer:
This PR introduces an `CapacityResult` used for insertion operations.
It can either be Ok or an error. If it is ok, it contains an return value. If it is an error, it contains an CapacityError and an `rest` item .

This result type makes the insertion API more ergonomic, as the `CapacityResult` can be transformed into a normal result by calling `CapacityResult::into_result()`. This regular result can then be used for error handling without the `rest` data.

ref #85 #76

### DIFF
--- a/src/linear_map.rs
+++ b/src/linear_map.rs
@@ -5,7 +5,7 @@ use core::iter::FromIterator;
 use generic_array::ArrayLength;
 
 use Vec;
-use errors::CapacityError;
+use CapacityResult;
 
 /// A fixed capacity map / dictionary that performs lookups via linear search
 ///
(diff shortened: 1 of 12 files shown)


src/linear_map.rs
```diff
@@ -186,14 +187,17 @@ where
...
             mem::swap(v, &mut value);
-            return Ok(Some(value));
+            return CapacityResult::ok(Some(value))
         }
 
-        self.buffer.push((key, value))?;
-        Ok(None)
+        let p = self.buffer.push((key, value));
+        if p.is_err() {
+            return p.map(|_| None); // convert type
```
XOSplicer: Maybe this can be made more ergonomic by implementing `From<CapacityResult<T, R>> for CapacityResult<T2, R2> where T2: From<T>, R2: From<R>`

PR #87, opened 2019-04-21
Adds 'peek' and 'peek_ref' functions.

thejpster:
These go nicely with 'pop' and 'push' when you're using Vec as a Queue.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -123,6 +123,28 @@ where
         }
     }
 
+    /// Returns a copy of the last element in the vector, or `None` if it's
+    /// empty.
+    pub fn peek(&mut self) -> Option<T> where T: Copy {
+        let len = self.len();
+        if len >= 1 {
+            self.get(len - 1).map(|x| *x)
+        } else {
+            None
+        }
+    }
+
+    /// Returns a reference to the last element in the vector, or `None` if
+    /// it's empty.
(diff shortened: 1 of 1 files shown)


japaric:
Thanks for the PR. These seem equivalent to `last()` and `last().cloned()` and I'd rather not extend the API beyond what's provided by the `std` counterpart so I'm not going to accept this PR. If you want to access `last` functionality under another name I would suggest using a newtype or a extension trait.

thejpster:
Ah, I hadn't twigged the desire for API compatibility. That might be worth making explicit. Or maybe I missed it. But yes, last() would do fine.

Although, I can't see 'fn last' on https://docs.rs/heapless/0.5.0-alpha.2/heapless/struct.Vec.html. I must be missing something?

japaric:
`last()` is a [slice method](https://doc.rust-lang.org/std/primitive.slice.html#method.last). `heapless::Vec` derefs to a slice so it inherits all its methods.

I thought `rustdoc` showed these docs; it shows them for [`std::Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html#deref-methods) -- not sure why that's not working in this crate; I think it did at some point. I'll add a note in the docs.

thejpster:
Aha! Thank you!

PR #97, opened 2019-05-19
Turn `binary_heap::Kind` into a sealed trait

### LINKED ISSUE #72: turn `binary_heap::Kind` into a sealed trait
this is technically a breaking change and should be done when we are ready for a new minor release

XOSplicer:
resolves #72

breaking change, since a previously public trait is now private


japaric:
Thanks!

bors r+

japaric:
bors retry

PR #99, opened 2019-05-22
add multiple-producer multiple-consumer lock-free queue

japaric:


XOSplicer:
Hi, great addition to heapless. I just wonder: Is there a particular reason you implemented mpmc::Q* only for specific buffer sizes up to 64? Is there a way to make the implementation more generic?

japaric:
> Is there a way to make the implementation more generic?

Not without losing the `const` constructors, AFAIK.

> Is there a particular reason you implemented mpmc::Q* only for specific buffer sizes up to 64?

No particular reason. If you need a larger capacity feel free to send a PR but note that the capacity needs to be a power of two (see code in the reference).

PR #100, opened 2019-05-24
Remove unecessary conditional

jamesmunns:
In both cases, head is subtracted from tail, wrapping when necessary.


japaric:
Thanks!

bors r+

japaric:
#106 needs to land first and bors got stuck, it seems so ...

bors r-

PR #107, opened 2019-07-12
extend_from_slice

burrbull:
Don't grow length until all slice been copied.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -63,11 +63,14 @@ where
             // won't fit in the `Vec`; don't modify anything and return an error
             Err(())
         } else {
-            for elem in other {
-                unsafe {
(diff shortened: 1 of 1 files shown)


japaric:
The issue with this implementation is that if the `clone` operation panics then all the elements that were inserted midway will get leaked. The leak doesn't happen with the current implementation (because the length is increased on each insertion).

If the goal is improving performance I would suggest looking at the implementation used in [the standard library](https://doc.rust-lang.org/src/alloc/vec.rs.html#1502). It doesn't have the leak problem (it increases the length on each iteration as the current implementation does) but (supposedly) is better optimized by the compiler. I would also recommend benchmarking this operation (for drop and non-drop types) and / or looking at the disassembly if you are concerned about performance.

PR #110, opened 2019-07-27
Add Queue::peek.

jmgao:


XOSplicer:
Hi @jmgao, interesting addition to heapless. 

Could you add some doctests to the new methods?

Furthermore, I am not sure about a return type of `Option<T>`, because of non-copy types. Would it be more correct to return `Option<&T>` ?

jonas-schievink:
Yeah, returning `T` by-value doesn't look correct (without a `T: Copy` bound). Returning an `Option<&mut T>` should also work, though, right? That would be a bit more flexible than returning an immutable reference.

jmgao:
Oops, it hadn't occurred to me that there wasn't already a `T: Copy` bound.

I can return an Option<&mut T>, but it seems to me that that should be called peek_mut or something along those lines. I went with Option<&T> in the latest revision of the patch, do you want me to add peek_mut as well?

src/spsc/mod.rs
```diff
@@ -341,6 +341,38 @@ macro_rules! impl_ {
...
+            /// use heapless::spsc::Queue;
+            /// use heapless::consts::*;
+            ///
+            /// let mut queue: Queue<u8, U235, _> = Queue::u8();
+            /// let (mut producer, mut consumer) = queue.split();
+            /// assert_eq!(None, consumer.peek());
+            /// producer.enqueue(1);
+            /// assert_eq!(Some(&1), consumer.peek());
+            /// assert_eq!(Some(1), consumer.dequeue());
+            /// assert_eq!(None, consumer.peek());
```
jonas-schievink: The example is missing the trailing code fence

src/spsc/split.rs
```diff
@@ -84,6 +84,31 @@ macro_rules! impl_ {
...
+            /// use heapless::spsc::Queue;
+            /// use heapless::consts::*;
+            ///
+            /// let mut queue: Queue<u8, U235, _> = Queue::u8();
+            /// let (mut producer, mut consumer) = queue.split();
+            /// assert_eq!(None, consumer.peek());
+            /// producer.enqueue(1);
+            /// assert_eq!(Some(&1), consumer.peek());
+            /// assert_eq!(Some(1), consumer.dequeue());
+            /// assert_eq!(None, consumer.peek());
```
jonas-schievink: (here too)

src/spsc/mod.rs
```diff
@@ -341,6 +341,38 @@ macro_rules! impl_ {
             N: ArrayLength<T>,
             C: sealed::XCore,
         {
+            /// Returns the item in the front of the queue without dequeuing, or `None` if the queue is empty
```
jonas-schievink:
```suggestion
            /// Returns a reference to the item in the front of the queue without dequeuing, or `None` if the queue is empty.
```

jmgao: Done.

XOSplicer:
@jmgao thanks a lot for coming back and addressing the issues

@japaric could you take a look at this PR and maybe run bors

@bors try

japaric:
Why do the `peek` methods take `&mut self`? That would prevent the following two operations, which seem perfectly fine to me, from compiling:

``` rust
fn foo(q: &mut Queue<i32, U2>) {
    let x = q.peek();
    let y = q.is_empty();
}

fn foo(c: &mut Consumer<i32, U2>) {
    let x = c.peek();
    let y = c.ready();
}
```

The `peek` operation does not mutate the head or tail pointer so it should be allowed through a shared reference (`&self`).

jmgao:
> Why do the `peek` methods take `&mut self`? That would prevent the following two operations, which seem perfectly fine to me, from compiling:

Latest revision switches the methods to `&self`.

japaric:
@jmgao thanks!

(Please next time push new commits into the branch instead of force-pushing. It's hard to see what has changed from revision to revision if you do the latter)

bors r+

jmgao:
> (Please next time push new commits into the branch instead of force-pushing. It's hard to see what has changed from revision to revision if you do the latter)

Apologies, I'm used to review tools like gerrit/phabricator(/pretty much everything that isn't github pull requests) that make it easy to diff between revisions. I'll push separate commits in the future.

PR #117, opened 2019-08-28
Added peek_mut for the binary heap

korken89:
As discussed in https://github.com/japaric/heapless/issues/109, please have a look and lets bike-shed the API.


korken89:
Now using the std lib API

japaric:
Thanks!

bors r+

PR #120, opened 2019-10-02
Allow for manual opt-out of CAS features

jamesmunns:
I've been working on a bare metal platform lately that does not have CAS (specifically ARMv4 bare metal), and this is not caught by the CPU detection logic in the build.rs. This PR adds a flag that allows for manual specification to avoid CAS features.


japaric:
Thanks for the PR and sorry it took me so long to get to it.

Ideally, I would like to use the output of `rustc --print cfg` (see the [`rustc-cfg`] crate) here:

[`rustc-cfg`]: https://crates.io/crates/rustc-cfg

``` console
$ rustc --print cfg thumbv6m-none-eabi
debug_assertions
target_arch="arm"
target_endian="little"
target_env=""
target_feature="mclass"
target_feature="v5te"
target_feature="v6"
target_has_atomic_load_store="16"
target_has_atomic_load_store="32"
target_has_atomic_load_store="8"
target_has_atomic_load_store="ptr"
target_os="none"
target_pointer_width="32"
target_vendor=""

$ rustc --print cfg --target thumbv7m-none-eabi
debug_assertions
target_arch="arm"
target_endian="little"
target_env=""
target_feature="mclass"
target_feature="v5te"
target_feature="v6"
target_feature="v6k"
target_feature="v6t2"
target_feature="v7"
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="8"
target_has_atomic="ptr"
target_has_atomic_load_store="16"
target_has_atomic_load_store="32"
target_has_atomic_load_store="8"
target_has_atomic_load_store="ptr"
target_os="none"
target_pointer_width="32"
target_vendor=""
```

`target_has_atomic` indicates the CAS support; `target_has_atomic_load_store` indicates support only for atomic loads and stores. However, that output seems to be nightly-only for now (I believe there are no immediate plans to stabilize that part of the output):

``` console
$ cargo +beta --print cfg --target thumbv7m-none-eabi
debug_assertions
target_arch="arm"
target_endian="little"
target_env=""
target_os="none"
target_pointer_width="32"
target_vendor=""
```

As that can't be used, using Cargo features is the next best option. However, Cargo features must be additive (\*); meaning that instead of an opt-in `no_cas` feature we should have an opt-out `cas` feature that makes the SPSC API available. The `cas` feature needs to be enabled by default to release it under a new patch version and not break existing users.

(\*) To see what can go wrong with `no_cas` imagine a crate A uses `heapless::spsc::Queue` then in your crate you enable the `no_cas` feature of `heapless` *and* add A as a dependency; `cargo build`-ining your crate will fail to build the A dependency.

jamesmunns:
I've inverted the logic here, and tried to add some helpful diagnostics:

```
cd heapless
cargo build --target thumbv7em-none-eabihf # works
cargo build --no-default-features --target thumbv6m-none-eabi # works
cargo build --target thumbv6m-none-eabi # doesn't work
cargo build --no-default-features --target thumbv6m-none-eabi --features cas # doesn't work
```

src/pool/mod.rs
```diff
@@ -154,7 +154,6 @@
 //!
 //! [1]: https://static.docs.arm.com/ddi0403/eb/DDI0403E_B_armv7m_arm.pdf
 
-#[cfg(not(armv6m))]
```
jamesmunns: Note: This is okay to remove, as the entire pool module is removed when the `cas` feature is not set, and it is not possible to set the `cas` feature with armv6m targets.

jamesmunns:
In this PR, by inverting the features here, I believe this is now a **breaking** change, as any users who were intentionally or unintentionally setting `no-default-features` would lose access to `mpmc` and `pool`.

With the `no_cas` feature, this would not be a breaking change in my opinion.

japaric:
> I believe this is now a breaking change, as any users who were intentionally or unintentionally setting no-default-features would lose access to mpmc and pool.

You are correct. This is in principle a breaking change. However, the crate currently has zero default features so the number of people using `no-default-features` should also be zero as that configuration is a no-op.

Thanks!

bors r+

src/lib.rs
```diff
@@ -90,10 +90,13 @@ mod ser;
 
 pub mod binary_heap;
 pub mod i;
-#[cfg(not(armv6m))]
+#[cfg(feature = "cas")]
 pub mod mpmc;
-#[cfg(not(armv6m))]
```
japaric: Right, so we actually need to keep this `not` as: `cfg(all(not(armv6m), feature = "cas"))` to keep `cargo build --target thumbv6m-none-eabi` working.

src/lib.rs
```diff
@@ -90,10 +90,13 @@ mod ser;
...
+#[cfg(feature = "cas")]
 pub mod pool;
 pub mod spsc;
 
+#[cfg(all(armv6m, feature = "cas"))]
+core::compile_error!("The 'cas' feature may not be used with `armv6m` targets.");
```
japaric: And we have to drop this

japaric: Basically, the state of the "cas" feature should have no effect on ARMv6-M.

src/lib.rs
jamesmunns: Is that the interface you would like (implicitly masking the features), rather than forcing the user to opt-out with `no-default-features`?

I mentioned this in [this comment](https://github.com/japaric/heapless/pull/120#issuecomment-566554281).

japaric: To get this PR into a new patch version we need to keep `cargo build --target thumbv6m-none-eabi` (note: "cas" is implicitly enabled) working so the logic needs to make the "cas" feature a no-op on that target.

jamesmunns: Happy to make the change, though I might suggest that we change the "implicit no-op" behavior on the next breaking (minor) release.

jamesmunns:
@japaric back to you.

src/lib.rs
japaric: And probably make "cas" opt-in as well; I don't like having a crate that by default doesn't build for all targets. Could you make an issue to make sure we don't forget? I'll put on the v0.6.0 milestone.

japaric:
Thanks!

bors r+

jamesmunns:
@japaric I missed you added a feature at the end of the file in my merge. This should work now.

PR #121, opened 2019-10-04
Bugfix in mpmc queue

KKuznecov:
This update prevents the entry into an infinite loop after calling Q_::dequeue() under certain conditions. The following code leads to an infinite loop:
```rust
#[test]
    fn blocking(){
        let q = Q2::new();
        for _ in 0..255 {
            q.enqueue(0).unwrap();
            q.dequeue();
        }
        q.dequeue();
    }
```


XOSplicer:
Hi @KKuznecov, thanks for pointing Out this bug and providing a fix already.

It would be great if the test case you provided it added to the unit tests for the mpmc module.

After bors passes, this PR is good to merge.

japaric:
Thanks for the PR; I'm merging this as PR #132, which includes a regression test.

PR #125, opened 2019-11-18
Deserialize maps as maps

dflemstr:
The maps are already being serialized as maps, and serde_json will detect
that despite us claiming that maps are seqs, it falls back to parsing them
as maps; however, serde-json-core does not have that fallback logic, so it
makes more sense to give the correct hint here (especially when this is
merged: https://github.com/japaric/serde-json-core/pull/23)

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -182,7 +182,7 @@ where
                 Ok(values)
             }
         }
-        deserializer.deserialize_seq(ValueVisitor(PhantomData))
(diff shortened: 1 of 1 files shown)


japaric:
Thanks for the PR. I suppose this is technically a breaking change but since serde-json can deserialize back what you get from serializing a map I don't think anyone was relying on the current behavior.

bors r+

PR #126, opened 2019-11-27
Add DoubleEndedIterator to queue iterators

david-sawatzke:
The implementation is pretty straightforward and it's useful


japaric:
Thanks!

bors r+

PR #129, opened 2019-11-29
Fix improper handling of overflow in `len_usize` (fix #128)

### LINKED ISSUE #128: Queue Iter loops [nearly] endlessy (and probably unsound behaviour)
Reproduction:
Add test in queue tests
```
    #[test]
    fn iter_endless() {
        let mut rb: Queue<i32, U4, u8> = Queue::u8();
        rb.enqueue(0).unwrap();
        for i in 0..300 {
            print!(
                "Iteration {}, {:?}, {:?}",
                i,
                rb.0.head.load_relaxed(),
                rb.0.tail.load_relaxed()
            );
            let mut items = rb.iter_mut();
            println!(" {} {}", items.index, items.len);
            assert_eq!(items.next(), Some(&mut 0));
            assert_eq!(items.next(), None);
            rb.dequeue().unwrap();
            rb.enqueue(0).unwrap();
        }
    }
```
Execute: 
```
cargo test -- --nocapture
```
Output: 
```
...
Iteration 253, 253, 254 0 1
Iteration 254, 254, 255 0 1
Iteration 255, 255, 0 0 18446744073709551361
...
```
Then stop cargo, since the drop implementation relies on correct iterator behaviour.
(This is then obviously "really bad", since it tries to drop something multiple times)

david-sawatzke:
This will lead to implausibly large lengths, nearly endless iterators
and possibly unsound behaviour through the Drop implementation

Switches the `truncate` method to refer to the common defintion of
simply chopping of the test bits. (Implemented by `as`, which is defined
to do this).

Adds a new `saturate` method to replace the previous `truncate` method

### DIFF
--- a/src/sealed.rs
+++ b/src/sealed.rs
@@ -22,6 +22,9 @@ unsafe impl XCore for MultiCore {
 
 pub unsafe trait Uxx: Into<usize> + Send {
     #[doc(hidden)]
+    fn saturate(x: usize) -> Self;
+
+    #[doc(hidden)]
     fn truncate(x: usize) -> Self;
 
     #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


david-sawatzke:
@japaric Could you look over this? I think this is a serious issue

japaric:
Sorry for taking so long to get to this.

Thanks for the PR!

japaric:
bors retry

PR #136, opened 2019-12-18
add the HistoryBuffer type

### LINKED ISSUE #112: "Classic" ringbuffer support?
I've needed a "classic" ringbuffer quite a few times now, and since it's not that hard rolled my own.  What I mean is:

* fixed capacity array plus index of "head"
* push() moves head and overwrites
* allows random access using iterator, and some standard things like first(), last()
* (maybe:) all items are initially filled with some default, allows getting rid of Option<>s for first() etc
* normal `&mut` methods, no lock-free support

This is nothing fancy, but seems like it falls under the mission statement of `heapless`?  I'll contribute an initial implementation PR, if yes.

birkenfeld:
Closes #112

Unfortunately `const` construction won't work since the `Default` or `Clone` bound on `T` can't be used then, so I've not added an internal type to `i.rs`.

Let me know if you think any other traits should be implemented...

### DIFF
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -1,11 +1,10 @@
 use generic_array::{ArrayLength, GenericArray, sequence::GenericSequence};
-use as_slice::{AsSlice, AsMutSlice};
 
-/// A "history buffer", similar to a write-only ring buffer.
+/// A "history buffer", similar to a write-only ring buffer of fixed length.
 ///
-/// This buffer keeps a fixed number of elements.  On push, the oldest element
-/// is overwritten. It is useful to keep a history of values with some desired
-/// depth, and for example calculate a rolling average.
+/// This buffer keeps a fixed number of elements.  On write, the oldest element
+/// is overwritten. Thus, the buffer is useful to keep a history of values with
+/// some desired depth, and for example calculate a rolling average.
 ///
 /// The buffer is always fully initialized; depending on the constructor, the
 /// initial value is either the default value for the element type or a supplied
@@ -20,18 +19,18 @@ use as_slice::{AsSlice, AsMutSlice};
 /// // Initialize a new buffer with 8 elements, all initially zero.
 /// let mut buf = HistoryBuffer::<_, U8>::new();
 ///
-/// buf.push(3);
-/// buf.push(5);
+/// buf.write(3);
+/// buf.write(5);
 /// buf.extend(&[4, 4]);
 ///
-/// // The first (oldest) element is still zero.
-/// assert_eq!(buf.first(), &0);
-/// // The last (newest) element is a four.
-/// assert_eq!(buf.last(), &4);
-/// for el in buf.iter() { println!("{:?}", el); }
+/// // The most recent written element is a four.
+/// assert_eq!(buf.recent(), &4);
+///
+/// // To access all elements in an unspecified order, use `as_slice()`.
+/// for el in buf.as_slice() { println!("{:?}", el); }
 ///
 /// // Now we can prepare an average of all values, which comes out to 2.
-/// let avg = buf.iter().sum::<usize>() / buf.len();
+/// let avg = buf.as_slice().iter().sum::<usize>() / buf.len();
 /// assert_eq!(avg, 2);
 /// ```
 #[derive(Clone)]
@@ -63,7 +62,7 @@ where
     /// // Allocate a 16-element buffer on the stack
     /// let mut x: HistoryBuffer<u8, U16> = HistoryBuffer::new();
     /// // All elements are zero
-    /// assert!(x.iter().eq([0; 16].iter()));
+    /// assert_eq!(x.as_slice(), [0; 16]);
     /// ```
     pub fn new() -> Self {
         Self {
@@ -95,7 +94,7 @@ where
     /// // Allocate a 16-element buffer on the stack
     /// let mut x: HistoryBuffer<u8, U16> = HistoryBuffer::new_with(4);
     /// // All elements are four
-    /// assert!(x.iter().eq([4; 16].iter()));
+    /// assert_eq!(x.as_slice(), [4; 16]);
     /// ```
     pub fn new_with(t: T) -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


src/histbuf.rs
```diff
@@ -0,0 +1,325 @@
...
+{
+    fn extend<I>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = &'a T>,
+    {
+        self.extend(iter.into_iter().cloned())
+    }
+}
+
+impl<T, N> AsSlice for HistoryBuffer<T, N>
```
japaric: Could you add `as_slice` and `as_mut_slice` as inherent methods so that `hb.as_slice()` works *without* having to import the `AsSlice` trait?

(inherent methods as in `impl HistoryBuffer { fn as_slice(&self) -> &[T] { .. } }`

src/histbuf.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+impl<T, N> HistoryBuffer<T, N>
+where
+    N: ArrayLength<T>,
+{
+    /// Returns the length of the buffer, which is the length of the underlying
+    /// backing array.
+    pub fn len(&self) -> usize {
```
japaric: Q: is there any use case for this returning not the capacity but the number of elements pushed in the buffer after clearing it (or starting from a cleared buffer)? As in `HIstoryBuffer::new().len()` should return 0.

src/histbuf.rs
```diff
@@ -0,0 +1,325 @@
...
+        let x: HistoryBuffer<u8, U4> = HistoryBuffer::new();
+        assert!(x.iter().eq([0; 4].iter()));
+    }
+
+    #[test]
+    fn push_iter() {
+        let mut x: HistoryBuffer<u8, U4> = HistoryBuffer::new();
+        x.push(1);
+        x.push(4);
+        assert!(x.iter().eq([0, 0, 1, 4].iter()));
```
japaric: Related to my previous question about `len` returning not always the capacity, I guess. Would someone expect this to iterate from the oldest pushed element to the newest pushed element, *not* including the not-pushed elements (i.e. the elements that have the `Default` value)? e.g. this iteration should yield `1` and then `4` and then `None`. Or to word this differently: I didn't expect this iteration to start with the unpushed / cleared values.

src/histbuf.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        x.push(5);
+        x.push(6);
+        x.push(10);
+        assert!(x.iter().eq([4, 5, 6, 10].iter()));
```
japaric: This iteration order, on the other hand, meets my expectations.

src/histbuf.rs
```diff
@@ -0,0 +1,325 @@
...
+        x.clear_with(1);
+        assert!(x.iter().eq([1; 4].iter()));
+    }
+
+    #[test]
+    fn first_last() {
+        let mut x: HistoryBuffer<u8, U4> = HistoryBuffer::new();
+        x.push(1);
+        x.push(4);
+        assert_eq!(x.first(), &0);
```
japaric: Again, I would have expected this to return `&1` but perhaps it's OK for the use cases of this data structure.

japaric: Thinking about this a bit more. `first` and `last` may not always be true because one can do something like `hb.as_mut_slice().sort()` or reorder the elements arbitrarily making `first` and `last` not return the original insertion order. We may want to point this out in the docs.

src/histbuf.rs
birkenfeld: Sure!

src/histbuf.rs
birkenfeld: Indeed; either the "mut slice" method can be removed or this has to be documented.

src/histbuf.rs
birkenfeld: Well, my use case is a buffer that once created is filled continuously with new values, so the "len < capacity" state is only observed for a very short time.

With the semantics that `len()` can be smaller than capacity, you also need to have basically full Vec-like semantics, i.e. `first()` and `last()` returning option, which I wanted to avoid. But it may be cleaner to do it that way.

(In that case, the implementation can probably use a `Vec` internally.)

japaric:
> Well, my use case is a buffer that once created is filled continuously with new values, so the "len < capacity" state is only observed for a very short time.

I think we should start with supporting only this use case where the 'len < capacity' state is never relevant. To do that we should remove `first`, `last`, `iter`, `iter_mut` and `len` (or rename this one to `capacity` to match what our other data structures do) since they create ambiguity about iteration order and whether a not-yet-written value should be first or last.

If people want to work with the HistoryBuffer in the 'len < capacity' state we can, in the future, either extend the API to what was in the first state of this PR but with the extra runtime overhead to track how many elements have actually been written, or add a *second* HistoryBuffer that provides that API and semantics (since it has an extra runtime cost).

birkenfeld:
> I think we should start with supporting only this use case where the 'len < capacity' state is never relevant. To do that we should remove `first`, `last`, `iter`, `iter_mut` and `len` (or rename this one to `capacity` to match what our other data structures do) since they create ambiguity about iteration order and whether a not-yet-written value should be first or last.

I'm mostly happy to do that, but I'd be sad to lose `last`, it can be very useful to decide what to do with a new data point. Of course it can also be stored in another variable.

japaric:
> I'd be sad to lose `last`

By "last" you mean the most *recently* inserted item? (I forget which end is first and which is last). A `recent` method that returns (by reference) either the most recently inserted item, or "whatever value" if the buffer was recently cleared, seems fine to me. Alternatively, we could make `push` return a reference to the item that was just inserted.

---

I wonder if it would be better to rename `push` to `write` as there's no `pop` counterpart and the data structure was presented as a "write-only ring buffer" in the docs. Also, we don't really track the "length" of the buffer when you write something new so perhaps `len` should be kept with that name and the documentation should say that the data structure has *fixed length* -- it's closer to an array than to a vector.

burrbull:
> I wonder if it would be better to rename `push` to `write` as there's no `pop` counterpart

or `append`

birkenfeld:
Updated with the results of discussion.

I chose `write` instead of `append` because the latter is used to add multiple elements to a vector in Rust, not just one.

korken89:
It seems to me all comments have been addressed

PR #140, opened 2020-01-15
extend singleton pool! support to ARMv7-A

japaric:


japaric:
I've verified that the Cortex-A machine code uses STREX + LDREX to implement CAS loops

bors r+

PR #143, opened 2020-02-11
Implement core::fmt::Write for heapless::Vec<u8, N>

ijl:
I didn't see a mention of this in issues or previous pull requests, but I expect it's in scope?

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -1003,6 +1003,7 @@ mod tests {
         assert_eq!(&v[..], b"4d2");
     }
 
+    #[test]
     fn extend_from_slice() {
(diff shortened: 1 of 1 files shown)


XOSplicer:
core::fmt::Write is implemented for heapless::String aswell, so I guess the implementation for Vec<u8> makes sense. String implements write_str() aswll as write_char(). Should both be implemented for Vec<u8> ?

ijl:
I happened to only need this, and it's the only required part, but I could.

korken89:
Hi could you add this for string as well?

MathiasKoch:
@korken89 I think it is already implemented for String?

korken89:
bors retry

PR #144, opened 2020-02-15
mpmc: use i8 to calculate dif to deal with u8 wrapping

debug-ito:
This p-r adds a new test case `full_at_wrapped_pos0`, which went into infinite loop. This p-r fixes the problem by using `i8` (instead of `i16`) to calculate the `dif` (`seq - pos` in `enqueue`, `seq - (pos + 1)` in `dequeue`).

The root cause of this problem was basically the same as #121: when the `pos` goes beyond 255, it wraps back to 0 because `pos` is in `u8`. When the `u8` wrapping occurs, the sign of `dif` becomes inverse, which caused the problem.

Using `i8` for `dif` fixes the problem, because even when the `u8` wrapping occurs, the `dif` is wrapped again in `i8`, so we get the right sign for `dif`. `i8` only allows values in [-127, 128], but `|dif|` is at most the size of the queue. The size of the queue is at most 64 in this library, so it's no problem.

This p-r removes the fix introduced by #121, because the fix in this p-r also fixes the problem reported in it.

### DIFF
--- a/src/mpmc.rs
+++ b/src/mpmc.rs
@@ -571,8 +571,8 @@ mod tests {
     fn drain_at_pos255() {
         let q = Q2::new();
         for _ in 0..255 {
-            q.enqueue(0).unwrap();
-            q.dequeue();
+            assert!(q.enqueue(0).is_ok());
+            assert_eq!(q.dequeue(), Some(0));
         }
         // this should not block forever
         assert_eq!(q.dequeue(), None);
--- a/src/mpmc.rs
+++ b/src/mpmc.rs
@@ -479,7 +479,7 @@ unsafe fn dequeue<T>(buffer: *mut Cell<T>, dequeue_pos: &AtomicU8, mask: u8) -> 
     loop {
         cell = buffer.add(usize::from(pos & mask));
         let seq = (*cell).sequence.load(Ordering::Acquire);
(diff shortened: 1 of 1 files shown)


debug-ito:
Do you have time to review this, @japaric ?

korken89:
This LGTM, any comments @japaric ?

korken89:
Ping @japaric

korken89:
bors retry

korken89:
Hi, the `mpmc_contention` test seems to hang. Can @debug-ito or @japaric have a look at this?

debug-ito:
I saw the test code and its result, but I have no idea about the reason for the error. I tried to run the Travis-CI test on my forked respository, but somehow I could not enable CI.

Where in `mpmc_contention` did the test stop? Does the error persist if we revert the changes I made?

korken89:
@debug-ito Could you rebase this on master to get the latest CI running?
A bit of a thorn in my back that we have not gotten this merged yet.

debug-ito:
Thanks for taking care of this p-r! I rebased the branch to the master and force-pushed it.

korken89:
The tests are passing, could you run formatting? Then we can merge :)

debug-ito:
Thanks. I ran cargo fmt. It passed the CI.

korken89:
bors merge

PR #145, opened 2020-03-24
Made Box StableDeref

korken89:


korken89:
> Nice. Thanks!
> 
> It may be worth considering switching things to the standard `Unpin` in future releases (of this crate and others)

Yeah, this would make sense

korken89:
bors r=japaric

PR #149, opened 2020-04-12
Make Producer<..., SingleCore> Send, like Consumer

BryanKadzban:
I assume there's no reason that only multi-core Producers are marked
Send, while Consumers are not (they're pretty symmetrical).

(Context: I'm trying to use an SPSC queue on a single cortex-m0 MCU.
So I can guarantee that a SingleCore queue is safe because there is no
other core it could possibly be shared with.  The RTFM framework adds
an assert_send on the producer, since it's shared between init and an
SPI receiver interrupt.  It also adds an assert_send on the Consumer,
since it's shared between init and the I2C transmit interrupt, but that one
works since Consumer has this extra C type in the impl already.)

### DIFF
--- a/src/spsc/split.rs
+++ b/src/spsc/split.rs
@@ -61,11 +61,12 @@ where
     _marker: PhantomData<&'a ()>,
 }
 
-unsafe impl<'a, T, N, U> Send for Producer<'a, T, N, U>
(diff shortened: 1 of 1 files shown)


korken89:
@japaric Can you have a look here?

PR #151, opened 2020-04-17
Implement Vec::from_slice

### LINKED ISSUE #150: Vec: Simpler creation from slice
Right now creating a Vec with fixed data can be a bit verbose...

```rust
let mut data = Vec::<u8, consts::U256>::new();
data.extend_from_slice(b"Hello").unwrap();
```

Would you accept a PR that adds a constructor function?

```rust
let data = Vec::<u8, consts::U256>::from_slice(b"Hello").unwrap();
```

Besides requiring less code, it also allows to drop the `mut`.

dbrgn:
Fixes #150.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -197,6 +197,28 @@ where
         Vec(crate::i::Vec::new())
     }
 
+    /// Constructs a new vector with a fixed capacity of `N` and fills it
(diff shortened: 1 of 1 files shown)


dbrgn:
Sure, done!

PR #152, opened 2020-04-17
Implement Vec::starts_with and Vec::ends_with

dbrgn:
Implementation mostly copied from `std`.

The logic should be covered by documentation tests, so no separate unit tests were added.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -366,6 +366,54 @@ where
     pub(crate) fn is_full(&self) -> bool {
         self.0.is_full()
     }
+
+    /// Returns `true` if `needle` is a prefix of the Vec.
+    ///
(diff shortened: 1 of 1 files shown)


dbrgn:
@korken89 I marked the functions as inline.

However, regarding the tests, `cargo test` will already run that code as doctests and at the same time it works as an example on how to use it. Or am I missing something? Would you like the example code to be smaller?

korken89:
It's more that asserts are not really easy to understand example for users.
I'd like to see a more concise example, like add 3 values and compare with the 2 fist and 2 last, while a more complete set of asserts be used as tests.

I hope you see what I mean :)

dbrgn:
Sure! Updated now.

MathiasKoch:
ping @korken89

PR #153, opened 2020-04-20
x86_64: "practically" thread-safe  Pool

japaric:
### Summary / motivation

this PR makes `Pool` `Sync` on x86_64 with the main goal of being able to test `no_std` code that uses `Pool` (e.g. via the singleton `pool!` macro) on x86_64 (as it's not straight forward to run tests on embedded devices, and even less to CI those tests).

### Details

this PR reduces the chance of Pool running into the ABA problem (which corrupts the underlying lock-free stack) by using 32-bit version tags (search term: IBM ABA-prevention tags). Version tags do not 100% prevent the ABA problem but make it almost impossible to run into it in practice (the bigger the tag, in bits, the less the chance of running into ABA). See module level docs for details and limitations.

As this does not eliminate ABA with 100% certainty perhaps it should live behind a Cargo feature?

It seems to me that hazard pointers may be able to completely get rid of ABA but implementing those, AFAICT, require Thread Local Storage (`#[thread_local]`) which is not supported in `no_std` code.

r? @korken89

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,7 +73,8 @@
 #![deny(missing_docs)]
 #![deny(rust_2018_compatibility)]
 #![deny(rust_2018_idioms)]
-#![deny(warnings)]
+// #![deny(warnings)]
+#![allow(warnings)] // FIXME
 
 pub use binary_heap::BinaryHeap;
 pub use generic_array::typenum::consts;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,8 +73,7 @@
 #![deny(missing_docs)]
 #![deny(rust_2018_compatibility)]
 #![deny(rust_2018_idioms)]
-// #![deny(warnings)]
-#![allow(warnings)] // FIXME
+#![deny(warnings)]
 
 pub use binary_heap::BinaryHeap;
 pub use generic_array::typenum::consts;
--- a/src/pool/cas.rs
+++ b/src/pool/cas.rs
@@ -1,0 +1,212 @@
+//! Stack based on CAS atomics
+//!
+//! To reduce the chance of hitting the ABA problem we use a 32-bit offset + a 32-bit version tag
+//! instead of a 64-bit pointer. The version tag will be bumped on each successful `pop` operation.
+
+use core::{
+    cell::UnsafeCell,
+    convert::TryFrom,
+    marker::PhantomData,
+    mem::{self, MaybeUninit},
+    num::NonZeroUsize,
+    ptr::NonNull,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+/// Unfortunate implementation detail required to use the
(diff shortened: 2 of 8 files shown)


src/lib.rs
```diff
@@ -65,7 +65,8 @@
 #![deny(missing_docs)]
 #![deny(rust_2018_compatibility)]
 #![deny(rust_2018_idioms)]
-#![deny(warnings)]
+// #![deny(warnings)]
+#![allow(warnings)] // FIXME
```
korken89, line 69: Should this be fixed?

src/pool/cas.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub(crate) data: UnsafeCell<T>,
+}
+
+impl<T> Node<T> {
+    fn next(&self) -> &Atomic<Node<T>> {
+        &self.next
+    }
+}
+
+pub struct Stack<T> {
```
korken89, line 29: Overall for this module documentation is missing

src/pool/llsc.rs
```diff
@@ -0,0 +1,76 @@
+//! Stack based on LL/SC atomics
+
+pub use core::ptr::NonNull as Ptr;
+use core::{
+    cell::UnsafeCell,
+    ptr,
+    sync::atomic::{self, AtomicPtr, Ordering},
+};
+
+pub struct Node<T> {
```
korken89, line 10: Overall documentation is missing for this module

japaric:
@korken89 fixed the warnings (not everything needed to be documented so I only document what was generating a warning). Do you have any thoughts on putting this behind a Cargo feature?

korken89:
Awesome! 
Hmm, as it is only for testing maybe just put it behind `#[cfg(test)]`? Else is that is not possible a cargo feature would be good.
I don't think this should be part of the official (semver) API?

japaric:
> Hmm, as it is only for testing maybe just put it behind #[cfg(test)]? 

yeah, I don't think that would work. `cargo test` doesn't compile the whole dependency tree with `--cfg test`; it only does that for the crates in the current package.

> Else is that is not possible a cargo feature would be good.

I'll add a Cargo feature.

> I don't think this should be part of the official (semver) API?

by "semver" you mean that this Cargo feature could be removed in a new patch release? or something else?

japaric:
(added Cargo feature and rebased)

korken89:
> by "semver" you mean that this Cargo feature could be removed in a new patch release? 

Yeah, exactly. As it is only for internal testing I mean.

korken89:
@japaric It seems you are using `ptr::cast` which is unstable, could you change this one out?

japaric:
@korken89 done, latest commit should fix compatibility with the MSRV

korken89:
Awesome, thanks!

PR #154, opened 2020-04-22
indexmap: expose PowerOfTwo, Bucket and Pos

willem66745:
This change allows the create custom structs based on `IndexMap` with size arguments outside
the heapless crate itself.

For example:
```rust
use heapless::consts::*;
use heapless::{ArrayLength, Bucket, FnvIndexMap, Pos, PowerOfTwo};

struct CustomMap<S>
where
    S: ArrayLength<Bucket<u8, u16>> + ArrayLength<Option<Pos>> + PowerOfTwo,
{
    map: FnvIndexMap<u8, u16, S>,
}

impl<S> CustomMap<S>
where
    S: ArrayLength<Bucket<u8, u16>> + ArrayLength<Option<Pos>> + PowerOfTwo,
{
    fn new() -> CustomMap<S> {
        CustomMap {
            map: FnvIndexMap::<_, _, S>::new(),
        }
    }
}

fn main() {
    let mut bla = CustomMap::<U8>::new();

    bla.map.insert(8, 16).unwrap();
}
```

I can imagine that exposing these types is not preferable, but at least it would help for my usecase. Or is there a way where exposing these types is not needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,9 +68,9 @@
 #![deny(warnings)]
 
 pub use binary_heap::BinaryHeap;
-pub use generic_array::typenum::consts;
+pub use generic_array::typenum::{consts, PowerOfTwo};
 pub use generic_array::ArrayLength;
-pub use indexmap::{FnvIndexMap, IndexMap};
+pub use indexmap::{Bucket, FnvIndexMap, IndexMap, Pos};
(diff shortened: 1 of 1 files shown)


korken89:
I am not 100% sure on the design-though behind this, could you have a look @japaric ?

PR #170, opened 2020-07-24
Buffer methods for Vec and as_mut_vec for String

samlich:
This adds `set_len` to `Vec`, identical to the `std` version, with the documentation copied from `std`. This is useful for using it as a buffer by writing to the uninitialized portion and then setting the length to include those bytes.

It also adds `as_mut_vec` to `String` (also from `std`), which enables using `String` as a buffer in the same manner.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -178,6 +178,34 @@ where
         unsafe { str::from_utf8_unchecked_mut(self.0.vec.as_mut_slice()) }
     }
 
+    /// Returns a mutable reference to the contents of this `String`.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not check that the bytes passed
+    /// to it are valid UTF-8. If this constraint is violated, it may cause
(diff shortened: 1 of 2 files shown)


src/string.rs
```diff
@@ -178,6 +178,36 @@ where
...
+    /// unsafe {
+    ///     let vec = s.as_mut_vec();
+    ///     assert_eq!(&[104, 101, 108, 108, 111][..], &vec[..]);
+    ///
+    ///     vec.reverse();
+    /// }
+    /// assert_eq!(s, "olleh");
+    /// ```
+    pub unsafe fn as_mut_vec(&mut self) -> &mut Vec<u8, N> {
+        core::mem::transmute::<&mut crate::i::Vec<GenericArray<u8, N>>, &mut Vec<u8, N>>(
```
japaric, line 206: Please change this to a raw pointer cast followed by dereference like this:

``` rust
unsafe { &mut *(self.0.vec as *mut SourceType as *mut TargetType) }
```

Issue #171, opened 2020-08-26
Defmt support


theunkn0wn1:
I recently discovered the [`defmt`](https://github.com/knurling-rs/defmt)  library and attempted to integrate it into one of my embedded projects.

Unfortunately it appears that `heapless::Vec` doesn't implement the `defmt::Format` trait yet, which blocks `defmt` integration in my downstream. 

Feature request: support for `defmt::Format` in `heapless::Vec`

korken89:
Hi, I have commented on the PR itself.

korken89:
Available in 0.7.1

mehmetalianil:
I get `error: unknown display hint: "[u8]"`  from:
```rust
let mut test_vec: Vec<u8, 128> = Vec::new();
   ... 
info!("the vector: {:[u8]}", test_vec);
```

Am I mistaken with the format specifier?

korken89:
Have you enabled the feature flag?

theunkn0wn1:
In my testing i used `{:?}` and `{=[?]}`, which both compile.

```rs
        let mut data: Vec<u32, 16> = Vec::new();
        data.push(0x42).expect("Failed to push into vec...");
        data.push(0x50).expect("Failed to push into vec...");
        data.push(0x10).expect("Failed to push into vec...");

        defmt::debug!("Writing data....");
        defmt::info!("{=[?]}", data);
        defmt::debug!("populating string...");
        let string: String<32> = String::from("I like pizza");
        defmt::debug!("Emitting string...");
        defmt::info!("{:?}", string);

```

PR #172, opened 2020-08-27
defmt impl for Vec

theunkn0wn1:
This PR adds support for encoding `heapless::Vec` for usage with `defmt`.
`defmt` can already handle slices, so this impl simply invokes that encoder method.

I chose to use the slice implementation is I did not perceive the difference between a Vec and a Slice to be of specific importance in a log file, but rather the contents of the vec / slice to be of interest. It also appeared to me as the path of least resistance.

I am marking this as a draft as I have yet to be able to validate this against my local hardware, `probe-run` is giving me some grief that I am reasonably sure is unrelated to this change set. 

Relates to #171

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ x86-sync-pool = []
 __trybuild = []
 # Enable larger MPMC sizes.
 mpmc_large = []
+defmt-impl = ["defmt"]
 
 [target.x86_64-unknown-linux-gnu.dev-dependencies]
 scoped_threadpool = "0.1.8"
@@ -49,3 +50,12 @@ optional = true
 
 [dev-dependencies.ufmt]
 version = "0.1"
+
+[dependencies.defmt]
+version = "0.2.1"
+optional = true
+
+[dev-dependencies.defmt]
+version = "0.2.1"
+features = ["unstable-test"]
+
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,8 @@ mod de;
 mod ser;
 
 pub mod binary_heap;
+#[cfg(feature = "defmt-impl")]
+mod defmt;
 #[cfg(all(has_cas, feature = "cas"))]
 pub mod mpmc;
 #[cfg(all(has_cas, feature = "cas"))]
--- a/src/defmt.rs
+++ b/src/defmt.rs
@@ -5,8 +5,8 @@ use crate::Vec;
 use defmt::Formatter;
 
 impl<T, const N: usize> defmt::Format for Vec<T, N>
-    where
-        T: defmt::Format,
+where
+    T: defmt::Format,
 {
     fn format(&self, fmt: Formatter<'_>) {
         defmt::write!(fmt, "{=[?]}", self.as_slice())
@@ -14,8 +14,8 @@ impl<T, const N: usize> defmt::Format for Vec<T, N>
 }
 
 impl<const N: usize> defmt::Format for crate::String<N>
-    where
-        u8: defmt::Format,
+where
+    u8: defmt::Format,
 {
     fn format(&self, fmt: Formatter<'_>) {
         defmt::write!(fmt, "{=str}", self.as_str());
--- a/src/defmt.rs
+++ b/src/defmt.rs
@@ -1,0 +1,56 @@
+//! Defmt implementations for heapless types
+//!
+
+use crate::Vec;
+use defmt::Formatter;
+
+impl<T, const N: usize> defmt::Format for Vec<T, N>
+    where
+        T: defmt::Format,
+{
+    fn format(&self, fmt: Formatter<'_>) {
+        defmt::write!(fmt, "{=[?]}", self.as_slice())
+    }
+}
+
+impl<const N: usize> defmt::Format for crate::String<N>
+    where
+        u8: defmt::Format,
+{
+    fn format(&self, fmt: Formatter<'_>) {
+        defmt::write!(fmt, "{=str}", self.as_str());
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::Vec;
+    use defmt::Format;
+
+    #[test]
+    /// Tests encoding Vec with defmt, asserting these types may be serialized
+    /// Note: the exact wire format is NOT checked since its an unstable implementation detail of an external crate.
+    /// based on https://github.com/knurling-rs/defmt/blob/697a8e807bd766a80ada2d57514a9da1232dbc9a/tests/encode.rs#L523
+    fn test_defmt_format_vec() {
+        let val: Vec<_, 8> = Vec::from_slice(b"abc").unwrap();
+
+        let mut f = defmt::InternalFormatter::new();
+        let g = defmt::Formatter { inner: &mut f };
+        val.format(g);
+        f.finalize();
+    }
+
+    /// Tests encoding String with defmt, asserting these types may be serialized
+    /// Note: the exact wire format is NOT checked since its an unstable implementation detail of an external crate.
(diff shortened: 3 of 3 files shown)


src/lib.rs
```diff
@@ -106,6 +106,8 @@ pub mod mpmc;
 pub mod pool;
 #[cfg(has_atomics)]
 pub mod spsc;
+#[cfg(feature = "defmt-impl")]
+pub mod defmt;
```
birkenfeld, line 110: I don't think this needs to be `pub`.

korken89:
Hi, thanks for the PR!

Since you have started this, do you feel like adding support for `String` as well?

theunkn0wn1:
I think that can be arranged if that is desirable.

I have largely been blocked by the [probe-rs wfi bug](https://github.com/probe-rs/probe-rs/issues/350) making doing anything with RTT painful, so by extension this includes defmt.

theunkn0wn1:
@korken89 
Looked a little into implementing `String` with defmt, and don't see a clear path forward.

In the case of Vec, I leveraged the existing slice impl because the distinction wasn't important.
In the case of a string, having heapless emit output a slice of bytes isn't as helpful as getting  a string out of it.

If i understand what defmt is doing correctly, for strings they are compile-time interned and arn't actually sent over-the-wire. I suspect defmt changes may need to be implemented for runtime strings such as those produced by heapless. 

At this point I offer this PR as-is. Ive done some tests working around the WFI bug and this PR appears to work correctly. Further the unit test I added is modeled after existing tests in defmt, so I have some confidence my PR works as intended. If `String` support is desired, I wish for that to be in its own ticket and out of scope for this pull request.

birkenfeld:
> If i understand what defmt is doing correctly, for strings they are compile-time interned and arn't actually sent over-the-wire.

Well that's for format strings and other constant strings.  But there are also primitive types "byte slice" and "string slice" which are sent as-is (see https://defmt.ferrous-systems.com/primitives.html).  I think you just need to call `fmt.str()` on the Formatter and that should be all.

theunkn0wn1:
> > If i understand what defmt is doing correctly, for strings they are compile-time interned and arn't actually sent over-the-wire.
> 
> I think you just need to call `fmt.str()` on the Formatter and that should be all.

Wow I completely missed that! You may be right, il do some investigating.

theunkn0wn1:
Tested both `String` and `Vec` formatting on my stm32f4 while working around wfi.

```
     Running `probe-run --defmt --chip STM32F446RETx target/thumbv7em-none-eabihf/debug/rust_stm32_motor`
  (HOST) INFO  flashing program
  (HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
0.000000 WARN  hello, world!
└─ rust_stm32_motor::init @ src/main.rs:99
0.000001 DEBUG msg: hello, heapless!
└─ rust_stm32_motor::init @ src/main.rs:102
...
0.000121 DEBUG emitting [1, 7, 130, 125, 18, 132, 1, 32, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0]
└─ rust_stm32_motor::uart4_on_rxne @ src/main.rs:321
0.000122 TRACE done processing packet. clearing buffer.
```

korken89:
Thanks for the update!

Now we just need to await `defmt` to be on crates.io and this is ready for inclusion.
Maybe @japaric knows when it will be released?

korken89:
Or, it is behind a feature-flag so I see no issue in including as is.
This feature will be considered unstable until a proper release of `defmt` is available.

korken89 (approved):

korken89:
bors merge

korken89:
bors retry

korken89:
Hi @theunkn0wn1, the CI seems to be failing for 1.36.
Could you have a look into why?

theunkn0wn1:
Wilco. Yes it's behind a feature flag. As ci hasn't executed when I opened the pr I'm guessing I need to add the feature flag to the ci run.

theunkn0wn1:
@korken89 
Looks like defmt depends on the `profile-overrides` feature, which isn't stable until `1.41`.
`1.41 > 1.36` so that may appears to be why. Can confirm build failure on `rust:1.36` but note build success on `1.41`

theunkn0wn1:
@thalesfragoso commented in matrix that depending on git dependencies would make this crate, to my understanding [unreleasable on crates.io](https://github.com/rust-lang/cargo/issues/6738). Until the time `defmt` gets a crates.io release, this issue will persist. 

Ultimately I do not see how the CI failure can be addressed without boosting this crates MSRV to at least rust stable `1.41`, or having the MSRV of `defmt` lowered to `1.36`.

korken89:
This crate has a quite low MSRV, and it seems like it starts to be time to look it over.
I am not sure if we will do it before const generics, as that will be a breaking change in any case.

theunkn0wn1:
Update to synchronize with current master.
As a note this PR appears to be broken by https://github.com/knurling-rs/defmt/pull/267, which added `heapless` as a dependency.

This causes this PR to fail to resolve in cargo, since the dependency becomes circular.

korken89:
bors try

fkohlgrueber:
It seems like defmt support for heapless should be possible now:
- defmt is released on crates.io
- defmt doesn't depend on heapless anymore (see https://github.com/knurling-rs/defmt/pull/340/files)

From what I see, this PR could be integrated when changing defmt in Cargo.toml from git to version "0.2". 

@theunkn0wn1 do you want to update this PR or should I open another one?

theunkn0wn1:
@fkohlgrueber 
Thanks for the the update!

`defmt` had fallen off my radar so i wasn't aware of the changes.

Looks like `defmt`'s interface changed, changing from git to version 0.2.1 results in build failures, it appears defmt's API changed.

Presently investigating.

theunkn0wn1:
Fixed this PR so it works with defmt 0.2 and I have tested it against my stm32f446 hardware.
Both a Vec and a Str appear to render correctly.

fkohlgrueber:
@theunkn0wn1 seems like there are conflicts that prevent integration. Can you try to resolve them?

theunkn0wn1:
> @theunkn0wn1 seems like there are conflicts that prevent integration. Can you try to resolve them?

The merge conflicts ended up running quite deep and i had to do some minor refactors to match the features, namely const-generics,  presently on master. 

I have verified this PR still appears to work on my f446.
image

Cargo.toml
```diff
@@ -24,6 +24,7 @@ x86-sync-pool = []
 __trybuild = []
 # Enable larger MPMC sizes.
 mpmc_large = []
+defmt-impl = ["defmt"]
```
fkohlgrueber, line 27: You could remove this line and use `defmt` instead of `defmt-impl` as the feature flag. I think that's that standard way of doing it (e.g. see https://github.com/smoltcp-rs/smoltcp/pull/455).

src/lib.rs
```diff
@@ -94,6 +94,8 @@ mod de;
 mod ser;
 
 pub mod binary_heap;
+#[cfg(feature = "defmt-impl")]
```
fkohlgrueber, line 97: you could use `defmt` instead of `defmt-impl` here (see my other comment).

fkohlgrueber (approved):
Looks good to me!

Cargo.toml
theunkn0wn1: I followed the template provided by ufmt-impl on line 20, if we change this PR we should also change the other in a follow up.

MathiasKoch (approved):

korken89 (approved):
LGTM, thanks!

PR #182, opened 2020-11-03
Queue: Expose len and capacity for Consumer and Producer

almusil:
This PR exposes `capacity` and `len` for The `Consumer` and `Producer` which were not available in either of those after split.

### DIFF
--- a/src/spsc/split.rs
+++ b/src/spsc/split.rs
@@ -134,6 +134,23 @@ macro_rules! impl_ {
                 self._dequeue(head) // ▲
             }
 
+            /// Returns the maximum number of elements the queue can hold
+            pub fn capacity(&self) -> $uxx {
+                unsafe { self.rb.as_ref().capacity() }
+            }
+
+            /// Returns the number of elements in the queue
+            ///
+            /// # Note
+            ///
+            /// This is a conservative estimate. Interrupt during this function
+            /// might cause that the `Consumer` actually has more than N items available.
+            pub fn len(&self) -> $uxx {
+                let head = unsafe { self.rb.as_ref().0.head.load_relaxed() };
+                let tail = unsafe { self.rb.as_ref().0.tail.load_acquire() };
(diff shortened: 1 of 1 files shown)


korken89 (changes requested):
Thanks for the PR!

Could you update the documentation to say that these are conservative estimates?
There might be an interrupt in between the two reads that causes the following:

* For the `Producer` will see N free spots, but there might be more free space.
* For the `Consumer` it will see M queued items, but there may be more than M in the queue. 

Else it looks good!

almusil:
> Thanks for the PR!
> 
> Could you update the documentation to say that these are conservative estimates?
> There might be an interrupt in between the two reads that causes the following:
> 
>     * For the `Producer` will see N free spots, but there might be more free space.
> 
>     * For the `Consumer` it will see M queued items, but there may be more than M in the queue.
> 
> 
> Else it looks good!

Done, thank you.

korken89 (approved):
bors merge

almusil:
@korken89 Do you have an idea why it failed on x86 in `std`? I am kinda clueless. Only one thing that comes to my mind is some kind of regression in rustc/llvm?

almusil:
Local run of: `TSAN_OPTIONS="suppressions=$(pwd)/suppressions.txt" RUSTFLAGS="-Z sanitizer=thread" cargo test --test tsan --features x86-sync-pool --target x86_64-unknown-linux-gnu` works just fine

korken89:
bors retry

korken89:
I'm not sure what is wrong here, it's passing for me locally as well.

Ping @japaric, you added these checks do you see what is causing them to fail?

almusil:
Hello,

any updates on this please?
I would love to use heapless with this PR included instead of the fork.

korken89:
bors retry

korken89:
It seems that Travis has shutdown and we need to switch to GHA.
I'll manually merge for now.

PR #186, opened 2020-12-03
First GitHub Actions test

AfoHT:
The test named `testtsan` fails with a 

```
WARNING: ThreadSanitizer: data race (pid=2180)
```

This seems to be a known issue? What would you suggest?

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,4 +1,3 @@
+block_labels = ["S-blocked"]
 delete_merged_branches = true
-status = [
-  "continuous-integration/travis-ci/push",
-]
+status = ["ci"]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,7 +205,7 @@ jobs:
         with:
           use-cross: false
           command: test
-          args: --test tsan --target=${{ matrix.target }} --features=${{ matrix.features }} ${{ matrix.buildtype }}
+          args: --test tsan --target=${{ matrix.target }} --features=${{ matrix.features }} ${{ matrix.buildtype }} -- --test-threads=1
 
   # Run cfail tests on MSRV
   testcfail:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,325 @@
+name: Build
+on:
+  pull_request:
+  push:
+    branches:
+      - master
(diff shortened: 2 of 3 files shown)


korken89:
Thanks for the PR!

Ping @japaric on the test, I've tried to reproduce locally but to no avail so if you can have a look it would be great!
Else I'll remove the test.

japaric:
I can't repro the failure locally with ~nightly-2020-10-24. Is that test blocking CI? I don't see any GitHub Actions run in this repo.

AfoHT:
Not sure why GHA doesn't detect the incoming GHA changes and try them directly, could be a special case with this being the introduction/enabling of GHA.

On my fork the result can be seen: [GHA test heapless](https://github.com/AfoHT/heapless/runs/1494991538?check_suite_focus=true)

The style/cargo fmt failure is fixed already.

korken89:
@japaric Yeah none of us are able to reproduce it, I have even tried different nightlies.
Not sure how to proceed with it. But if we can't reproduce, I'll remove that test for the time being.

japaric:
>  Yeah none of us are able to reproduce it, I have even tried different nightlies.

Yesterday, I had a GHA build failed because I set `CARGO_TERM_COLOR=always`. I doubt that's the case here though.

> On my fork the result can be seen: GHA test heapless

That's hard to read. I would suggest passing `--test-threads=1` to the `cargo test --test tsan --target=x86_64-unknown-linux-gnu --features=x86-sync-pool` command to make it easier (I hope) to spot which test is triggering the sanitizer error.

AfoHT:
How interesting! See [GHA Run](https://github.com/AfoHT/heapless/actions/runs/413625347)

By adding `test-threads=1` the test magically fixed itself, now everything is green :heavy_check_mark:

japaric:
> By adding test-threads=1 the test magically fixed itself, now everything is green 

then it may have been a false positive in the (standard library) test runner. The heapless unit tests don't share share heapless data structures between them (as far as I remember)

korken89:
Then it seems fine, lets merge this!

korken89:
Is there anything you feel is missing from this PR @AfoHT ? Else I'll merge it :)
Great work!

korken89 (approved):

PR #187, opened 2020-12-17
IndexMap, IndexSet: hint to FnvIndex...: Extend explanation, add examples & links

Lotterleben:
in our embedded workshops we tripped over the Example for IndexMap using FNVIndexMap, but `FNVIndexMap` having no example or explan ation for this. Especially when youre in a hurry or new to Rust, it's easy to miss what's going on. This PR is an attempt to clarify this in the docs and copy examples to concrete instantiations of `IndexMap` and `IndexSet`.

I hope I got the wording right, "instantiation" doesn't seem like the exactly right word to describe whats going on...

### DIFF
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -12,7 +12,48 @@ use hash32::{BuildHasher, BuildHasherDefault, FnvHasher, Hash, Hasher};
 
 use crate::Vec;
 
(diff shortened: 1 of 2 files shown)


korken89 (approved):
Thanks for improving!

PR #192, opened 2021-03-02
Update generic-array and as-slice to security patched versions

AfoHT:
Resolves RUSTSEC-2020-0145

https://github.com/RustSec/advisory-db/pull/779


korken89 (approved):
bors r+

PR #198, opened 2021-03-26
Const generics port

### LINKED ISSUE #115: Suggestion: consider exposing the Uxx and XCore traits for SPSC queues
It is currently not possible to be generic over the index type or single core/multi core SPSC queue variants because the traits `Uxx` and `XCore` can't be declared as type parameters because they are private.

If the traits were sealed so that the crate retained control over what types could implement `Uxx` and `XCore` it would be possible to write more generic code utilizing SPSC queues.

### LINKED ISSUE #158: Cannot create static HistoryBuffer
With `Vec` it is possible to do this:

```Rust
use heapless::Vec; // fixed capacity `std::Vec`
use heapless::consts::U8; // type level integer used to specify capacity
static mut XS: Vec<u8, U8> = Vec(heapless::i::Vec::new());
let xs = unsafe { &mut XS };
```

However [HistoryBuffer](https://github.com/japaric/heapless/blob/master/src/histbuf.rs#L37) is missing in `heapless::i`.

### LINKED ISSUE #168: Const generics
Const generics are getting more popular. I can try to implement `heapless` using const generics and push it into a separate branch until they are stable. It is a nice challenge both for coding skills and for const generics.
Shall I start?

### LINKED ISSUE #207: spsc::Queue corrupts its head/tail pointers upon wrapping, if capacity is non-power-of-two.
Hi,

spsc::Queue supports arbitrary capacities (i.e., non-power-of-two), but does not wrap its `head` and `tail` pointers correctly when en/dequeueing. Instead, only `wrapping_add` / `wrapping_sub` is used for pointer/index arithmetics, while an actual modulo operation is only used for accessing the array elements. Effectively, these pointers are incremented modulo 2^8 or whatever bitness the underlying integer has.

This is only sound if the capacity is a divisor of 2^8, i.e. only for power-of-two capacities. (`((a+1) % b) % c == (a+1) % c` iif `b` is a multiple of `c`, which is not the case here.)

The following test case illustrates this bug. (It can be inserted to the test suite in src/spsc/mod.rs and executed using `cargo test --features 'serde','x86-sync-pool' spsc`):

```rust
    #[test]
    fn overflow_tail() {
        let mut rb: Queue<i32, U5, u8> = Queue::u8();

        for i in 0..4 {
            rb.enqueue(i).unwrap();
        }
        for _ in 0..70 {
            for i in 0..4 {
                assert_eq!(rb.dequeue().unwrap(), i);
                rb.enqueue(i).unwrap();
            }
        }
    }
```

Possible solutions include:
1. Doing the modulo step on every write operation on head/tail; this will likely not incur additional costs, because now the modulo operation for accessing the array element becomes unneccessary. However, it will make a "queue full" condition indistinguable from "queue empty" (both yield `head == tail`).
2. Like 1, but with double the capacity as the modulus. This incurs an additional cost, because the second modulo will still be necessary. Additionally, this limits the allowed max capacity to half of what the underlying integer can hold.
3. Like 1, but wasting one element of queue space. In this case, "queue empty" is represented by `head == tail`, while "queue full" is `head == tail + 1 (mod capacity)`.
4. Simply disallow non-power-of-two queue sizes.

korken89:
Closes #158
Closes #168
Closes #207
Closes #115


kpp:
Hey, there are several fixes for this PR: https://github.com/japaric/heapless/pull/199. All tests with const generics passed successfully.

cfail/ui/freeze.stderr
```diff
@@ -1,9 +1,13 @@
...
-6 |     let (_p, mut _c) = q.split();
-  |                        - first mutable borrow occurs here
-7 |     q.enqueue(0).unwrap();
-  |     ^ second mutable borrow occurs here
-8 |     _c.dequeue();
-  |     -- first borrow later used here
+error[E0107]: this struct takes 3 generic arguments but 4 generic arguments were supplied
+   --> $DIR/freeze.rs:4:16
+    |
+4   |     let mut q: Queue<u8, _, _, 4> = Queue::new();
```
kpp, line 4: Looks like wrong stderr was captured.

cfail/ui/not-send.stderr
```diff
@@ -1,83 +1,60 @@
-error[E0277]: `*const ()` cannot be sent between threads safely
-  --> $DIR/not-send.rs:19:5
+error[E0107]: this struct takes 3 generic arguments but 4 generic arguments were supplied
```
kpp, line 1: This stderr is wrong too

src/de.rs
```diff
@@ -1,36 +1,31 @@
...
-use crate::{
-    indexmap::{Bucket, Pos},
-    sealed::binary_heap::Kind as BinaryHeapKind,
-    BinaryHeap, IndexMap, IndexSet, LinearMap, String, Vec,
-};
-
 // Sequential containers
 
-impl<'de, T, N, KIND> Deserialize<'de> for BinaryHeap<T, N, KIND>
+impl<'de, T, KIND, const N: usize> Deserialize<'de> for BinaryHeap<T, KIND, N>
```
kpp, line 11: Is it okay to put `N` after `KIND`? This PR is a breaking change however it will take more effort for users to upgrade to a new version with const generics.

burrbull: const generics can be passed only after lifetimes and after regular generics by design

cfail/ui/freeze.stderr
korken89: Thanks for noticing!

src/de.rs
korken89: Indeed, we wanted to not do it but const generics forces it. Same with the default type parameter not working when using const generics :/

burrbull: https://github.com/rust-lang/rust/pull/75384

korken89: Ah I was referring to default type parameters note working before the const generics parameter.

cfail/ui/freeze.stderr
kpp: I will fix it

src/histbuf.rs
```diff
@@ -1,72 +1,65 @@
...
-/// assert_eq!(buf.recent(), &4);
+/// assert_eq!(buf.recent(), Some(&4));
 ///
 /// // To access all elements in an unspecified order, use `as_slice()`.
 /// for el in buf.as_slice() { println!("{:?}", el); }
 ///
-/// // Now we can prepare an average of all values, which comes out to 2.
+/// // Now we can prepare an average of all values, which comes out to 4.
 /// let avg = buf.as_slice().iter().sum::<usize>() / buf.len();
-/// assert_eq!(avg, 2);
```
kpp, line 34: Wow, this change should be explicitly noticed or even ported to a non const generic version.

src/histbuf.rs
```diff
@@ -1,72 +1,65 @@
...
 
-impl<T, N> HistoryBuffer<T, N>
-where
-    N: ArrayLength<T>,
-    T: Default,
-{
-    /// Constructs a new history buffer, where every element is filled with the
-    /// default value of the type `T`.
+impl<T, const N: usize> HistoryBuffer<T, N> {
+    const INIT: MaybeUninit<T> = MaybeUninit::uninit();
```
kpp, line 42: Is there any reason for `const INIT`?

src/histbuf.rs
```diff
@@ -146,32 +153,36 @@ where
...
+            Some(unsafe { &*self.data[self.write_at - 1].as_ptr() })
         }
     }
 
     /// Returns the array slice backing the buffer, without keeping track
     /// of the write position. Therefore, the element order is unspecified.
     pub fn as_slice(&self) -> &[T] {
-        &self.data
+        if self.filled {
+            unsafe { slice::from_raw_parts(self.data.as_ptr() as *const _, self.capacity()) }
```
kpp, line 178: Oh no! UB detected (out of bounds).

It should be `unsafe { slice::from_raw_parts(self.data.as_ptr() as *const _, self.len()) }`. And no `if self.filled` is required.

kpp:
```suggestion
            unsafe { slice::from_raw_parts(self.data.as_ptr() as *const _, self.len()) }
```

src/histbuf.rs
```diff
@@ -195,26 +205,37 @@ where
...
 #[cfg(test)]
 mod tests {
-    use crate::{consts::*, HistoryBuffer};
+    use crate::HistoryBuffer;
 
     #[test]
     fn new() {
-        let x: HistoryBuffer<u8, U4> = HistoryBuffer::new_with(1);
+        let x: HistoryBuffer<u8, 4> = HistoryBuffer::new_with(1);
         assert_eq!(x.len(), 4);
```
kpp, line 226: I thought `.len()` should be 1, `.capacity()` should be 4

src/spsc/mod.rs
```diff
@@ -83,42 +83,32 @@
...
 mod split;
 
-/// Multi core synchronization - a memory barrier is used for synchronization
-pub struct MultiCore;
-
-/// Single core synchronization - no memory barrier synchronization, just a compiler fence
-pub struct SingleCore;
-
-// Atomic{U8,U16, Usize} with no CAS operations that works on targets that have "no atomic support"
+// Atomic{U8, U16, Usize} with no CAS operations that works on targets that have "no atomic support"
```
kpp, line 95: Unfortunately U8 and other consts should be fixed in docs.

src/spsc/mod.rs
```diff
@@ -702,7 +600,7 @@ mod tests {
 
     #[test]
     fn iter() {
-        let mut rb: Queue<i32, U4> = Queue::new();
+        let mut rb: Queue<i32, u16, 4> = Queue::u16();
```
kpp, line 603: This is wrong. `::new` is an alias for `::usize()`.

kpp: Every change from `Queue::new();` to `Queue::$uxx();` should be checked, because not only we are changing the code but tests too... Whereas we should not change tests as much as possible.

src/spsc/split.rs
```diff
@@ -268,11 +244,11 @@ impl_!(usize);
 
 #[cfg(test)]
 mod tests {
-    use crate::{consts::*, spsc::Queue};
+    use crate::spsc::Queue;
 
     #[test]
     fn sanity() {
-        let mut rb: Queue<i32, U2> = Queue::new();
+        let mut rb: Queue<i32, u8, 2> = Queue::u8();
```
kpp, line 251: Again, ::new -> ::u8 is questionable.

cfail/ui/freeze.stderr
kpp: Fixed. This one can be resolved.

cfail/ui/not-send.stderr
kpp: Fixed. This one can be resolved.

src/histbuf.rs
kpp: Fixed in https://github.com/japaric/heapless/pull/203

src/histbuf.rs
korken89: We will have it in the CHANGELOG as a breaking change

src/histbuf.rs
korken89: This is to get around the issues of the following not working:

```rust
fn new() -> Self {
    Self {
        buffer: [MaybeUninit::uninit(); N]
    }
}
```

src/histbuf.rs
korken89: `new_with` fills the entire buffer with a specific value, so `len` will be 4.

src/spsc/mod.rs
korken89: They should be fixed, this is part of the `Atomic`'s type signature which has `Uxx` in its name.

src/spsc/mod.rs
korken89: Good catch!

src/histbuf.rs
```diff
@@ -108,22 +101,36 @@ where
...
+    pub fn capacity(&self) -> usize {
+        N
     }
 
     /// Writes an element to the buffer, overwriting the oldest value.
     pub fn write(&mut self, t: T) {
-        self.data[self.write_at] = t;
+        if self.filled {
+            // Drop the old before we overwrite it.
+            unsafe { ptr::drop_in_place(self.data[self.write_at].as_mut_ptr()) }
```
AfoHT, line 126: For all uses of unsafe, what is the heapless policy regarding docs justifying correctness?

Just curious

src/sealed.rs
```diff
@@ -63,17 +38,11 @@ pub mod spsc {
...
-            if C::is_multi_core() {
-                (*(x as *const AtomicU8)).load(Ordering::Acquire)
-            } else {
-                let y = (*(x as *const AtomicU8)).load(Ordering::Relaxed); // read
-                atomic::compiler_fence(Ordering::Acquire); // ▼
-                y
-            }
+        unsafe fn load_acquire(x: *const Self) -> Self {
+            (*(x as *const AtomicU8)).load(Ordering::Acquire)
+            // let y = (*(x as *const AtomicU8)).load(Ordering::Relaxed); // read
```
AfoHT, line 43: If multi-core is being removed, does it make sense keeping this, or will multi-core reappear later on?

AfoHT (commented):
Much cleaner with consts support :+1:

src/histbuf.rs
korken89: It currently seems like there is none, only functions with unsafe interfaces are currently documented.
Adding a motivation to all accesses would be good though, but I think that will have to be its own PR.

src/sealed.rs
korken89: I was thinking if I should keep it or remove it, I have not favor towards either.

Sympatron:
Why are `spsc::Queue::u8()` and `spsc::Queue::u16()` unsafe now?

korken89:
@Sympatron The docs state (for `Queue::u16()`): "Creates an empty queue with a fixed capacity of `N`. **Safety**: Assumes `N <= u16::MAX`.

More or less, we cannot check in const context if N is within the limit of the index.

burrbull:
We can generate safe constructors for limited range of `N` with macro.

korken89:
@burrbull How? Const asserts are not allowed to be used with const generics for now.

burrbull:
> @burrbull How? Const asserts are not allowed to be used with const generics for now.

Draft
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=01ab8d3a76ad017bd2c365d981ff2bc3

burrbull:
See #205

src/indexmap.rs
```diff
@@ -373,22 +360,19 @@ where
...
+impl<K, V, S, const N: usize> IndexMap<K, V, BuildHasherDefault<S>, N>
 where
     K: Eq + Hash,
     S: Default + Hasher,
-    N: ArrayLength<Bucket<K, V>> + ArrayLength<Option<Pos>> + PowerOfTwo,
 {
-    // TODO turn into a `const fn`; needs `mem::zeroed` to be a `const fn`
     /// Creates an empty `IndexMap`.
     ///
     /// **NOTE** This constructor will become a `const fn` in the future
```
MabezDev, line 378: Could this be `const` now? I notice there is a `const-fn` feature of `hash32` to allow the construction of the hasher in `const` time.

burrbull:
Is this mean you've decided to use atomic `usize` everywhere?

src/indexmap.rs
korken89: Hi, this requires an update to `hash32`, the feature is not enough.
I'll look into it.

korken89:
@burrbull Yes, the implementation that existed was not easily amendable to the fix of #207, and the known good implementation performs very well.
The key here is simplicity, easy to check for correctness - where the old implementation was unfortunately not.

So for now I'll remove the support for smaller queue indexes, and we can reintroduce them in the future if necessary.

src/histbuf.rs
```diff
@@ -1,72 +1,65 @@
...
-/// initial value is either the default value for the element type or a supplied
-/// initial value. This simplifies the API and is mostly irrelevant for the
-/// intended use case.
-///
 /// # Examples
 /// ```
 /// use heapless::HistoryBuffer;
-/// use heapless::consts::*;
 ///
 /// // Initialize a new buffer with 8 elements, all initially zero.
```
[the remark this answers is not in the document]
korken89: Thanks for finding!

src/indexmap.rs
korken89: I have made a PR to solve this issue.

https://github.com/japaric/hash32/pull/9

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 ## [Unreleased]
 
+### Changed
+
+- [breaking-change] Converted all data structures to use the `const generics` MVP
+- [breaking-change] `HistoryBuffer` is now working with const constructors and non-`Copy` data
+- [breaking-change] `HistoryBuffer::as_slice` and others now only return initialized values
+- [breaking-change] `MultiCore`/`SingleCore` is now removed from `spsc::Queue`
+- [breaking-change] `spsc::Queue` is now `usize` only
+- [breaking-change] `spsc::Queue` now sacrifices one element for correctness (see issue #207), i.e. it creates an `N - 1` sized queue instead of the old that generated an size `N` queue
```
Sympatron, line 17: I think that is counter intuitive. It should create an `N` sized queue and reserve one element more for correctness.

Also depending on the type of element this could waste a lot of space. Maybe some kind of flag would be better for correctness than sacrificing an element.

korken89: Hi, thanks for the input!

> This implementation detail should not bleed through in my opinion.

I agree that this would be ideal, but unfortunately not possible today.
This is not possible to fix with the current const generics, as math on const generics is not allowed, i.e. `buffer: [MaybeUninit<T>; { N + 1 }]` is not supported.

> Maybe some kind of flag would be better for correctness than sacrificing an element.

This method exists, but is non-trivial to implement. It becomes problematic to have the `Producer` / `Consumer` pairs as they are now memory dependent via this `full` flag.
The `full` flag which is either a `bool` or a counter that checks the fill level of the queue both implementation get a requirement on having CAS operations or `interrupt::free` for accessing and updating this flag.

The new implementation is selected for its simplicity in showing correctness, and correctness is the no. 1 priority.
I however welcome a stab at trying to solve it, I did do some experiments and I was unable getting TSAN happy while keeping simplicity.

> Also depending on the type of element this could waste a lot of space. 

This is, to me, a moot point. If the size of the object is large one should ideally already be using `pool` and send `pool::Box`es.

Sympatron: > i.e. `buffer: [MaybeUninit<T>; { N + 1 }]` is not supported.

I though so. That's why I suggested flags.
> This method exists, but is non-trivial to implement.

I feared this would be the case. I just hoped that somebody smarter than me could figure out some way to do it ;)
Now that I think about it, I only ever implemented lock-free queues with `N-1` usable elements and locking queues with a full flag. Never both at the same time, but I thought I was just too lazy to think about it hard enough to figure it out :upside_down_face:.

>correctness is the no. 1 priority.

I totally agree. If it's not guaranteed to be correct it would basically be useless.

> ...one should ideally already be using `pool` and send `pool::Boxes`.

You are probably right about that.

korken89: Yeah, you know, it's unfortunate.
It will take a bit of getting used to, but this is also what is common for SPSC queues in embedded where almost everyone (including us) base their design on this article: https://www.codeproject.com/Articles/43510/Lock-Free-Single-Producer-Single-Consumer-Circular
A great resource with a lot of details on correctness and pros/cons.

Thanks for you feedback, I really appreciate it!
I hope we could solve it easily but as it is now I think we need to solve it in a future release when const generics have gained more power.

PR #200, opened 2021-04-02
remove as_slice dependency

burrbull:


korken89:
Did the core trait cover the as slice use? I remember it did not do it before?

burrbull:
`as_slice` was introduced because array impls (including `AsRef<[T]>`) were limited by 32. Similar situation was with `GenericArray`.
As now all arrays are generic it's time to forget about `as_slice`.

korken89:
Ah perfect, thanks for fixing!

Issue #139, opened 2020-01-09
Allow changing the `len` type of `Vec`


jonas-schievink:
`Queue` supports switching out the type to use for head and tail pointers, which can save memory (instead of 8 Bytes for 2 `usize`s, I can choose to pay only 2 Bytes for 2 `u8`s). It would be nice to have the same option for `Vec`, which currently always uses `usize`.

YuhanLiin:
I looked into implementing this using a generic length type and I'm running into issues initializing the generic length to 0. The constructor for `Vec` is a `const fn` so I can't initialize the length with any trait methods. I might look into using macros like `Queue` does to implement most methods.

korken89:
With the `const generics` MVP and found soundness issues, I have decided to keep `heapless` simple and easier to show correctness.
Hence I will not be pursuing these micro optimizations for the time being.

PR #204, opened 2021-04-07
Allow u8 and u16 to be used as the length of Vec-based containers

YuhanLiin:
I added a new type parameter for length to `Vec` and containers that use `Vec`. This allows `u8` and `u16` to be used for the length of `Vec` to save space, similar to the head and tail pointers of `Queue`. Additionally, the capacity of `Vec` is statically checked to be within the range of the length type using `typenum` trait bounds (for example, a `Vec` with length type of `u8` and capacity 256 is not allowed). Addresses #139.

### DIFF
--- a/cfail/ui/u16-vec.rs
+++ b/cfail/ui/u16-vec.rs
@@ -1,0 +1,9 @@
+//! Vec with u16 length and capacity of 65536
+
+use core::marker::PhantomData;
+
+use heapless::{consts::*, Vec};
+
+fn main() {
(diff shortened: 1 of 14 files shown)


korken89:
Hi! We are about to merge the `const generics` port, could you rebase this of that work please?
The merge conflicts this will have will be huge unfortunately.

YuhanLiin:
Might be easier to rewrite the whole thing

korken89:
Unfortunately I think this might be true.

PR #205, opened 2021-04-08
safe new for pow2 sized spsc queues

burrbull:


korken89:
Hi, unfortunately this misses most of the possible size values that the size can have.
It's not only limited to powers of 2.

PR #206, opened 2021-04-08
Allow u8 and u16 to be used as the length of Vec-based containers (const generics edition)

YuhanLiin:
Basically the same as #204 except this one's on top of the const_generics branch. The compile-time capacity checking is gone because setting trait bounds with const generics is not supported (this is arguably a good thing because it makes the trait bounds much simpler). Also there's no default length type since default types and const generics don't mix. Still addresses #139.

### DIFF
--- a/cfail/ui/not-send.rs
+++ b/cfail/ui/not-send.rs
@@ -16,5 +16,5 @@ fn main() {
     is_send::<Consumer<NotSend, _, 4>>();
     is_send::<Producer<NotSend, _, 4>>();
     is_send::<Queue<NotSend, _, 4>>();
-    is_send::<heapless::Vec<NotSend, usize, 4>>();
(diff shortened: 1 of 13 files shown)


burrbull:
@korken89 
As it is very big breaking change, if we are going to allow this for merge, we need to do this before release, or forget.

korken89:
Indeed. I'll try to get the const generics branch merged so this can be rebased. Then I'll review this.

burrbull:
Maybe it is make sense to rename base structs and use type aliases for `usize` variants? Like:
```
pub type Vec<T, const N: usize> = VecBase<T, usize, N>;
```

korken89:
Hi, this seems to have been closed due to the merge and delete of the const generics branch.
Can you rebase this on master?

PR #208, opened 2021-04-22
Allow u8 and u16 to be used as the length of Vec-based containers (3rd time's the charm)

YuhanLiin:
Same PR as #206, targeting the master branch. Still addresses #139.


korken89:
Hi @YuhanLiin, thank you for the PR!

Great work, but unfortunately I have to say that I cannot merge this.
I cannot motivate the code complexity vs the very small memory gain for a `Vec`, or containers based on it.
The use-case where one is dependent on this optimization is only if you have an enormous amount of `Vec`s, and for this use-case I'd recommend to make a special implementation for this specific use-case.

Thank you again for your hard work!

YuhanLiin:
Would it be possible to close #139 as well since that's the issue this PR is addressing? It would prevent future maintainers from attempting a similar patch.

korken89:
Absolutely, thanks for pointing that out!

PR #209, opened 2021-05-07
Generic MPMC

vinaychandra:
Current MPMC has a very limited size set. This PR makes it generic over different sizes.

The PR also preserves the current behavior with U8 as max size. This should be completely backward compatible. 
There is also an alternative with Usized size which doesn't have the limitation on the number of elements. This can be enabled by the `mpmc_large` feature flag.


korken89:
Hi, thanks for the PR and the improvement of the MPMC Queue!

It looks good to me.

korken89 (approved):

korken89 (approved):

PR #211, opened 2021-06-25
Add new `Vec::into_array` method

GrantM11235:
With the addition of this method, `heapless::Vec` can be used to safely and easily create an array one element at a time.

I don't have much experience with `MaybeUninit`, so please double check my implementation.


korken89 (approved):
Looks good, thanks!

korken89:
bors merge

PR #219, opened 2021-07-14
Add Deque.

Dirbaio:
This PR adds a heapless `Deque` struct. which is the heapless equivalent of the good old [`VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html).

This is different from the MPMC/SPSC `Queue`s. This does not aim to be lock-free. All mutating operations require `&mut self`. This allows for much more functionality, such as pushing/popping from both front and back, iterating the contents (even mutably), etc.

### DIFF
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -1,0 +1,775 @@
+use core::iter::FusedIterator;
+use core::marker::PhantomData;
+use core::mem::MaybeUninit;
+use core::{ptr, slice};
+
+/// A fixed capacity double-ended queue.
+///
+/// # Examples
+///
+/// ```
+/// use heapless::Deque;
+///
+/// // A deque with a fixed capacity of 8 elements allocated on the stack
+/// let mut deque = Deque::<_, 8>::new();
+///
+/// // You can use it as a good old FIFO queue.
+/// deque.push_back(1);
+/// deque.push_back(2);
+/// assert_eq!(deque.len(), 2);
+///
+/// assert_eq!(deque.pop_front(), Some(1));
+/// assert_eq!(deque.pop_front(), Some(2));
+/// assert_eq!(deque.len(), 0);
+///
+/// // Deque is double-ended, you can push and pop from the front and back.
+/// deque.push_back(1);
+/// deque.push_front(2);
+/// deque.push_back(3);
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -82,6 +83,7 @@ pub use string::String;
 pub use vec::Vec;
 
 // NOTE this code was last ported from v0.4.1 of the indexmap crate
+mod deque;
```
thalesfragoso, line 86: Maybe this comment should be moved ?

src/deque.rs
```diff
@@ -0,0 +1,775 @@
...
+
+    #[test]
+    fn static_new() {
+        static mut _V: Deque<i32, 4> = Deque::new();
+    }
+
+    #[test]
+    fn stack_new() {
+        static mut _V: Deque<i32, 4> = Deque::new();
+    }
```
thalesfragoso, line 522: copy-paste error ?

thalesfragoso (commented):
This looks great, thank you!

src/deque.rs
Dirbaio: It's also wrong in `Vec` :rofl:. Fixed in both.

src/lib.rs
Dirbaio: :shrug: Seems like it was already wrong before, as I don't think it applied to `histbuf`. It seems very old anyway, I've removed it.

korken89:
Hi,

I hope I can review this soon. In the meanwhile, could you have a look at the failing tests?

Dirbaio:
@korken89 thanks! I believe the tests should now pass, they pass locally with 1.51

korken89:
I gave it a look and mostly looks good!

One request, could you add `peek_front` and `peek_back` methods?
Analogous to the `peek` method in `Queue`. Thanks!

Dirbaio:
@korken89: added `front, front_mut, back, back_mut`, matching std's VecDeque.

korken89:
@Dirbaio  Thanks! This looks good to me!
Anything you want to add before merge?

korken89 (approved):

Dirbaio:
Nothing in particular, it's complete for what I need it :)

PR #221, opened 2021-07-20
spsc: fix optimal queue size docstring

jordens:
`capacity` is defined as `N - 1`. But best performance (cheap modulo) is achieved if `N` is a power of two, not `capacity`.
This fixes the docstring.


korken89 (approved):
bors merge

PR #223, opened 2021-07-23
Adding assume_init function for Box

### LINKED ISSUE #222: Box::init performs a deep copy when using uninitialized memory
In cases where a Box contains a large value (e.g. `[u8; 1024]`), the copy can be expensive to initialize the `Box` to a defined value. 

In trivial cases like this, where only a buffer is stored, there's no need to necessary overwrite the array if the data will be written later.

The proposed solution in #223 is to utilize `MaybeUnint<[u8; 1024]>` and initialize to the `MaybeUninit::uninit()` variant. However, as shown in [godbolt](https://godbolt.org/z/n8r6EexK6), initialization in this way still results in a copy of 1024 bytes.

It would be desirable to initialize the Box to `MaybeUninit<[u8; 1024]` without copying the 1024 byte array.

ryan-summers:
This PR fixes #222 by adding an `assume_init()` for pool::Box structures. This allows the underlying box'd data to not require memory copies for initialization (e.g. for the case when an array is Box'd and the contents are not yet written).

### DIFF
--- a/src/pool/mod.rs
+++ b/src/pool/mod.rs
@@ -425,6 +425,17 @@ impl<T> Box<T, Uninit> {
             _state: PhantomData,
         }
     }
+
+    /// Assumes the memory block is already initialized.
+    ///
+    /// # Safety
+    /// The underlying memory must be already initialized to an acceptable, defined state.
+    pub unsafe fn assume_init(self) -> Box<T, Init> {
+        Box {
+            node: self.node,
+            _state: PhantomData,
+        }
+    }
 }
 
 /// Uninitialized type state


korken89:
Hi, I'm not sure this is needed. If you want uninitialised data you can use `[MaybeUninit<T>; N]` and init with `MaybeUninit::new()`.
This should give the same effect you are after without copies.

I think using this `assume_init()` to have uninitialised byte buffers is UB, as you are then allowed to have references to initialized data (which is UB, even if you never read uninitialised data).
Maybe there is a use case I'm missing here?

ryan-summers:
It's my understanding that because of the semantics of `heapless::pool::Box`, it's ultimately irrelevant whether the internal type is `MaybeUninit<>` or not because a `ptr.write()` is done on the type. When the type is `MaybeUninit<>`, a union type, the whole `MaybeUninit<>` type is copied. If the MaybeUninit is over a buffer, this means the whole buffer is also copied (e.g. `MaybeUninit<[u8; 1024]>` still results in a copy of 1024 bytes.

Check out the assembly output when analyzing `ptr.write()`, which is what I believe heapless uses internally:
https://godbolt.org/z/n8r6EexK6

ryan-summers:
> I think using this `assume_init()` to have uninitialised byte buffers is UB, as you are then allowed to have references to initialized data (which is UB, even if you never read uninitialised data).
> Maybe there is a use case I'm missing here?

Yes, this definitely can result in UB if care isn't taken on the underlying type T - for simple byte buffers, any memory value is a valid value for the memory, so it could be argued that raw u8 arrays can never be invalid, but this would definitely be an issue if you had a `Box<bool>`, as any value other than 0 or 1 is UB. That was the intent of the `unsafe` specifier.

korken89:
Hi,

Unfortunately that argument does not hold: https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html
With this specifically:

> An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained from uninitialized memory, or uninitialized memory in a `str`.

This is the same issue as we ran into designing `Vec` and `Queue`, you must first write a valid value.
E.g. fill with 0, and then do the `assume_init()`.

---

On the issue of the copy, I am not sure why this is happening.
I think this could be considered a bug. I'll have to investigate that.

ryan-summers:
> Hi,
> 
> Unfortunately that argument does not hold: https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html
> With this specifically:
> 
> > An integer (`i*`/`u*`), floating point value (`f*`), or raw pointer obtained from uninitialized memory, or uninitialized memory in a `str`.
> 
> This is the same issue as we ran into designing `Vec` and `Queue`, you must first write a valid value.
> E.g. fill with 0, and then do the `assume_init()`.

You make a good point in regard to the UB. Even with an unsafe function, we shouldn't expose the functionality of UB. Regardless of my desire to know how a i*/u* type could be invalid, it's very clear for other types like `bool`, etc. so the proposed solution in this PR is definitely not the desired solution. Thanks for the links!

> On the issue of the copy, I am not sure why this is happening.
> I think this could be considered a bug. I'll have to investigate that.

I have a suspicion it's because of the `ptr::write` instead of assignment. When doing an assignment to a `MaybeUninit` type, if it's `uninit()`, I believe it would only need to copy a few bytes (to indicate the `Uninit` enum variant), but I'm not sure.

ryan-summers:
I've updated #222 to reflect the issue of deep-copying even on MaybeUninit data, and I'll close this PR as unsound. Thanks for the feedback!

PR #227, opened 2021-08-08
Added sorted linked list implementation

korken89:
A linked list priority queue, similar to binary heap.
Has different useful properties.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,6 +101,7 @@ mod defmt;
 pub mod mpmc;
 #[cfg(all(has_cas, feature = "cas"))]
 pub mod pool;
+pub mod sorted_linked_list;
 #[cfg(has_atomics)]
 pub mod spsc;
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
Name should maybe be `SortedLinkedList`? I would expect `LinkedList` to be the good old linked list where the user is in control of the order.

korken89:
@Dirbaio I see your point, but this not done for the binary heap, so I think the name as it is is proper. Maybe you had a different issue with the name in mind?
I can also expand on the module documentation.

One can also get a non sorted list by implementing Ord to equal the Ordering of Min/Max, though a bit cumbersome. :)

Dirbaio:
binary heaps are "sorted" by definition (as in a node is greater/smaller than its children), so the name "sorted binary heap" would be redundant. Linked lists aren't: you put things in and they stay in the order you put them in. This is a linked list that additionally sorts itself, hence SortedLinkedList :)

korken89:
@Dirbaio Good point! I changed name.

perlindgren (approved):
Overall structure and safety assessments look good. Minor details have been confirmed with @korken89 during the review.

korken89:
Thanks for the review!

PR #229, opened 2021-08-16
Made pop_unchecked public, added is_empty

korken89:


perlindgren (approved):
Looks good, and makes the API symmetric.

PR #233, opened 2021-08-25
Sync Pool support for (32-bit) x86

### LINKED ISSUE #231: Sync Pool support for i686
The `x86-sync-pool` Cargo feature lets you use the `pool!` macro on x86_64 with the caveat that it has a small probability of running into Undefined Behavior (UB). The main use case of `x86-sync-pool` are *short-lived* tests running on a time-sliced scheduled (e.g. Linux). Using `x86-sync-pool` on a long lived process or a process that uses prioritized threads has a high(er) chance of running into UB.

Because Rust does not support / expose 128-bit atomics (the x86_64 architecture supports them natively) the x86_64 implementation uses a 64-bit atomic to store *both* a 32-bit pointer and a 32-bit 'version tag'. The downside of using a 64-bit atomic is that the memory addresses that the `Pool` can manage are limited to +- 2 GB around some `ANCHOR` address. This `ANCHOR` is placed in `.bss` so the `Pool` may not be able to manage heap allocated (e.g. `std::Box`) memory: any memory outside the 4GB space around `ANCHOR` will be silently "dropped" (leaked).

A practical solution to this problem is to run a 32-bit version of the program on the x86_64 processor, e.g. using `cargo run --target i686-linux-unknown-musl`. When using 32-bit mode on x86_64 the OS will memory map available (RAM) memory to a 32-bit memory space; this removes the need for an `ANCHOR` address in the implementation. The i686 supports `AtomicU64` so that can be used to store a 32-bit pointer plus the 32-bit version tag in a i686 `Pool` implementation. Using a 32-bit version of `Pool` on a x86_64 processor does not change (increase or reduce) the probability of running into the ABA problem (UB).

japaric:
the implementation uses a 64-bit atomic on `x86` to avoid the `ANCHOR` variable and the address
space limitation seen with the x86_64 compilation target

this PR also adds the i686-linux-musl target to the test matrix to exercise the new implementation

closes #231

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,8 +41,8 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - i686-unknown-linux-musl
           - riscv32imc-unknown-none-elf
-          - x86_64-unknown-linux-gnu
           - armv7r-none-eabi
           - thumbv6m-none-eabi
           - thumbv7m-none-eabi
@@ -102,6 +102,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - i686-unknown-linux-musl
         toolchain:
           - stable
           - nightly
@@ -157,6 +158,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - i686-unknown-linux-musl
         toolchain:
           - nightly
         features:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,6 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -41,8 +41,8 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - i686-unknown-linux-musl
           - riscv32imc-unknown-none-elf
-          - x86_64-unknown-linux-gnu
```
japaric, line 45: this target was listed twice in the matrix so I removed it

japaric:
(sorry for the noise; PEBCAK issues with git; I think it should be green-er now?)

korken89 (approved):

korken89:
Looks good

PR #238, opened 2021-09-10
add Arc pointer backed by a memory pool

japaric:
Like `std::sync::Arc` but backed by a memory pool (just like `heapless::Box`). The memory pool (singleton) is defined using the `arc_pool!` macro.

The [`Weak`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.downgrade) API is not included with this PR, but it could be added in the future if desired.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,6 +43,7 @@
 //!
 //! List of currently implemented data structures:
 //!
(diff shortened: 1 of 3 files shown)


korken89:
Thanks for the addition!
Overall it looks good, I have only one request: Could you add the `grow_exact` method to have the same as with `pool!`?

japaric:
@korken89 added!

korken89:
Looking good!

korken89 (approved):

PR #240, opened 2021-09-12
Miri fixes

Dirbaio:
- Fix 2 miri errors in `Deque`
- Ignore a few tests that take way too long in miri (gave up after waiting 5+ min)


korken89 (approved):
Thanks for fixing!

PR #241, opened 2021-09-13
Allow arc pool on thumbv6

korken89:


korken89:
Hey @japaric was there a reason for not allowing the arc pool's macro on thumbv6?
I added it here, but I want to check with you before merging.

japaric:
> was there a reason for not allowing the arc pool's macro on thumbv6?

I used the same cfg that the "Box" `pool` has and v6M is not there either. The pool implementation needs CAS operations which v6M does not natively have. I think there's some interrupt-free based CAS fallback but I have never used it myself and I don't have v6M hardware on hand to test. If you are positive this works, could please also add v6m to the `pool!` macro?

On a side note: as a refactor it would be nice to use something like `cfg(all(has_cas, has_llsc))` instead of the long list of ARM targets. That cfg would need to be OR-ed with the x86 / cmpxchg targets

korken89:
Ah yes, we should change to use `has_cas, has_llsc` instead. The polyfill fixes it for thumbv6 with critical sections.

japaric (approved):
this looks good to me. feel free to land as it is or with the `pool!` addition or the `has_llsc` refactor (or feel free to postpone those for follow-up PRs)

japaric:
bors delegate+

PR #244, opened 2021-09-22
Fix MPMC size asserts

### LINKED ISSUE #243: mpmc only works with sizes that are power of 2
This goes into an endless loop on dequeue:
```
let q = mpmc::MpMcQueue::<i32, 3>::new();
q.enqueue(0).unwrap();
q.enqueue(1).unwrap();
q.enqueue(0).unwrap();
q.dequeue().unwrap();
```

This panics:
```
let q = mpmc::MpMcQueue::<i32, 6>::new();
q.enqueue(0).unwrap();
q.enqueue(1).unwrap();
q.enqueue(0).unwrap();
q.enqueue(1).unwrap();
q.enqueue(0).unwrap();
q.enqueue(1).unwrap();
q.dequeue().unwrap();
```

[Original implementation](https://www.1024cores.net/home/lock-free-algorithms/queues/bounded-mpmc-queue) seems to have an assert for power of 2. So if that's an implementation requirement it would be nice to mention in the docs and add an assert.

I can submit a PR someday, if nobody fixes this until then

chemicstry:
Fixes #243

N must also be at least 2

### DIFF
--- a/src/mpmc.rs
+++ b/src/mpmc.rs
@@ -122,6 +122,7 @@ pub type Q32<T> = MpMcQueue<T, 32>;
 pub type Q64<T> = MpMcQueue<T, 64>;
 
 /// MPMC queue with a capacity for N elements
+/// N must be a power of 2
 /// The max value of N is u8::MAX - 1 if `mpmc_large` feature is not enabled.
 pub struct MpMcQueue<T, const N: usize> {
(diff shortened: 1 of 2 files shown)


korken89 (approved):

korken89:
Thanks for fixing!

PR #247, opened 2021-10-01
Add an "order()" method for re-arranging the backing buffer of HistoryBuffer from oldest data to newest

dzimmanck:
This is a simple no-copy method for re-arranging the underlying buffer from oldest data to newest.  Very useful when used in from of as_slice(), which makes no such guarantees.

PR #253, opened 2021-11-12
Feature/historybuf ordered iterator

dzimmanck:
This adds a method called .ordered() to the HistoryBuffer which returns an iterator over the underlying buffer that starts from the oldest data and stops at the last data entered (regardless if the buffer is full).  This implementation will be in lieu of #247, which will be rejected.

It would be easy to modify this to be simple an implentation of the IntoIterator trait, which would allow it to be used in certain functions  and methods that expect a generic that implements such a trait.  However, given the slightly specialized nature of what it does I thought it more suitable to use a more descriptive name.


korken89:
bors try

korken89:
Could you also rebase on `master`?

PR #254, opened 2021-11-16
Implemented sealed traits according to Rust API Guidelines

eivindbergem:
They way the sealed trait `Kind` is implemented for `BinaryHeap` and `SortedLinkedList` prevents it from being used as bounds:

```Rust
struct SomeStruct<T, K, const N: usize>
where K: Kind {
    binary_heap: BinaryHeap<T, K, N>
}
```

As suggested by the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed), we can create a sealed trait by having an internal private trait that is required by the public trait. This way we can import the trait, but we can't implement it because of the private requirement.

This PR converts the sealed traits to this pattern. Because of a name conflict in `sorted_linked_list.rs` I renamed the generic type `Kind` to `K` throughout the module.


korken89:
Thanks for finding this, could you have a look at fixing the failing tests?

korken89 (approved):
Thanks for fixing

PR #262, opened 2021-12-28
add BinaryHeap::into_vec

Lambda-Logan:
std::collections::BinaryHeap provides an into_vec method https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.into_vec

My use case is I want a sorted iterator  similar to BinaryHeap.into_sorted_iter. I found my self creating a new heapless:Vec from heapless::BinaryHeap::into_iter and then sorting it. It seemed better to return the underlying vec and sort that.

But the functionality seemed generally useful and helps the API jive a little better with std::collections::BinaryHeap.

Do you see any issues with this implementation? Also, apologies for anything weird... This is my first PR :)

### DIFF
--- a/src/binary_heap.rs
+++ b/src/binary_heap.rs
@@ -100,7 +100,7 @@ impl private::Sealed for Min {}
 
 pub struct BinaryHeap<T, K, const N: usize> {
     pub(crate) _kind: PhantomData<K>,
-    pub(crate) data: ManuallyDrop<Vec<T, N>>,
+    pub(crate) data: Vec<T, N>,
 }
 
 impl<T, K, const N: usize> BinaryHeap<T, K, N> {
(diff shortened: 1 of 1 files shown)


japaric:
thanks for the PR! sorry that it felt through the cracks :-(

turns out there was a bug in the BinaryHeap implementation and it was dropping the elements twice. that was fixed in PR #287

with that change the implementation of `into_vec` should be the same as of the one in the standard library: `self.data` with any addition of `ManuallyDrop`

could you update the implementation? the current one is not running destructors after the change in PR #287

Lambda-Logan:
@japaric  Like this?

japaric (approved):
yes, thanks!

japaric:
release v0.7.13 includes this change

PR #264, opened 2022-01-14
Add support for AVR.

jeremysalwen:
In order to support AVR, the following changes were made:

 - AVR was added to the list of architectures not supporting atomics or cas.
 - AVR was configured to not use AtomicUsize, similarly to armv6m.
 - SortedLinkedList was modified to account for the fact that usize is differently sized on
   different architectures (16 bits on AVR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 ### Added
+- Added support for AVR architecture.
 
 ### Changed
 
--- a/src/sorted_linked_list.rs
+++ b/src/sorted_linked_list.rs
@@ -95,7 +95,7 @@ where
 
 // Internal macro for generating indexes for the linkedlist and const new for the linked list
 macro_rules! impl_index_and_const_new {
-    ($name:ident, $ty:ty, $new_name:ident, $max_val:literal) => {
+    ($name:ident, $ty:ty, $new_name:ident, $max_val:expr) => {
         /// Index for the [`SortedLinkedList`] with specific backing storage.
         #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
         pub struct $name($ty);
@@ -178,9 +178,9 @@ macro_rules! impl_index_and_const_new {
     };
 }
 
-impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, 254); // val is 2^8 - 2 (one less than max)
-impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, 65_534); // val is 2^16 - 2
-impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, 4_294_967_294); // val is 2^32 - 2
+impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, { u8::MAX as usize - 1 });
+impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, { u16::MAX as usize - 1 });
+impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, { usize::MAX - 1 });
 
 impl<T, Idx, K, const N: usize> SortedLinkedList<T, Idx, K, N>
 where


src/pool/singleton/arc.rs
```diff
@@ -81,10 +81,10 @@ use core::{
     sync::atomic,
 };
 
-#[cfg(armv6m)]
+#[cfg(any(armv6m, avr))]
```
TDHolmes, line 84: because `avr` doesn't have atomics, I don't think it will even try to include anything in `pool` so I don't think these changes are needed

build.rs
```diff
@@ -19,20 +19,27 @@ fn main() -> Result<(), Box<dyn Error>> {
...
         println!("cargo:rustc-cfg=armv7a");
+    } else if target.starts_with("avr-") {
+        println!("cargo:rustc-cfg=avr");
     }
-
     // built-in targets with no atomic / CAS support as of nightly-2019-12-17
     // see the `no-atomics.sh` / `no-cas.sh` script sitting next to this file
     match &target[..] {
-        "msp430-none-elf" | "riscv32i-unknown-none-elf" | "riscv32imc-unknown-none-elf" => {}
+        "avr-atmega328p"
```
TDHolmes, line 28: I think this target name should be `avr-unknown-gnu-atmega328` based on the mentioned scripts

src/pool/singleton/arc.rs
TDHolmes: I think you made these changes because of the incorrect target name in `build.rs`

build.rs
jeremysalwen: Done.

src/pool/singleton/arc.rs
jeremysalwen: Fixed.

korken89:
I think this was added in another PR by @TDHolmes . Can this be closed?

TDHolmes:
The changes in sorted linked list are still needed

korken89:
Okey, @jeremysalwen could you rebase this oh master?

jeremysalwen:
Rebased on master and updated the changelog entry to the new format.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Add support for AVR architecture.
```
TDHolmes, line 9: should this go under `### Added`?

korken89:
Indeed, please it under the added  header.

jeremysalwen:
Oops, forgot to push that change.  Should be good now.

TDHolmes:
@jeremysalwen additionally we need changes to `build.rs` and `Cargo.toml`. Can you follow the pattern of `riscv32i-unknown-none-elf` in both of those files?

jeremysalwen:
Hmm, the build file already includes the changes to support AVR, since I rebased on master. Has something else changed?  Are you saying that I need to change this PR so that avr uses atomic-polyfill? (is it even supported?)

TDHolmes:
yeah, it would be to have AVR use `atomic-polyfill`. It is supported ([see here](https://github.com/embassy-rs/atomic-polyfill/blob/main/build.rs#L35))

TDHolmes:
Looks like only recently! v0.1.7

jeremysalwen:
Hmm, it seems like matching the configuration of `riscv32i-unknown-none-elf` is now causing it to try compiling the files in `/pool`, which are failing.

src/sorted_linked_list.rs
```diff
@@ -178,9 +178,9 @@ macro_rules! impl_index_and_const_new {
     };
 }
 
-impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, 254); // val is 2^8 - 2 (one less than max)
-impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, 65_534); // val is 2^16 - 2
-impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, 4_294_967_294); // val is 2^32 - 2
+impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, { u8::MAX as usize - 1 });
+impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, { u16::MAX as usize - 1 });
+impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, { usize::MAX - 1 });
```
japaric, line 183: @korken89 could you confirm that this change is ok? it'll change `$max_val` on 64-bit targets. from looking at the code it doesn't seem like it'd be a problem given that the storage is also `usize`

japaric (approved):
these changes look fine to me. merge pending an answer from @korken89

src/sorted_linked_list.rs
korken89: This LGTM, I would say it's not an issue in practice

japaric:
thanks again for the PR
bors r+

PR #265, opened 2022-01-14
Add cas support for risc-v via atomic-polyfill and fix mpmc/pool documentation

TDHolmes:
I at first was only going to fix the documentation for `pool` and `mpmc`, clarifying that `thumbv6m` in fact _does_ include these modules via the `atomic-polyfill` crate, but then noticed that we could add the two `riscv` targets as they also have `atomic-polyfill` support as of `v0.1.4`. `MSP430` does not have support at this time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `OldestOrdered` iterator for `HistoryBuffer`
 
+### Changed
+
+- `atomic-polyfill` is now enabled and used for `cas` atomic emulation on `riscv` targets
+
 ## [v0.7.9] - 2021-12-16
 
 ### Fixed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,6 @@ version = "0.1"
 [dependencies.defmt]
 version = ">=0.2.0,<0.4"
 optional = true
+
+[package.metadata.docs.rs]
+all-features = true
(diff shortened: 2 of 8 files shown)


TDHolmes:
alright, now the docs are all looking good and I've tested a `thumbv6m` and `thumbv7m` build and they're building the right things. This is good to go.

If the merge strategy isn't squash let me know and I can clean my history up a bit before merging...

TDHolmes:
What is the real logic of this check?
```
#[cfg(any(
    armv6m,
    armv7a,
    armv7r,
    armv7m,
    armv8m_main,
    all(
        any(target_arch = "x86_64", target_arch = "x86"),
        feature = "x86-sync-pool"
    ),
    test
))]
```

I'm wondering if this should be changed to:
```
#[cfg(any(
    use_atomic_polyfill,
    armv7a,
    armv7r,
    armv7m,
    armv8m_main,
    all(
        any(target_arch = "x86_64", target_arch = "x86"),
        feature = "x86-sync-pool"
    ),
    test
))]
```
but I didn't want to assume

TDHolmes:
tested locally now with risc-v (full polyfill), thumbv6m (cas polyfill) and thumbv7m (full native) and we compile

korken89:
Hi, thanks for the effort!
Can you add the new targets to the CI?

korken89:
Ah wait, we already added it.

TDHolmes:
done :+1:

korken89 (approved):

PR #266, opened 2022-01-16
Feature/historybuf `OldestOrdered` iterator

finnbear:
My use case requires an oldest..newest iterator. I tried https://github.com/japaric/heapless/pull/253 but ran into a crash when the buffer was empty (len = 0).

This PR:
- fixes the crash when len is 0
- adds more test cases
- addresses https://github.com/japaric/heapless/pull/253#issuecomment-1012914870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


korken89:
Thanks, looks good. Could you have a look at the failing test?
It asks for a line in the CHANGELOG.

korken89 (approved):
bors merge

PR #267, opened 2022-01-16
assume native compilation if target_os is not "none" to allow non-x86 hosts to run tests

TDHolmes:
I have an M1 Mac so the x86 specific test prevented me from running `cargo test`. I'm not sure why `scoped_threadpool` isn't just a plain old [`dev-dependency]` though... should I just make it an unqualified `[dev-dependency]` instead?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `cargo test` can now run on non-`x86` hosts
+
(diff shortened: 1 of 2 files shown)


TDHolmes:
@korken89 do you think this requires a CHANGELOG entry or is it too minor of a change?

korken89:
Hi, all changes require a CHANGELOG entry.

TDHolmes:
Wasn't sure if it was a "notable change" or not per the changelog header.
I'll add an entry

korken89 (approved):
LGTM, thanks!

PR #272, opened 2022-03-07
Gate `spsc` under the `cas` feature (Fixes #271)

### LINKED ISSUE #271: Doesn't build on riscv32i-unknown-none-elf without default features (regression 0.7.10)
When adding `heapless` as a dependency:
```toml
[dependencies]
heapless = { version = "=0.7.10", default-features = false, features = [] }
```
it does not build:
```
   Compiling heapless v0.7.10
error[E0432]: unresolved import `atomic_polyfill`
  --> /home/kriskras99/.cargo/registry/src/github.com-1ecc6299db9ec823/heapless-0.7.10/src/spsc.rs:90:5
   |
90 | use atomic_polyfill::{AtomicUsize, Ordering};
   |     ^^^^^^^^^^^^^^^ use of undeclared crate or module `atomic_polyfill`

For more information about this error, try `rustc --explain E0432`.
```
This regression was introduced with version `0.7.10` (aka it works with `version = "=0.7.9"`)
Also see [the Rust Crypto issue](https://github.com/RustCrypto/AEADs/issues/398).

Kriskras99:
The `spsc` module should only be built if the `cas` feature is enabled.
This fixes issue #271.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,7 +106,7 @@ pub mod mpmc;
 #[cfg(all(has_cas, feature = "cas"))]
 pub mod pool;
 pub mod sorted_linked_list;
-#[cfg(has_atomics)]
+#[cfg(all(has_atomics, feature = "cas"))]
 pub mod spsc;
 
 #[cfg(feature = "ufmt-impl")]


TDHolmes:
hmm, but `spsc` doesn't actually require `cas` operations. What we really need I think is to make `atomic-polyfill` a requirement for targets that have no atomic support (i.e., have `full_atomic_polyfill` cfg set). Does dropping `optional = true` like below fix it?
```
[target.riscv32i-unknown-none-elf.dependencies]
 atomic-polyfill = { version = "0.1.4" }

 [target.riscv32imc-unknown-none-elf.dependencies]
 atomic-polyfill = { version = "0.1.4" }
```

Kriskras99:
I would suggest adding an atomic load/store gate and hide it behind that.
See 2a8f38c8d1633d1f6c7e875588d035f7b0062136.
In this way people who don't care for atomic types don't have to pull in 12 extra dependencies.

Kriskras99:
I have created an alternative pull request in #273 .
@TDHolmes you're solution also works but I'd like to avoid pulling in `atomic_polyfill` if I can.

PR #273, opened 2022-03-08
Add feature gate for atomic load/store, gate `spsc` behind `loadstore` (Alternative fix for #271)

Kriskras99:
The `spsc` module should only be build when `atomic_polyfill` is available. It's a big dependency (pulls in 12 extra dependencies) so for projects that don't need the atomics it's better to hide it behind a (default) feature gate.

I added two new features:
* `loadstore` that pulls in `atomic_polyfill` for architectures that need it and gates `spsc`
* `atomics` that enables `loadstone` and `cas` (replaces `cas` as the default feature)

Doing it this way should allow for more granular configuration of architectures, i.e. only pull-in `atomic_polyfill` if an architecture uses `loadstore` and doesn't have native support. This pull request does not contain that granularity as I'm not familiar with other architectures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,10 @@ repository = "https://github.com/japaric/heapless"
 version = "0.7.10"
 
 [features]
-default = ["cas"]
+default = ["atomics"]
+atomics = ["loadstore", "cas"]
 cas = ["atomic-polyfill"]
+loadstore = ["atomic-polyfill"]
 ufmt-impl = ["ufmt-write"]
 # read the docs before enabling: makes `Pool` Sync on x86_64
 x86-sync-pool = []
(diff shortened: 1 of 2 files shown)


TDHolmes:
But this will now unnecessarily pull in `atomic-polyfill` even for targets with native atomic support

Kriskras99:
> It only gets pulled in for targets which have 'atomic_polufill' as an (optional) dependency. For targets that support load store natively you can use cfgs in the Cargo.toml to only use 'atomic_polyfill' for 'cas'

This is slightly wrong. It's true that specifying `atomic_polyfill` as a feature only pulls it in as an optional dependency. However it's not possible to only pull it in for `cas` for some architectures. For that we'd need something [like this](https://github.com/rust-lang/cargo/issues/1839).

PR #275, opened 2022-03-23
Add intoiter for Indexmap

MarcusGrass:
I ran up against some problems trying to consume Indexmaps. 

Fairly low-effort implementation of IntoIter maybe ordering should be commented but I think that when iterating over a map ordering can be assumed to be _implementation specific_ and not promising an ordering is better for future implementation flexibility.

### DIFF
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -857,6 +857,34 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


japaric:
only rustfmt failed. I'll run it locally.

PR #276, opened 2022-03-23
Add entry API for maps

MarcusGrass:
Fixes https://github.com/japaric/heapless/issues/103 
Also fixes https://github.com/japaric/heapless/issues/258

It's a naive implementation where an entry holds a reference to the underlying CoreMap and indirectly modifies that.
To make that modification smoother I merged `insert_phase_1` and `insert_phase_2` so it was easier to return the underlying vector index which makes internal Entry-access more efficient.

It only implements Occupied and Vacant entry's corresponding stable methods in std, the currently unstable methods could be implemented for this library immediately if you'd like to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Fixed `pool` example in docstring.
 * Fixed undefined behavior in `Vec::truncate()`, `Vec::swap_remove_unchecked()`,
   and `Hole::move_to()` (internal to the binary heap implementation).
+* Inserting an item that replaces an already present item will no longer 
+fail with an error
 
 ### Added
+
+* Add Entry Api to IndexMap
 
 ### Changed
 
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -98,7 +98,11 @@ impl Pos {
     }
 }
 
-struct Insert<V> {
+enum Insert<K, V> {
+    Success(Inserted<V>),
+    Full((K, V)),
+}
+struct Inserted<V> {
     index: usize,
     old_value: Option<V>,
 }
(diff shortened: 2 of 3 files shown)


MarcusGrass:
After benchmarking (Only locally so take that with a grain of salt).
First: The map implementation contains a "bug", might be an undocumented implementation detail. If overwriting a map-key using insert on a full map you get an Error back because the capacity check happens before checking if the key exists.
Ignoring that error and running benchmarks `if map.contains_key... map.insert...` performs slightly worse than `if let Entry::Occupied` for small maps, with the performance difference getting larger as the map grows.
Not ignoring that error and using
``` Rust
if map.contains_key(black_box(&i)) {
    *map.get_mut(black_box(&i)).unwrap() = black_box(i + 1);
}
```
is more than two times slower than 
``` Rust
if let heapless::Entry::Occupied(o) = map.entry(black_box(i)) {
    o.insert(black_box(i + 1));
}
```
makes the entry implementation a bit over twice as fast.

MarcusGrass:
Drive-by also fixes https://github.com/japaric/heapless/issues/258
Still just one capacity check, but it's not as eager

korken89:
Very nice addition, I'll look into merging with the next breaking change as this does impact user facing usage.
Though I'd agree with you that the insert on full is probably technically a bug.

Could you add a line in the CHANGELOG as well?

fralalonde:
@MarcusGrass 

That's a great PR, the lack of entry() API is painful.

FWIW I do not hesitate to use unsafe `.unwrap_unchecked` here since the call cannot fail as we just verified that the key exists. This might make things a bit faster but the code is heavier still.

```
if map.contains_key(black_box(&i)) {
    unsafe {*map.get_mut(black_box(&i)).unwrap_uncheck() = black_box(i + 1); }
}
```

MarcusGrass:
> Very nice addition, I'll look into merging with the next breaking change as this does impact user facing usage. Though I'd agree with you that the insert on full is probably technically a bug.
> 
> Could you add a line in the CHANGELOG as well?

Sure thing!

PR #282, opened 2022-04-29
update TSAN supressions file

japaric:
to catch false positives in newer version of scoped_threadpool

this change is required to prevent this snippet triggering a TSAN error (data race)
``` rust
// scoped_threadpool = "0.1.9"

use scoped_threadpool::Pool;

fn main() {
    Pool::new(2).scoped(move |scope| {
        scope.execute(move || {});
        scope.execute(move || {});
    });
}
```

### DIFF
--- a/suppressions.txt
+++ b/suppressions.txt
@@ -4,4 +4,7 @@ race:*drop_slow*
 race:__call_tls_dtors
 
 # false positives in scoped_threadpool (?)
-race:*drop*
+race:call_once*scoped_threadpool
+race:drop*scoped_threadpool
+race:scoped_threadpool*cgu
+race:scoped_threadpool::


japaric:
the supressions seem to be too heavy handed.

this triggers a data race as expected (debug profile)
``` rust
use std::thread;

static mut X: i32 = 0;

let t1 = thread::spawn(|| unsafe { ptr::addr_of_mut!(X).write_volatile(1) });
unsafe {
    ptr::addr_of_mut!(X).write_volatile(2);
}
t1.join().ok();
```

``` text
WARNING: ThreadSanitizer: data race (pid=45823)
(..)
Location is global 'foo::main::X::h5af7d740da9a743d' of size 4
```

with the above supressions, this does NOT trigger a data race

``` rust
Pool::new(2).scoped(move |scope| {
    for i in 0..2 {
        scope.execute(move || unsafe { ptr::addr_of_mut!(X).write_volatile(i) });
    }
});
```

without the supressions file, the false positives seem to mask the real data race because the variable `X` is not included in the reported data races

it may be best to remove the `scoped_threadpool` dependency from the tests ...

japaric:
closing in favor of PR #283

PR #283, opened 2022-04-29
use std::thread::scope instead of scoped_threadpool

japaric:
as it's easier to deal with TSAN false positives in the former API

as it surfaced in PR #280 the current supression rules don't handle newer versions of the
`scoped_threadpool` crate

trying to update the supression rules related to `scoped_threadpool` in PR #282 revealed that the
supression rules are masking (hiding) real data races:
https://github.com/japaric/heapless/pull/282#issuecomment-1113173358

`std::thread::scope` requires less supression rules and does not mask real data races -- for instance,
the data race in the linked issue comment is not masked when using `std::thread::scope`

tradeoffs:
- pro: one less dev dependency but added a small build dependency
- pro: supressions file is simpler
- cons: `std::thread::scope` is only available on recent nightlies but the API is only used in tests so it doesn't affect the MSRV

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,9 +27,6 @@ mpmc_large = []
 # This flag has no version guarantee, the `defmt` dependency can be updated in a patch release
 defmt-impl = ["defmt"]
 
-[target.'cfg(not(target_os = "none"))'.dev-dependencies]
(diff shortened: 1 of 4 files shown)


korken89 (approved):
bors merge

PR #284, opened 2022-04-29
[breaking-change] use hash32 v0.3.0

japaric:
this release of `hash32` has the advantage that 32-bit hashers can be used to hash types that
implement the `core::hash::Hash` trait removing the need for the `hash32::Hash` implementations in
this crate and the uses of the `#[derive(Hash32)]` macro (which did not support enums) in dependent
crates

with this change the following code works
``` rust
// NOTE no derive(Hash32)
#[derive(Hash)]
struct Int(i32);

let mut x = FnvIndexSet::<_, 4>::default();
let _ = x.insert(Int(0));
```

this change is technically a breaking change because the following code is no longer accepted

``` rust
// assume this type comes from a dependency
// NOTE no derive(Hash)
#[derive(Hash32)]
struct Int(i32);

let mut x = FnvIndexSet::<_, 4>::default();
let _ = x.insert(Int(0)); // error: does not implement Hash
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [breaking-change] `IndexMap` and `IndexSet` now require that keys implement the `core::hash::Hash`
+  trait instead of the `hash32::Hash` (v0.2.0) trait
+
 ### Fixed
 
 ## [v0.7.16] - 2022-08-09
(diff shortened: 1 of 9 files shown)


korken89:
Nice improvement!

pwnorbitals:
Awesome ! Can’t wait to have it merged :)

japaric:
bors merge

japaric:
should work now that MSRV checks are gone
bors merge

pwnorbitals:
Thanks Japaric, awesome !

PR #297, opened 2022-06-05
Add Vec::from_array_unchecked

Bajix:
I'm working on a new version of [swap_queue](https://crates.io/crates/swap-queue) that integrates with `heapless::Vec`, but for my use case I want to be able to construct this by passing ownership of a buffer  `[MaybeUninit<T>; M]` so that there's no cloning. Please add this behavior and feel free to make any changes per your discretion

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -92,6 +92,16 @@ impl<T, const N: usize> Vec<T, N> {
         Ok(v)
     }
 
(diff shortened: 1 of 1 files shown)


vrmiguel:
Are you sure that this does not cause a stack-on-stack memcpy?

PR #298, opened 2022-06-14
compile-time check power of two invariant in Index{Set,Map}'s default method

### LINKED ISSUE #295: compile-time assertions are not applied to the `default` constructor
this is correctly rejected by the compiler (because the capacity is not a power of 2)
``` rust
let _: FnvIndexSet<(), 3> = FnvIndexSet::new();
```

this is incorrectly accepted and should be fixed
``` rust
let _: FnvIndexSet<(), 3> = FnvIndexSet::default();
```

japaric:
closes #295

I tested this manually with both `FnvIndexSet` and `FnvIndexMap`. I tried to add a compile fail test but discovered (the hard way) that all the const assert trickery in `sealed` is checked when the code is built with `cargo build` but *not* when the code is built with `cargo check`. As `trybuild` uses `cargo check` under the hood it's not possible to compile fail test any of those const asserts.

r? @korken89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- `IndexSet` and `IndexMap`'s `default` method now compile time checks that their capacity is a power of two.
+
 ## [v0.7.13] - 2022-05-16
 
 ### Added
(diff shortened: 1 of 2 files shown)


korken89 (approved):
bors merge

japaric:
PR #299 should have fixed CI
bors r=korken89

PR #301, opened 2022-07-04
Add some more Vec methods.

Dirbaio:
Added

- `Vec::insert(index, element)`
- `Vec::remove(index)`
- `Vec::retain(f)`
- `Vec::retain_mut(f)`

Behavior matches `std` except `insert` which is now fallible and returns back the element when full. Implementation and docs taken from `std`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


japaric (approved):
thanks!

japaric:
release v0.7.15 includes this PR

Dirbaio:
So fast! :rocket: 

Thank you! :D

PR #302, opened 2022-07-05
Add clarification in String

mriise:
Needed to look into implementation to be sure String was allocating bytes not chars.


japaric (approved):
thanks!

PR #303, opened 2022-07-17
Vec - PartialEq where Rhs is the Vec

elpiel:
I was using the Vec for tests and initially, I was confused because I saw the impl for PartialEq but then I realized that PartialEq with Rhs of Vec was not implemented.

These impls are not mandatory but might help other people when using Vec on Rhs.

PS: Amazing crate btw, thanks you for the awesome work!

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -1024,6 +1024,16 @@ where
     }
 }
 
+// [B] == Vec<A, N>
+impl<A, B, const N: usize> PartialEq<Vec<A, N>> for [B]
+where
+    A: PartialEq<B>,
+{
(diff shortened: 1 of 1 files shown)


japaric (approved):
awesome addition, thanks!

elpiel:
Should I also add a Changelog entry because I see that the CI is failing?

japaric:
> Should I also add a Changelog entry because I see that the CI is failing?

the changelog being missing is not a showstopper -- the CI check can be skipped. if you can update the changelog in this PR, that'd be great

japaric:
v0.7.16 has been published and includes these changes

PR #306, opened 2022-08-03
Add `String::from_utf16`.

reitermarkus:
Add equivalent to `std::string::String::from_utf16` and additional `push_utf16` helper method.

### DIFF
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -12,7 +12,7 @@ use hash32::{BuildHasherDefault, FnvHasher};
 
 use crate::Vec;
 
-/// A [`IndexMap`] using the default FNV hasher
+/// An [`IndexMap`] using the default FNV hasher.
 ///
 /// A list of all Methods and Traits available for `FnvIndexMap` can be found in
 /// the [`IndexMap`] documentation.
@@ -681,7 +681,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Get the first key-value pair
     ///
-    /// Computes in **O(1)** time
+    /// Computes in *O*(1) time
     pub fn first(&self) -> Option<(&K, &V)> {
         self.core
             .entries
@@ -691,7 +691,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Get the first key-value pair, with mutable access to the value
     ///
-    /// Computes in **O(1)** time
+    /// Computes in *O*(1) time
     pub fn first_mut(&mut self) -> Option<(&K, &mut V)> {
         self.core
             .entries
@@ -701,7 +701,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Get the last key-value pair
     ///
-    /// Computes in **O(1)** time
+    /// Computes in *O*(1) time
     pub fn last(&self) -> Option<(&K, &V)> {
         self.core
             .entries
@@ -711,7 +711,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Get the last key-value pair, with mutable access to the value
     ///
-    /// Computes in **O(1)** time
+    /// Computes in *O*(1) time
     pub fn last_mut(&mut self) -> Option<(&K, &mut V)> {
         self.core
             .entries
@@ -721,7 +721,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Return the number of key-value pairs in the map.
     ///
-    /// Computes in **O(1)** time.
+    /// Computes in *O*(1) time.
     ///
     /// ```
     /// use heapless::FnvIndexMap;
@@ -737,7 +737,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Returns true if the map contains no elements.
     ///
-    /// Computes in **O(1)** time.
+    /// Computes in *O*(1) time.
     ///
     /// ```
     /// use heapless::FnvIndexMap;
@@ -753,7 +753,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Remove all key-value pairs in the map, while preserving its capacity.
     ///
-    /// Computes in **O(n)** time.
+    /// Computes in *O*(n) time.
     ///
     /// ```
     /// use heapless::FnvIndexMap;
@@ -815,7 +815,7 @@ where
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form *must* match those for the key type.
     ///
-    /// Computes in **O(1)** time (average).
+    /// Computes in *O*(1) time (average).
     ///
     /// ```
     /// use heapless::FnvIndexMap;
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -94,7 +94,7 @@ mod histbuf;
 mod indexmap;
 mod indexset;
 mod linear_map;
-mod string;
+pub mod string;
```
burrbull, line 97: docs for module

src/string.rs
```diff
@@ -140,6 +190,43 @@ impl<const N: usize> String<N> {
...
+    /// assert_eq!(s, "𝄞music");
+    ///
+    /// // 𝄞mu<invalid>ic
+    /// let mut s: String<10> = String::new();
+    /// let v = &[0xD834, 0xDD1E, 0x006d, 0x0075,
+    ///           0xD800, 0x0069, 0x0063];
+    /// assert!(s.push_utf16(v).is_err());
+    /// ```
+    pub fn push_utf16(&mut self, v: &[u16]) -> Result<(), Error> {
+        for c in decode_utf16(v.iter().cloned()) {
```
japaric, line 216: nit: could you please use a qualified function call here? i.e. `char::decode_utf16`

src/string.rs
```diff
@@ -140,6 +190,43 @@ impl<const N: usize> String<N> {
...
+    /// s.push_utf16(v).unwrap();
+    /// assert_eq!(s, "𝄞music");
+    ///
+    /// // 𝄞mu<invalid>ic
+    /// let mut s: String<10> = String::new();
+    /// let v = &[0xD834, 0xDD1E, 0x006d, 0x0075,
+    ///           0xD800, 0x0069, 0x0063];
+    /// assert!(s.push_utf16(v).is_err());
+    /// ```
+    pub fn push_utf16(&mut self, v: &[u16]) -> Result<(), Error> {
```
japaric, line 215: this is not an API provided in the standard library. I'd like to keep the API surface of `heapless::String` as close as libstd's. could you please make this method private?

japaric (changes requested):
Thanks for the PR! I'd like to keep `push_utf16` private for now but otherwise this looks good. could you please address my inline comments and update the changelog?

src/string.rs
reitermarkus: I specifically need this actually for building a file path string while walking through a FAT32 volume.

src/lib.rs
reitermarkus: What should be documented here?

src/string.rs
japaric: this is not a method that requires access to private parts of `String` so it can be implemented out of tree, either as a free function or as an extension method.

I'd still prefer to omit this. it'd be problematic if upstream libstd adds a method named the same that has a different signature or semantics, or add this functionality under a different name.

src/string.rs
```diff
@@ -1,8 +1,30 @@
...
+    fmt::Write,
+    hash, iter, ops, str,
+};
 
 use crate::Vec;
 
-/// A fixed capacity [`String`](https://doc.rust-lang.org/std/string/struct.String.html)
+/// The error type for [`String`] operations.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Error {
```
japaric, line 15: this should be called `FromUtf16Error` and its documentation should indicate that's it's only used in the `from_utf16` method. otherwise, it sounds like this `Error` type is going to be used in all of String's fallible API, which is not the case.

src/string.rs
reitermarkus: > this is not a method that requires access to private parts of String so it can be implemented out of tree, either as a free function or as an extension method.

In that case I would have to essentially copy-paste this function, which seems unnecessary.

> if upstream libstd adds a method named the same that has a different signature or semantics

If this happens, you can bump the version of this crate. In the meantime, this provides useful functionality.

> or add this functionality under a different name

If this happens, worst case is the different name needs to be added and this one deprecated.

reitermarkus: I'd be fine if this was behind a feature, e.g. `non-libstd-methods`. I just don't want to reimplement the wheel and make this PR useless to me.

japaric: > In that case I would have to essentially copy-paste this function, which seems unnecessary.

you would be maintaining code that you need; that seems normal to me. if you needed this extra API on `std::Vec` you would likely do the same.

> I'd be fine if this was behind a feature, e.g. non-libstd-methods

that would increase the maintenance work: the new Cargo feature would need to be tested in CI, in combination with other Cargo features. it also doesn't address the API issues; API behind a Cargo feature still need to adhere to semver.

reitermarkus:
@japaric, I have removed the `push_utf16` method now.

src/string.rs
```diff
@@ -10,13 +11,26 @@ use core::{
 
 use crate::Vec;
 
-/// A fixed capacity [`String`](https://doc.rust-lang.org/std/string/struct.String.html)
+/// A possible error value when converting a [`String`] from a UTF-16 byte slice.
+///
+/// This type is the error type for the [`from_utf16`] method on [`String`].
+///
+/// [`from_utf16`]: String::from_utf16
+#[derive(Debug, PartialEq, Eq)]
```
newAM, line 19:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]
```

Matches the `DecodeUtf16Error` struct.

newAM (changes requested):
Looks good to me, thanks for the feature & documentation update!  Just one small change that I think it worth making.

src/indexmap.rs
```diff
@@ -681,7 +681,7 @@ impl<K, V, S, const N: usize> IndexMap<K, V, S, N> {
 
     /// Get the first key-value pair
     ///
-    /// Computes in **O(1)** time
+    /// Computes in *O*(1)* time
```
Dirbaio, line 684: there's an extra `*` after the `()`.

src/lib.rs
```diff
@@ -97,7 +97,7 @@ mod histbuf;
...
```
Dirbaio, line 100: why make this module public? String is already reexported at the crate root.

newAM: That's for the new `FromUtf16Error` type.

This does bring up a good point though, `string::format` should have `#[doc(hidden)]` to keep it out-of-sight if this module is public.

Dirbaio: ah okay. it feels a bit strange to me that you can now refer to `String` via either `heapless::String` or `heapless::string::String` though. Maybe we should remove the reexport from the crate root, so all string stuff is available on `heapless::string::*` only? It's a breaking change though, and we just released v0.8.

(and maybe we should do it for other stuff. the map stuff is cluttering the crate root quite a bit too)

newAM: I think that's a good idea, but for now it doesn't hurt to have it available from two locations, and it keeps it non-breaking.

Dirbaio: okay, sounds good to me then.

src/string.rs
reitermarkus: https://doc.rust-lang.org/std/string/struct.FromUtf16Error.html doesn't have `Clone` though.

reitermarkus: It also doesn't have `PartialEq` and `Eq` for that matter.

newAM: I was thinking it makes sense to match the inner type of [`DecodeUtf16Error`](https://doc.rust-lang.org/core/char/struct.DecodeUtf16Error.html) for this, since it is separate from `std::string::FromUtf16Error` by necessity of having the `Capacity` variant.

reitermarkus: Given that `std` doesn't have these I don't think these traits are necessary for this type. I think it makes more sense to follow `std` here given that I already had to remove the `push_utf16` method here which was the original incentive for this PR.

newAM: Fair enough, sounds good to me!

newAM (approved):

PR #311, opened 2022-08-12
ci: run miri tests first

japaric:
it seems runners are assigned to jobs in the order they appear in the
yaml file. miri takes the longest so it's better if it appears first

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,46 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  # Run MIRI tests on nightly
+  # NOTE first because it takes the longest to complete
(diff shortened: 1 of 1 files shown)


japaric:
bors merge

.github/workflows/build.yml
```diff
@@ -3,12 +3,52 @@ on:
...
   push:
-    branches: [main, staging, trying]
+    branches: [staging, trying]
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  # Run MIRI tests on nightly
+  # NOTE first because it takes the longest to complete
```
japaric, line 13: seems like the runners are assigned to jobs at random so this didn't help :-(

PR #313, opened 2022-08-24
switch to atomic-polyfill-1.0.0 on AVR to get around old code in critical-section

mutantbob:
Should fix https://github.com/japaric/heapless/issues/312

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ atomic-polyfill = { version = "0.1.4" }
 atomic-polyfill = { version = "0.1.4" }
 
 [target.'cfg(target_arch = "avr")'.dependencies]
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -37,10 +37,10 @@ atomic-polyfill = { version = "0.1.4" }
 atomic-polyfill = { version = "0.1.4" }
 
 [target.'cfg(target_arch = "avr")'.dependencies]
-atomic-polyfill = { version = "0.1.8", optional = true }
```
ryan-summers, line 40: Can we release a patch `heapless` with _just_ the `atomic-polyfill` update? This would allow us to release an updated `heapless` that still works with other heapless 0.7 releases.

Issue #180, opened 2020-10-18
The llsc stack is not safe from ABA with how LL/SC is used


rustonaut:
> On the Cortex-M, STREX will always fail if the processor takes an exception between it and its corresponding LDREX operation 

But the load of `next` at `8000136` is done outside of the LDREX...STREX sequence.

So if the process is preempted between `8000136`  and `800013a` and during the
preemption we cause the ABA problem (e.g. given h->A->B->C->0, do pop A, pop B, push A resulting in h->A->C->0 with B being in use).

Now given Cortex-M implementations might not have that problem:

> A context switch might cause a subsequent Store-Exclusive to fail, requiring a load ... store sequence to be replayed. To minimize the possibility of this happening, ARM recommends that the Store-Exclusive instruction is kept as close as possible to the associated Load-Exclusive instruction, see Load-Exclusive and Store-Exclusive usage restrictions. 

(quote from the ARM®v7-M ArchitectureReference Manual, linked in the sourcecode)

The problem is this is a *might* so depending on where you get your Cortex-M chip from this might or might not be a sound implementation.

---

https://github.com/japaric/heapless/blob/9ff3a5fa8968fb263cece456ccc9505dc913147e/src/pool/mod.rs#L123-L139

rustonaut:
As a side not, I don't know which of the current sold Cortex-M based processors have that problem, neither do I know it there are any.

So this is a specification related bug, **but it might not be a bug in the real world**.

talchas:
No, it's **absolutely** a problem in practice. The load of next is outside of the llsc region, so it doesn't matter about the "might" in the spec. A preemption before that (or concurrent accesses in actual SMP hardware) is not going to trigger an abort. Taking advantage of LL/SC like this is just not possible with C11-based intrinsics - you need the LL(head) part of compare_exchange_weak before the load of next. This is doable in ASM, but not in the typical rust/C/etc intrinsics. The asm shown in the docs is clearly shows how this is incorrect, in order for it to be ok it would need to have reordered the `ldrex`/`lrd.w` instructions (and then had different register allocation):

```
ldrex   r3, [r0] 
ldr.w   ip, [r2] 
cmp     r3, r1 
bne.n   800014a <<heapless::pool::Pool<T>>::pop+0x1a> 
strex   r3, ip, [r0] 
```

Meanwhile the warned-about x86-64 generation counter algorithm is safe in basically any realistic scenario.

korken89:
[...]

PR #315, opened 2022-09-07
revamp pool API

### LINKED ISSUE #167: Pool: Store the pointer to the next node and data in the same memory location.
Instead of having:

```rust
pub struct Node<T> {
    next: AtomicPtr<Node<T>>,
    pub(crate) data: UnsafeCell<T>,
}
```

We could just use one location to store both things, since we don't need to know both at the same time. I made a small PoC of this:

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e22288ddcb535929e939ec891749c7d1)

I still have to check a few things, but I would like some opinions about this, to see if it's really sound and worth it. I then can adapt this to heapless' code and make a PR if there's a disposition to merge.

### LINKED ISSUE #286: [RFC] spring clean (revamp) the Pool API
All memory Pools are backed by a Treiber stack, which is effectively a linked list (an intrusive collection). as of v0.7.x, The "node" of this linked list (struct shown below) stores the data *next* to the "next" pointer, which points to the next item in the linked list. This results in each node having a size of `sizeof(T) + max(sizeof(usize), alignof(T))` bytes

``` rust
// treiber stack (linked list) node
struct Node<T> {
    // in the CAS variant, this is AtomicPtr (still pointer sized)
    next: Option<NonNull<T>>,
    data: MaybeUninit<T>,
}
```

This RFC proposes to change the layout of Node to be an union (shown below) to reduce the footprint of Node to `max(sizeof(T), sizeof(usize))`

``` rust
union Node<T> {
   next: Option<NonNull<T>>,
   data: MaybeUninit<T>,
}
```

(also see #167)

This document also proposes changing the existing Pool API to make it closer to the API of Box the standard library; and adding an `ObjectPool` API that manages initialized objects that are never destroyed

Finally, this document proposes removing the `x86-sync-pool` Cargo feature and providing the family of `Pool` APIs **only** on targets where it can be implemented in a sound lock-free manner, i.e. x86 and targets that have LL/SC operations, like ARM and RISC-V

## API changes

Summary:
- `Box` loses its `Uninit` type state.
  - rationale: the type state can be layered on top of `Box<T>`, e.g. `Box<Initialized<T, Yes>>` / `Box<Initialized<T, No>>`
- remove the deprecated `Box::freeze` API
- rename `Pool` to `BoxPool`
- rename `arc::Pool` to `ArcPool`
- `grow` is removed and `grow_exact` gets renamed to `manage`
- `Node` gets renamed to `MemoryBlock`
- add `ObjectPool` API

### Changes in existing API

the following snippets highlight the changes. the snippets are using the non-singleton API for simplicity; v0.8.x will also include a singleton version of the API

common code

``` rust
struct Thing {
    data: [u8; 128],
}

impl Thing {
    fn new() -> Self {
        Self { data: [0; 128] }
    }
}

impl Drop for Thing {
    fn drop(&mut self) {
        todo!() // side effect
    }
}
```

current / v0.7.x

``` rust
static P: Pool<Thing> = Pool::new();

static mut BLOCKS: MaybeUninit<[Node<Thing>; 8]> = MaybeUninit::uninit();
// emulate cortex-m-rt::entry transform
let blocks: &'static mut _ = unsafe { &mut BLOCKS };

P.grow_exact(blocks);

let uninit: Box<T, Uninit> = P.alloc().unwrap();
let mut x: Box<T, Init> = uninit.init(Thing::new());
// .. do stuff ..
P.free(x); // Thing's destructor runs
```

proposed / v0.8.x

``` rust
static P: BoxPool<Thing> = BoxPool::new();

const BLOCK: MemoryBlock<Thing> = MemoryBlock::new();
static mut BLOCKS: [MemoryBlock<Thing>; 8] = [BLOCKS; 8];
// emulate cortex-m-rt::entry transform
let blocks: &'static mut _ = unsafe { &mut BLOCKS };

for block in blocks {
    P.manage(block);
}

let mut x: Box<Thing> = P.alloc(Thing::new()).ok().unwrap();
// .. do stuff ..
P.free(x); // Thing's destructor runs
```

### new `ObjectPool` API

`ObjectPool` manages already initialized objects. Objects managed by this pool will never be `drop`-ped. The main use case is reusing POD (Plain Old Data) types, like arrays of bytes, without re-initializing the object every time is requested. Example:

``` rust
static P: ObjectPool<[u8; 128]> = ObjectPool::new();

const BLOCK: ObjectBlock<[u8; 128]> = ObjectBlock::new([0; 128]);
static mut BLOCKS: [ObjectBlock<[u8; 128]>; 8] = [BLOCK; 8];
// emulate cortex-m-rt::entry transform
let blocks: &'static mut _ = unsafe { &mut BLOCKS };

for block in blocks {
    P.manage(block);
}

// NOTE no initialization required
let mut x: Object<[u8; 128]> = P.request().unwrap();
// .. do stuff ..
P.return_(x); // no destructor runs here
```

## Sketch implementation

the singleton API has been omitted for simplicity. See https://gist.github.com/japaric/cb0d6e460b6253f17f954557c4ab0baa

thoughts @korken89 ?

### LINKED ISSUE #318: Heads-up: const_err lint is going away
This crate has a `deny(const_err)`. That is a no-op in current compilers, `const_err` is deny-by-default (and has been for years). In the near future, the lint will be [turned into a hard error](https://github.com/rust-lang/rust/pull/102091), at which point this crate will fail to build since the `deny(warnings)` means that the warning due to the removed lint becomes an error.

The `deny(const_err)` should probably just be removed.

japaric:
closes #286 
closes #167 
closes #318


korken89:
This is super, so much nicer API! :D

japaric:
still need to grapple with CI but review / feedback is welcome at this point. I think it's easiest to read through the API docs and examples and just try it out

korken89:
One thing I noticed, is the support for thumb v6 dropped?

japaric:
I didn't have time to look into that. CAS semantics does not produce a sound treiber stack because of the ABA problem; you need LL/SC intrinsics for a sound implementation (see #180).

I don't know if the stack would be sound with pseudo (non-lock-free) CAS wrapped in a critical section.

korken89:
Hmm that's unfortunate, using it (even with critical sections) was a nice feature do it worked anywhere.

I don't have time to review everything right now, but I'll have a look soon :)

japaric:
bors merge

PR #317, opened 2022-09-23
impl {Clone, PartialEq} for HistoryBuffer {...}

finnbear:
- `impl Clone for HistoryBuffer`
- `impl PartialEq for HistoryBuffer`

Both are tested fairly rigorously.

Motivation is to use this type in a `yew::Properties` struct, which requires both traits to be implemented.


finnbear:
Note sure why the `testcfail` check fails. When I check the diff, I see some seemingly unrelated changes:
image

japaric (approved):
thanks for the PR. the changes look good to me. the test failure should be fixed by #319

japaric:
bors merge

PR #322, opened 2022-10-20
clarify the iteration order of IndexMap / IndexSet

### LINKED ISSUE #321: What "order" is it for IndexSet ? doc is a bit confusing
The doc says:
`Return an iterator over the values of the set, in their order`
and
`Will print in an arbitrary order.`
from:
https://github.com/japaric/heapless/blob/df890b1a4bc512ef6e8eb41165861a46a9c35b5a/src/indexset.rs#L114-L128

But `in their order` and `arbitrary order` are somewhat confusing, especially since the books [example](https://github.com/japaric/heapless/blob/df890b1a4bc512ef6e8eb41165861a46a9c35b5a/src/indexset.rs#L62-L80) appears to return the items in insertion order. (and my test below confirms it **EDIT:** but it was a coincidence - it is arbitrary order)

**Can this order be clarified in the docs, please?** is it really `arbitrary order` or is it insertion order? 
 and is the order the same for `.iter()` as it's for `for book in &books` (this I might know if I weren't a Rust newbie, ie. maybe that `for` is secretly just calling `books.iter()` ?)

My test shows that insertion order is kept, but maybe that's just a coincidence!!?:
1. `cd /tmp`
2. `cargo new --lib wtw`
3. `cd wtw`
4. `echo 'heapless = "0.7.9"' >> Cargo.toml`
5. add the following test into `src/lib.rs`:
```rust
#[test]
    fn kept_insertion_order_ints() {
        use heapless::FnvIndexSet;
        const LAST: usize = 128; //must be power of two, FnvIndexSet requirement
        type Numbers = u32;
        let mut nums = FnvIndexSet::<Numbers, LAST>::new();
        const FIRST:Numbers=1;
        const THROWER:Numbers=1000;
        const PICKER:Numbers=100;
        for i in FIRST..=LAST as Numbers {
            let r: Result<bool, Numbers>;
            r=if i==PICKER {
                nums.insert(THROWER)
            } else {
                nums.insert(i)
            };
            //println!("{r:?}");
            assert_eq!(r,Ok(true));
        }
        let mut i:Numbers = FIRST;
        let mut iter=nums.iter();
        for j in &nums {

            if let Some(cur)=iter.next() {
                println!("{:?}", cur);
                if i==PICKER {
                    assert_eq!(*cur,THROWER);
                    assert_eq!(*j, THROWER);
                } else {
                    assert_eq!(*cur,i);
                    assert_eq!(*j, i);
                }
            }

            i += 1;
        }
    }
```

6. `cargo test -- --show-output`

```
1
2
3
4
...
96
97
98
99
1000
101
102
103
104
105
...
125
126
127
128
```

japaric:
closes #321

### DIFF
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -510,7 +510,7 @@ where
         N
     }
 
-    /// Return an iterator over the keys of the map, in their order
+    /// Return an iterator over the keys of the map, in insertion order
     ///
     /// ```
     /// use heapless::FnvIndexMap;
@@ -528,7 +528,7 @@ where
         self.core.entries.iter().map(|bucket| &bucket.key)
     }
 
-    /// Return an iterator over the values of the map, in their order
+    /// Return an iterator over the values of the map, in insertion order
     ///
     /// ```
     /// use heapless::FnvIndexMap;
@@ -546,7 +546,7 @@ where
         self.core.entries.iter().map(|bucket| &bucket.value)
     }
 
-    /// Return an iterator over mutable references to the the values of the map, in their order
+    /// Return an iterator over mutable references to the the values of the map, in insertion order
     ///
     /// ```
     /// use heapless::FnvIndexMap;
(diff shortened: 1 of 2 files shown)


japaric:
bors merge

PR #323, opened 2022-10-20
spsc::Queue: make example less unsafe

### LINKED ISSUE #314: Example usage of `heapless::spsc` triggers UB?
The `split` [example](https://docs.rs/heapless/latest/heapless/spsc/index.html#examples) of `heapless::spsc` attempts to avoid the need for using a `Mutex<RefCell<Queue>>` _as well as avoiding `static` scope for `Producer` and `Consumer` by using `unsafe` code:

```
fn main() {
    // NOTE(unsafe) beware of aliasing the `consumer` end point
    let mut consumer = unsafe { Q.split().1 };
    ...
```

```
fn interrupt_handler() {
    // NOTE(unsafe) beware of aliasing the `producer` end point
    let mut producer = unsafe { Q.split().0 };
    ...
```

I'm not 100% sure, but I _believe_ this usage is in and of itself UB; the lifetime of `consumer` and `producer` is tied to `Q`, and the _mutable_ borrow of `Q` lasts as long as `consumer` or `producer` lives. If an interrupt happens while `consumer` exists, `Q` is mutably borrowed twice when `producer` is created.

I would appreciate clarification, as I'm not 100% sure if this is UB (and if it is, what an alternate `unsafe` solution should look like). Here is a playground snippet of me attempting to minimize the `spsc` code to show UB in Miri: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=597cb05f9f1003515b01e82f79abb9c1

japaric:
closes #314

### DIFF
--- a/src/spsc.rs
+++ b/src/spsc.rs
@@ -22,23 +22,30 @@
 //! assert_eq!(rb.dequeue(), Some(0));
 //! ```
 //!
-//! - `Queue` can be `split` and then be used in Single Producer Single Consumer mode
+//! - `Queue` can be `split` and then be used in Single Producer Single Consumer mode.
+//!
+//! "no alloc" applications can create a `&'static mut` reference to a `Queue` -- using a static
+//! variable -- and then `split` it: this consumes the static reference. The resulting `Consumer`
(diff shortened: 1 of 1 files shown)


japaric:
bors merge

PR #324, opened 2022-10-24
Add an async implementation of `Queue`

datdenkikniet:
This PR adds an async version/wrapper around `spsc::Queue` with fallible interfaces that can be used in sync contexts.

Currently it uses a (non-public) single slot queue to store wakers.


src/async_impl/spsc/producer.rs
```diff
@@ -0,0 +1,136 @@
...
+        let failed_enqueue_value = if let Some(value) = prod.inner.enqueue(value).err() {
+            value
+        } else {
+            // Try to wake the consumer because we've enqueued our value
+            return try_wake_consumer(me);
+        };
+
+        me.value_to_enqueue = Some(failed_enqueue_value);
+
+        if !me.producer.register_waker(cx.waker().clone()) {
```
Dirbaio, line 129: This should replace the previous waker instead of spinning the CPU at 100%.

Also, cloning wakers is moderately expensive in many executors (it does atomic refcount ops. Not in embassy or RTIC though), so it's nice to avoid cloning if possible. The way you do it is:

- If no waker registered, clone and register it.
- If a waker is registered:
  - If `old_waker.will_wake(new_waker)` do nothing. This saves a clone.
  - Otherwise, drop the old one, clone the new one and register it.

korken89:
Hi, I think `heapless` should have an `async` module in where we can look at adding `async` specific data structures. Maybe @Dirbaio also has some from `embassy` that would make sense to put here?

PR #326, opened 2022-11-30
Recover StableDeref trait for pool::object::Object and pool::boxed::Box

qwerty19106:
I add ReadBuffer and WriteBuffer to pool::object::Object and pool::box::Box.
It allows to use embedded DMA of HAL crates, such as [stm32f4xx_hal::dma::Transfer](https://docs.rs/stm32f4xx-hal/latest/stm32f4xx_hal/dma/struct.Transfer.html).

### DIFF
--- a/src/pool/boxed.rs
+++ b/src/pool/boxed.rs
@@ -84,6 +84,8 @@ use core::{
     ops, ptr,
 };
 
+use stable_deref_trait::StableDeref;
+
 use super::treiber::{NonNullPtr, Stack, UnionNode};
 
 /// Creates a new `BoxPool` singleton with the given `$name` that manages the specified `$data_type`
@@ -211,6 +213,8 @@ where
         unsafe { &mut *self.node_ptr.as_ptr().cast::<P::Data>() }
     }
 }
+
+unsafe impl<P> StableDeref for Box<P> where P: BoxPool {}
 
 impl<A> fmt::Display for Box<A>
 where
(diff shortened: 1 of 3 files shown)


burrbull:
Nice.

Although it is strange for me `heapless` depends on `ufmt`, `defmt` and `embedded-dma`. I'd better made conversely.

qwerty19106:
The ```embedded_dma``` contains 5 traits only and provides common implementation for system types: u8, i8, [u8], et.al.
The ```stm32f4hal``` crate contains implementation with <BUF: ReadBuffer>.
In my opinion it is logical that ```heapless``` contains trait implementation next to the struct.

qwerty19106:
I thought about it for a bit and realized that to use `embedded_dma::ReadBuffer` enough implement `StableDeref`.
The `StableDeref` trait was previously implemented for `pool::Box` (before adding object pool API).

qwerty19106:
@burrbull Could you take another look?

burrbull (approved):

qwerty19106:
First `testtsan` was cancelled.
Next `testtsan` pass ok locally. It is noise?
In any case, the errors do not belong to the `pool::object` or `pool::boxed`.

korken89:
Hi please rebase to fix tsan issues

qwerty19106:
Rebased

src/pool/boxed.rs
```diff
@@ -301,6 +305,32 @@ where
 {
 }
 
+#[cfg(feature = "embedded-dma-impl")]
```
korken89, line 308: Should these be removed now when using `StableDeref`?

qwerty19106: Yes, I removed it.

korken89 (approved):
bors merge

PR #327, opened 2022-12-01
Add HistoryBuffer::as_slices()

hulthe:
This method returns the contents of a HistoryBuffer as two ordered slices, similar to [Deque::as_slices()](https://docs.rs/heapless/0.7.16/heapless/struct.Deque.html#method.as_slices).


korken89:
Hi, can you change so it uses a tuple instead?

hulthe:
Sure can!

I personally think an array is more helpful though (e.g. like how i used into_iter in the test).
Though I could see an argument for keeping it more in line with similar functions, like Deque::as_slices.

hulthe:
@korken89 here you go

korken89 (approved):
bors merge

korken89 (approved):
bors merge

korken89 (approved):
bors merge

PR #328, opened 2022-12-23
Use `portable-atomic` instead of `atomic-polyfill`.

Dirbaio:
I intend to deprecate `atomic-polyfill` soon, in favor of `portable-atomic`. It supports more archs and is more actively maintained. ~The only thing it can't do yet is use `critical-section`, but it should soon~ it already can: https://github.com/taiki-e/portable-atomic/pull/51

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,7 @@ jobs:
           - nightly
         features:
           - ""
-          - "cas"
+          - "cas,portable-atomic/critical-section"
           - "serde"
     steps:
       - name: Checkout
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [breaking-change] export `IndexMapKeys`, `IndexMapValues` and
   `IndexMapValuesMut` iterator types.
 
-- [breaking-change] this crate now depends on `atomic-polyfill` v1.0.1, meaning that targets that
-  require a polyfill need a `critical-section` **v1.x.x** implementation.
+- [breaking-change] this crate now uses `portable-atomic` v1.0 instead of `atomic-polyfill` for emulating
+  CAS instructions on targets where they're not natively available.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ version = "0.8.0"
 
 [features]
 default = ["cas"]
-cas = ["atomic-polyfill"]
+cas = ["portable-atomic"]
 ufmt-impl = ["ufmt-write"]
 # only for tests
 __trybuild = []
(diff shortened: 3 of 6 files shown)


korken89:
Hi please rebase to fix tsan failure

korken89:
Hi, I gave this a check and saw something weird.

```
RUSTFLAGS="--cfg=portable_atomic_unsafe_assume_single_core"
```

Is there a reason this is not a feature to `heapless`?

Dirbaio:
it's a feature for `atomic-polyfill`. See its [readme](https://github.com/taiki-e/portable-atomic/#optional-cfg). The reasoning for making it a `--cfg` and not a Cargo feature is that enabling it on non-multicore chips can cause unsoundness, so the author wants to enforce only the final binary crate can enable it, not library crates. (it's a decision I don't agree with, see some discussion [here](https://github.com/tokio-rs/bytes/pull/573#issuecomment-1332723396)... :shrug: )

korken89:
Hmm, I see - that was one long discussion. :sweat_smile: 

Not sure what I think of it yet, I'll leave this open until I can come to terms with `--cfg` or if @japaric forces it through.

Dirbaio:
@taiki-e: There's many cases where a crate knows it's going to run in single-core (for example `embassy-stm32` with `stm32f4xx` feature. All F4 chips are single-core). In these cases it'd want to enable the `atomic-polyfill/unsafe-assume-single-core` Cargo feature if it existed, so that it all Just Works for the end user.

Asking the user to add nonstandard `--cfg` args is very annoying. Using the `critical-section` feature to avoid it is suboptimal, it'll result in worse performance unnecessarily (it has to acquire the CS for load/store, while `unsafe-assume-single-core` only has to for CAS ops).

Could you please consider adding a `unsafe-assume-single-core` Cargo feature?

taiki-e:
> `unsafe-assume-single-core` Cargo feature

After discussion in https://github.com/tokio-rs/bytes/pull/573, I'm starting to think that adding per-target/implementation `unsafe-assume-single-core-*` features (e.g., `unsafe-assume-single-core-cortex-m`) might be fine.

The cross-platform `unsafe-assume-single-core` feature can be roughly enabled by the user, and also cargo v1 resolver may enable it for unintended targets, so I'm not positive to add it.

However, I think per-target/implementation `unsafe-assume-single-core-*` features that cause compile error on unsupported targets might be enough (though not perfect[^1]) to prevent such an accidental misuse (cross-platform `unsafe-assume-single-core` feature silently causes problems). (I have previously seen a GUI-related crate that causes a compile error and suggests the use of cargo v2 resolver, when the feature is accidentally enabled on unsupported targets.)

[^1]: It does not address all of my original concerns, but I think it can address the most annoying part " even if the library maintainer understands the target correctly and takes the correct action, it may not work correctly for the end user".

Dirbaio:
ping @korken89 

`atomic-polyfill` is now [deprecated](https://crates.io/crates/atomic-polyfill), and `portable-atomic` 1.4 added the `unsafe-assume-single-core` Cargo feature so now there's no case where you need `--cfg`s in embedded usage.

newAM (approved):

PR #331, opened 2023-01-05
Relax trait requirements on IndexMap/IndexSet.

jeandudey:
Some requirements are not necessary and the specification can be removed/simplified. This makes these types a bit more ergonomic when using them in a trait or function definition as not all traits need to be specified for some cases.


korken89:
Hi please rebase to fix tsan failure

korken89 (approved):
Thanks!

bors merge

PR #332, opened 2023-01-05
Export Iter from IndexSet and IndexMap

jeandudey:
Exports the `Iter` types for `IndexSet` and `IndexMap`.

My use case is having a trait generic over a `BTreeSet` and `IndexSet` which has a `type Iter`.

PR #338, opened 2023-01-13
export all IndexMap iterators

jeandudey:
Exports all the iterator types of `IndexMap`.

I wonder if it'd make sense to have a module structure such as the one in `alloc::collections`, e.g.: `heapless::indexmap` and re-export `heapless::indexmap::IndexMap` on the crate root, to have all the iterator types inside the `indexmap` module but still public.


jeandudey:
Didn't include a change in the changelog as it already includes a line for that from the previous PR #332

korken89:
Looks good to me!

Could you update the CHANGELOG to indicate that these are breaking changes?

jeandudey:
Updated the CHANGELOG

korken89:
Looks good! 
As a final thing could you extend the tests to test the new features added?

korken89 (approved):
bors merge

PR #347, opened 2023-02-25
Add `recent_index`, `oldest_index`, `oldest,` and `is_filled` to HistoryBuffer API

YuhanLiin:
I added some convenience methods to the HistoryBuffer API to help keep track of both ends of the buffer.


pleepleus:
I was just about to fork this repo because I needed these.  I will be happy to review.

pleepleus (changes requested):
The code changes themselves look excellent.  In fact I already pointed my project to this commit because I needed the "oldest()" api and it passes my application tests.

Before this can be merged though it should:

1.  Refresh with a merge from main
2.  Modify the changelog with a note on the additions

newAM (changes requested):
Sorry for the slow review!  Life has been keeping me busy lately.

newAM:
Thanks for making the change!

I took a closer look before merging, and I have one last question (sorry!).  What is the use-case of the `_index` methods being public?  That seems like an implementation detail.

YuhanLiin:
The index methods can be used with `as_slice` (which also seems like an implementation detail) to do indexing operations relative to the head and tail indices.

newAM (approved):
Thanks for adding this!  Sorry it took so long to get it merged.

PR #350, opened 2023-02-27
Fix pool (mis)compile on AArch64

### LINKED ISSUE #349: LLSC probe incorrectly includes `pool` module on AArch64
Attempting to execute test suite on M1 MacBook fails with:

```
heapless % cargo t
   Compiling heapless v0.8.0 (~/Code/heapless)
error: unrecognized instruction mnemonic, did you mean: clrex, ldr?
   --> src/pool/treiber/llsc.rs:129:15
    |
129 |         asm!("ldrex {}, [{}]", out(reg) value, in(reg) addr, options(nostack));
    |               ^
    |
note: instantiated into assembly here
   --> <inline asm>:1:2
    |
1   |     ldrex x8, [x9]
    |     ^

error: unrecognized instruction mnemonic, did you mean: str?
   --> src/pool/treiber/llsc.rs:138:15
    |
138 |         asm!("strex {}, {}, [{}]", out(reg) outcome, in(reg) value, in(reg) addr, options(nostack));
    |               ^
    |
note: instantiated into assembly here
   --> <inline asm>:1:2
    |
1   |     strex x8, x9, [x10]
    |     ^

error: could not compile `heapless` due to 2 previous errors
```

Looks like the LLSC probe in `build.rs` is wrongly assuming presence of `ldrex` and `strex` in instruction set if `clrex` is known mnemonic. [The supported A64 instructions](https://developer.arm.com/documentation/100069/0609/A64-General-Instructions/A64-instructions-in-alphabetical-order?lang=en) include `clrex` but not `ldrex` or `strex`.

We can avoid this by refusing to run the probe on AArch64 or updating the probe to check for all instructions.

rjsberry:
Fixes #349

Alternatively we could change the probe to something like:

```rust
#[no_mangle]
pub unsafe fn probe(x: &usize, y: usize) -> usize {
    let mut ret;
    asm!(
        "       clrex",
        "2:",
        "       ldrex {2}, [{0}]",
        "       strex {tmp}, {1}, [{0}]",
        "       cmp {tmp}, #0",
        "       bne 2b",
        in(reg) x,
        in(reg) y,
        out(reg) ret,
        tmp = out(reg) _,
    );
    ret
}
```

Without explicitly denying any `aarch64*` target.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -88,9 +88,13 @@ fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 
-    match compile_probe(ARM_LLSC_PROBE) {
-        Some(status) if status.success() => println!("cargo:rustc-cfg=arm_llsc"),
-        _ => {}
+    // AArch64 instruction set contains `clrex` but not `ldrex` or `strex`; the
+    // probe will succeed when we already know to deny this target from LLSC.
+    if !target.starts_with("aarch64") {
+        match compile_probe(ARM_LLSC_PROBE) {
+            Some(status) if status.success() => println!("cargo:rustc-cfg=arm_llsc"),
+            _ => {}
+        }
     }
 
     Ok(())


AdinAck:
Please merge this. Tests cannot be run on Darwin processors without this PR.

newAM (dismissed):
Thanks for the fix!  Will merge soon after CI is fixed.

newAM (approved):

Issue #351, opened 2023-03-01
Const assertions are not guaranteed to cause compile failures


rjsberry:
In [`sealed.rs`](https://github.com/japaric/heapless/blob/644653bf3b831c6bb4963be2de24804acf5e5001/src/sealed.rs), there are a number of associated constants which are used in const fns to try and trigger compile failures when const generics don't match particular requirements.

These const fns contain constant _expressions_, which may (but are not guaranteed to) evaluate at compile time. AFAICT from https://doc.rust-lang.org/reference/const_eval.html while this works now it might stop working in the future and instead cause a runtime panic. 

Running `cargo check` on an example which we know won't build:

```
% cat examples/bad_indexmap.rs  
fn main() {
    // needs N to be a power of 2
    let _map = heapless::FnvIndexMap::<u8, u8, 3>::new();
}

% cargo check --example bad_indexmap
    Checking heapless v0.8.0 (~/Code/heapless)
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s

% cargo build --example bad_indexmap
   Compiling heapless v0.8.0 (~/Code/heapless)
error[E0080]: evaluation of `heapless::sealed::Assert::<3, 0>::POWER_OF_TWO` failed
  --> ~/Code/heapless/src/sealed.rs:57:37
   |
57 |     pub const POWER_OF_TWO: usize = 0 - (L & (L - 1));
   |                                     ^^^^^^^^^^^^^^^^^ attempt to compute `0_usize - 2_usize`, which would overflow

note: the above error was encountered while instantiating `fn heapless::sealed::power_of_two::<3>`
   --> ~/Code/heapless/src/indexmap.rs:541:9
    |
541 |         crate::sealed::power_of_two::<N>();
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0080`.
error: could not compile `heapless` due to previous error
```

We know this is a bit of a hack evidenced by the comments: `// Const assert hack`. It would be good to try and add coverage for these compile fails in the `cfail` test suite so that we can detect if/when this stops being a compile-time error. Unfortunately the `trybuild` crate is using `cargo check` under the hood and would not be able to hit this.

burrbull:
This is not actually related to the issue.
This line can be rewritten like:
```
pub const POWER_OF_TWO: () = assert!(L.is_power_of_two());
```

newAM:
[...]

Issue #344, opened 2023-02-15
Feature request create a vec macro


seanybaggins:
I was wondering if you could make a vec macro that is similar to the one in the standard lib.

Issue #341, opened 2023-01-30
Infallible Vec operations


robin-nitrokey:
Some operations on a `Vec<T, N>` are infallible but can only be implemented using fallible methods.  What do you think about adding more infallible methods?  For example:

- `from_array(data: [T; N])`: construct a `Vec` from an array with the same capacity (would also be useful for constants)
- `fill(&mut self, value: T)`:  resize to capacity and clone `value` to empty slots
- `fill_default(&mut self)`: resize to capacity and write `T::default()` to empty slots
- `replace(&mut self, other: &Self)`: clear this `Vec` and copy the data from `other`

reitermarkus:
`from_array` is now implemented in https://github.com/rust-embedded/heapless/pull/352.

`fill` already exists on `slice`, i.e. via `DerefMut`, same with `fill_with(Default::default)` instead of `fill_default`. These only fill up to the current length, though. What is the use case for the fill methods filling up to capacity? Is it covered now with `.spare_capacity_mut().fill(…)`?

For `replace`, I assume this should work only for `Vec` with the same capacity, otherwise it would still need to be fallible. Again, what is the use case for this method?

robin-nitrokey:
Good to see `from_array` being added!  Unfortunately, the implementation is not `const`.  It would still be useful to be able to create non-empty constants.

My use case for `fill` is preparing a buffer for functions that expect a `&mut [u8]` for writing.  Currently, you have to call `v.resize_default(v.capacity())` which is fallible.  `spare_capacity_mut().fill(MaybeUninit::new(...))` should work but is not really intuitive.

For `replace`, I don’t remember the specific use case.  I’ll report back if I find it again.

reitermarkus:
@robin-nitrokey, can you give a full example on how `fill` would be used?

Maybe `v.extend(core::iter::repeat(...))` works?

robin-nitrokey:
I don’t think `extend` + `repeat` would work as [`extend`](https://docs.rs/heapless/latest/heapless/struct.Vec.html#method.extend) “[p]anics if the vec cannot hold all elements of the iterator.”

Here is a real world example:

https://github.com/trussed-dev/ctap-types/blob/084db87a800e1204bcfdc3e25ca46bc4a81650ba/src/ctap2.rs#L148-L175

[...]

PR #352, opened 2023-03-02
Impl conversion from array to Vec

rjsberry:
This PR implements conversion from arrays to vec. The array does not have to be the same length as the capacity of the vec. Const generics are asserted at compile time. Unfortunately the method currently requires a copy of the source array.

Partially addresses requests in #341 and #344.

Also I'm not totally happy about the const assertions given #351. I've not added any tests in the `cfail` suite as I don't think we can practically check this with `trybuild` at the moment. There is a bit of a work-around you can do but it's not pretty.

## Questions

1. Would we prefer to just implement `From<[T; N]> for Vec<T, N>`? One of the issues with this impl is that users will need to be explicit about vec capacity. Because lengths `N` and `M` can be different, you can't write:

   ```rust
   let v = Vec::from([1, 2, 3]);
   ```

   It needs to be:

   ```rust
   let v = Vec::<_, 3>::from([1, 2, 3]);
   ```

2. Should `from_array` be internal, and the API exposed by the crate simply be the `From` trait?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `format` macro.
 - Added `String::from_utf16`.
 - Added `is_full`, `recent_index`, `oldest`, and `oldest_index` to `HistoryBuffer`
+- Added infallible conversions from arrays to `Vec`.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


Dirbaio (dismissed):
thank you!

Dirbaio (approved):

PR #364, opened 2023-03-29
Add ordered_get() to HistoryBuffer

schultetwin1:
This allows users to get access to the elements of the history buffer in historical order.

### DIFF
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -527,16 +527,7 @@ impl<T, S: HistBufStorage<T> + ?Sized> HistoryBufferInner<T, S> {
     /// assert_eq!(buffer.ordered_get(6), None);
     /// ```
     pub fn ordered_get(&self, idx: usize) -> Option<&T> {
-        if self.len() <= idx {
-            None
-        } else {
-            let (front, back) = self.as_slices();
-            if idx < front.len() {
-                Some(&front[idx])
-            } else {
-                Some(&back[idx - front.len()])
-            }
-        }
+        self.oldest_ordered().nth(idx)
     }
 
     /// Returns double ended iterator for iterating over the buffer from
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -504,14 +504,14 @@ impl<T, S: HistBufStorage<T> + ?Sized> HistoryBufferInner<T, S> {
         }
     }
 
-    /// Returns a reference to the element in the order from oldest to newest
+    /// Returns a reference to the element in the order from oldest to newest.
     ///
-    /// `buf.ordered_get(0)` will always return the oldest element in the buffer
+    /// `buf.ordered_get(0)` will always return the oldest element in the buffer.
     ///
     /// `buf.ordered_get(buf.len() - 1)` will always return the newest element
-    /// in the buffer
+    /// in the buffer.
     ///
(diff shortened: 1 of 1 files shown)


src/histbuf.rs
```diff
@@ -204,6 +204,41 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
...
     }
 
+    /// Returns a reference to the element in the order from oldest to newest
+    ///
+    /// `buf.ordered_get(0)` will always return the oldest element in the buffer
+    ///
+    /// `buf.ordered_get(buf.len() - 1)` will always return the newest element
+    /// in the buffer
+    ///
+    /// Returns None if `index >= self.len()`
```
reitermarkus, line 214:
```suggestion
    /// Returns None if `index >= self.len()`.
```

src/histbuf.rs
```diff
@@ -204,6 +204,41 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
         }
     }
 
+    /// Returns a reference to the element in the order from oldest to newest
+    ///
+    /// `buf.ordered_get(0)` will always return the oldest element in the buffer
```
reitermarkus, line 209:
```suggestion
    /// `buf.ordered_get(0)` will always return the oldest element in the buffer.
```

src/histbuf.rs
```diff
@@ -204,6 +204,41 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
...
+    ///
+    /// `buf.ordered_get(buf.len() - 1)` will always return the newest element
+    /// in the buffer
+    ///
+    /// Returns None if `index >= self.len()`
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::HistoryBuffer;
```
reitermarkus, line 219:
```suggestion
    /// use heapless::HistoryBuf;
```

src/histbuf.rs
```diff
@@ -204,6 +204,41 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
         }
     }
 
+    /// Returns a reference to the element in the order from oldest to newest
+    ///
+    /// `buf.ordered_get(0)` will always return the oldest element in the buffer
+    ///
+    /// `buf.ordered_get(buf.len() - 1)` will always return the newest element
+    /// in the buffer
```
reitermarkus, line 212:
```suggestion
    /// in the buffer.
```

src/histbuf.rs
```diff
@@ -495,6 +508,24 @@ mod tests {
         }
     }
 
+    #[test]
+    fn ordered_get() {
+        let mut buffer: HistoryBuffer<u8, 6> = HistoryBuffer::new();
```
reitermarkus, line 513:
```suggestion
        let mut buffer: HistoryBuf<u8, 6> = HistoryBuf::new();
```

reitermarkus:
Okay, after refactoring this to `self.oldest_ordered().nth(idx)`, I am not convinced we should have a separate method for this. Most often you will need `recent()` or `oldest()`, for the other times I think `oldest_ordered().nth(n)` is actually more readable.

schultetwin1:
I did not realize I could just use `self.oldest_ordered().nth(idx)`, I agree that's fine. Thank you for all the time you spent looking at this!

PR #365, opened 2023-04-11
Add method drain for Vec

### LINKED ISSUE #334: Add method drain for Vec, like std::vec::Vec?
std drain retures a new struct Drain, but heapless can drop the drained data by using method copy_within & truncate.

remme123:
Closes #334

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -822,6 +822,35 @@ impl<T, const N: usize> Vec<T, N> {
         // All item are processed. This can be optimized to `set_len` by LLVM.
         drop(g);
     }
+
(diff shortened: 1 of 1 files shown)


Dirbaio:
We should have the same API as std Vec's `drain`.

- Take a range, so you can drain from the middle, not just from the start.
- Return an iterator that gets the drained elements.

Dirbaio (changes requested):
We should have the same API as std Vec's drain.
-    Take a range, so you can drain from the middle, not just from the start.
-    Return an iterator that gets the drained elements.

PR #368, opened 2023-05-25
Update spin 0.9.8

jentfoo:
Update spin to 0.9.8.  Although not directly exploitable this will help with dependent libraries which are flagging this security issue: https://github.com/advisories/GHSA-2qv5-7mw5-j3cg


Dirbaio:
turns out `spin` is completely unused, it has been removed in #389

PR #369, opened 2023-05-30
add format macro implementation

ede1998:
Motivation:
If you want to create a formatted `heapless::String` at the moment, it is necessary to create a mutable string and modify it:

```rust
let mut string = heapless::String::<10>::new();
write!(&mut string, "{:>3}cm", height).unwrap();
```

It's a bit nicer to have a `format!` macro that constructs the `String` in-place without requiring an intermediate step, just as Rust's standard library offers.

```rust
let string = format!(10, "{:>3}cm", height);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Added `format` macro.
+
 ### Changed
 
 - Changed `stable_deref_trait` to a platform-dependent dependency.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,7 +97,7 @@ mod histbuf;
 mod indexmap;
 mod indexset;
 mod linear_map;
-pub mod string;
+mod string;
 mod vec;
 
 #[cfg(feature = "serde")]
@@ -134,3 +134,10 @@ pub mod spsc;
 mod ufmt;
 
 mod sealed;
+
+/// Implementation details for macros.
+/// Do not use. Used for macros only. Not covered by semver guarantees.
+#[doc(hidden)]
+pub mod _export {
+    pub use crate::string::format;
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,7 +97,7 @@ mod histbuf;
 mod indexmap;
 mod indexset;
 mod linear_map;
-mod string;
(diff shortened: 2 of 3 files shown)


Dirbaio (changes requested):
Thanks for the PR!

Could you change it to return a `Result` instead of panicking if it runs out of space? We should avoid implicit panics as much as possible. It's true std's `format!` panics on fmt errors, but these are super rare. Running out of space is more likely.

Also I wonder if we could let rust infer the string length instead of having to specify it manually. 

Thanks!

ede1998:
Sure thing.

Regarding inferring the capacity: This should be possible. However I'm not sure this is a good idea. If you want to bind the string to a new variable, you now have to specify the entire type just to be allowed to add the capacity, which adds quite a bit of noise:
```rust
let foo: helpless::String<10> = format!(...);
```
It pretty much comes down to what is the more common usage pattern.
If you use the macro to initialize a field of an existing struct (-> type and capacity already fixed), it makes it nicer as you don't have to repeat the capacity in the macro.

Dirbaio:
yeah that's true. Supporting both syntaxes sounds nice! like `format!("foo {}", foo)` or `format!(32, "foo {}", foo)` sounds good to me!

birkenfeld:
> ```rust
> helpless::String
> ```

That's another crate still ;)

ede1998:
Seems like my phone's autocorrect was in a funny mood.

I updated the PR. The macro now returns a `Result<String<N>, core::fmt::Error>`.

The inferred capacity unfortunately did not work as planned. I had to use a semicolon as separator instead.

```rust
struct Inferred {
  name: String<10>,
}

let inf = Inferred {
  name: format!("Hello {}", "World")?,
};

// Notice semicolon here: otherwise Rust would pick up 10 as the format string and fail.
let explicit = format!(10; "Foo {}", "bar")?;
```

Hope this is ok too.

Dirbaio (dismissed):
looking great! Thanks for the PR

Dirbaio:
did a small fix, instead of making `mod string` public which affects the public API, it's nicer to make a "hidden" `mod _export` that's explicitly unstable for these.

Dirbaio (approved):

ede1998:
Oh, yeah that's much nicer. I figured there was only `String` in the module so it wouldn't be a problem but you're way is better! Thanks for merging!

PR #376, opened 2023-08-31
Fix Linear Map for Cargo 1.71

AdinAck:
## Motivation

In the latest version of Cargo the implementation of `Drop` for linear map does not compile.

## Changes

Revert `Drop` impl to previous implementation.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,65 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "lldb",
+            "request": "launch",
+            "name": "Debug unit tests in library 'heapless'",
+            "cargo": {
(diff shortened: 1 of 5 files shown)


birkenfeld:
This patch isn't correct. Like the new code says, `Vec` already implements `Drop`, so your change will call `drop()` twice on the contents. Look at this example:

```rust
struct S;

impl Drop for S {
    fn drop(&mut self) {
        println!("dropped an S");
    }
}

fn main() {
    let mut v = heapless::LinearMap::<&str, S, 5>::new();
    v.insert("a", S).ok().unwrap();
}
```

It will print "dropped an S" twice with the PR.

Just removing the `drop(&self.buffer)` is enough.

AdinAck:
I did notice it continued to work when the entire `Drop` impl was removed, didn't realize it *actually* is not necessary. I will just remove it then, thanks.

newAM:
There seems to be something wrong with the commits here, perhaps a `git revert` gone wrong?

I made a separate PR to fix in #388

AdinAck:
cool yeah this PR desperately needed a rebase

Dirbaio:
fixed in #388

PR #377, opened 2023-08-31
HistoryBuffer Add pop_oldest

Ryder-C:
## Motivation
There currently is no interface for removing elements from the HistoryBuffer.

## Added
- `HistoryBuffer.pop_oldest()`
- `HistoryBuffer.filled()` getter to replace `filled` member variable
- Unit test for `HistoryBuffer.pop_oldest()`

## Changed
- `HistoryBuffer.len()` to be a getter rather than computed on call.
- `HistoryBuffer.write()`
- `HistoryBuffer.recent()`
- `HistoryBuffer.oldest_ordered()`
- `OldestOrdered.next()`

## Warning
`HistoryBuffer.as_slice()` no longer has the same meaning. The slice returned may contain unused values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implemented `retain` for `IndexMap` and `IndexSet`.
 - Recover `StableDeref` trait for `pool::object::Object` and `pool::boxed::Box`.
 - Add polyfills for ESP32S2
-- - `HistoryBuffer.pop_oldest()`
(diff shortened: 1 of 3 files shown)


Dirbaio:
what's the use case for this? It seems to me there's a lot of overlap with Deque here. IMO if you need push/pop you should use Deque, and HistoryBuffer should be used only when you want to do "statistics" on the last N values, not as a queue you pop things from.

reitermarkus:
Closing due to no response. Feel free to re-open if there is a good use case where `Deque` isn't sufficient.

PR #385, opened 2023-10-20
linear_map: allow dropping references

newAM:
Testing to see if there are other CI issues that need fixing.


newAM:
`testcfail` addressed in #387

Dirbaio:
fixed in #388

PR #388, opened 2023-10-20
Fix dropping_references warning

newAM:
Please do review this carefully.  I'm not confident this (the fix and the test) is correct.  Note there is `unsafe` in `impl Drop for Vec` which is the backend data structure for `LinearMap`.

A couple CI failures will still occur, see #387.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on:
     branches: [main]
   push:
     branches: [staging, trying]
+  workflow_dispatch:
 
 env:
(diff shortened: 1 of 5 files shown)


Dirbaio (dismissed):
LGTM. I'm not sure what was up with that drop code lol, the underlying Vec will already drop the contents fine.

I've merged #387 into this one because there's no way to merge them separately if CI fails.

Dirbaio:
omg there's also a tsan test failing

Dirbaio (approved):

PR #390, opened 2023-10-20
Update GHA

newAM:
No visible change to users, not worth updating the changelog IMO.


Dirbaio (approved):

Dirbaio:
thanks! maybe we should move off the `actions-rs` actions, the whole org was recently archived :cold_sweat:

newAM:
Yeah, I'm generally happy with https://github.com/dtolnay/rust-toolchain for my other repos, that's the next step here.

PR #395, opened 2023-10-21
string: replace From<&str> with TryFrom<&str>

### LINKED ISSUE #381: Implement `TryFrom<&str>` for fallible conversion
It's currently a bit cumbersome to go from a regular Rust string to a `heapless::String` and properly catch if it does not fit. The current `From<&str>` impl panics if the string is too long. It would be nice with a corresponding `TryFrom` impl.

The current best way is to create an empty heapless string and use `push_str` and catch the error there. But it's not very ergonomic.

newAM:
Fixes #381.

This is a breaking change.

The `From` trait is not allowed to fail: [docs].  The current implementation breaks this contract.

As-is `TryFrom` cannot be implemented to fix the usability issue raised in #381 because `From` provides `TryFrom` with `type Error = Infallible`.

[docs]: https://doc.rust-lang.org/std/convert/trait.From.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - [breaking-change] this crate now depends on `atomic-polyfill` v1.0.1, meaning that targets that
   require a polyfill need a `critical-section` **v1.x.x** implementation.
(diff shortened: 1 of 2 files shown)


Dirbaio:
needs changelog rebase

PR #399, opened 2023-10-30
String: add from_utf8

### LINKED ISSUE #378: String::from_utf8?
Im using nom to parse binary data, coming from a modem. The data is sometimes uft8 and sometimes other binary data.  At some point I want to convert from u8 slices to a heapless String...

```
struct Request
{
    method: String<5>,
    url: String<120>,
    version: String<8>
}

fn request_line(i: &[u8]) -> IResult<&[u8], Request, VerboseError<&[u8]>> {
// ....
}

```

Currently I use `core::str::from_utf8(method).unwrap().into()` to convert from the u8 slice into the `String<...>`.
Is there another, simpler way? Should we add a `from_uft8` method to heapless::String, similar to how std has one?

newAM:
Closes #378

Leverages the error type from [`core::str::from_utf8`](https://doc.rust-lang.org/std/str/fn.from_utf8.html), but it takes a `heapless::Vec` as an argument similar to [`core::String::from_utf8`](https://doc.rust-lang.org/beta/alloc/string/struct.String.html#method.from_utf8) to avoid allocation errors.

I figured we may as well add the `unsafe` version too for completeness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implemented `retain` for `IndexMap` and `IndexSet`.
 - Recover `StableDeref` trait for `pool::object::Object` and `pool::boxed::Box`.
 - Add polyfills for ESP32S2
+- Added `String::from_utf8` and `String::from_utf8_unchecked`.
 
 ### Changed
 
--- a/src/string.rs
+++ b/src/string.rs
@@ -1,4 +1,10 @@
-use core::{cmp::Ordering, fmt, fmt::Write, hash, iter, ops, str};
+use core::{
+    cmp::Ordering,
+    fmt,
+    fmt::Write,
+    hash, iter, ops,
+    str::{self, Utf8Error},
+};
 
 use crate::Vec;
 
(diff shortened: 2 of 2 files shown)


Dirbaio:
if this is going to copy the bytes anyway, wouldn't it be better to take `&[u8]`? this way the source of bytes can be something other than a `Vec`.

(the `std` method takes a `Vec` because it can convert the allocatino from `Vec` to `String`, without copying)

newAM:
If it takes `bytes: &[u8]` then it can run into an allocation error when `bytes.len() > N`.

Several options:

1. Leave it as-is, for converting from `&[u8]` the user calls `Vec::from_slice` followed by `String::from_utf8`, with each of those having a single error type.
2. Add an error enum with two variants for `Utf8Error` and `bytes.len() > N`.
3. Map the `Utf8Error` onto `()` to simplify the error type, users can leverage external methods for handling UTF-8 errors.

It's sadly not very ergonomic with any option, I prefer 1, but 2/3 (or another option) would be fine by me.  What do you think?

Dirbaio:
ah true I didn't think about capacity errors. Sounds good to me then. We can always add `from_utf8_slice` in the future.

Dirbaio (approved):

PR #403, opened 2023-11-01
Fix constant assertions

### LINKED ISSUE #351: Const assertions are not guaranteed to cause compile failures

newAM:
Fixes #351
Closes #404 (this is a superset)

I still need to check that this doesn't have a runtime cost.


newAM:
I tested this out on a `thumbv7em-none-eabihf` target, with and without the new assertions, the resulting binary size was the same.  No runtime cost that I can see.

newAM:
...I misread #351, I thought something was asserting at runtime. This doesn't improve the compatibility with `cargo check`, going to close for now.

PR #409, opened 2023-11-12
deps: make `stable_deref_trait` an optional dependency

rmsyn:
Configures the `stable_deref_trait` to only be included for `arm_llsc` and `target_arch = "x86"` platforms. This is because the dependency is only used for the `pool` module, and the `pool` module is only used on the above platforms.

Reduces the dependency graph for all other platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- configure `stable_deref_trait` as a platform-dependent dependency
+
 ## [v0.8.0] - 2023-11-07
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,9 +43,12 @@ mpmc_large = []
 portable-atomic = { version = "1.0", optional = true }
 hash32 = "0.3.0"
 serde = { version = "1", optional = true, default-features = false }
-stable_deref_trait = { version = "1", default-features = false }
 ufmt-write = { version = "0.1", optional = true }
 defmt = { version = ">=0.2.0,<0.4", optional = true }
+
+# for the pool module
+[target.'cfg(any(target_arch = "arm", target_arch = "x86"))'.dependencies]
+stable_deref_trait = { version = "1", default-features = false }
 
 [dev-dependencies]
 ufmt = "0.2"


newAM:
I think this can also be excluded on aarch64: https://github.com/rust-embedded/heapless/blob/9feb6647225316e9022f3ebdd63ab0c2380edeb8/build.rs#L31

rmsyn:
> I think this can also be excluded on aarch64

Should I also remove the associated constant and `compile_probe` function?

Cargo.toml
```diff
@@ -43,10 +43,13 @@ mpmc_large = []
 portable-atomic = { version = "1.0", optional = true }
 hash32 = "0.3.0"
 serde = { version = "1", optional = true, default-features = false }
-stable_deref_trait = { version = "1", default-features = false }
 ufmt-write = { version = "0.1", optional = true }
 defmt = { version = ">=0.2.0,<0.4", optional = true }
 
+# for the pool module
+[target.'cfg(any(target_arch = "arm", target_arch = "aarch64", target_arch = "x86"))'.dependencies]
```
newAM, line 50:
```suggestion
[target.'cfg(any(target_arch = "arm", target_arch = "x86"))'.dependencies]
```

Sorry, I'm on a desktop now.  I mean that I think `aarch64` can be excluded here and it should still work.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- configure `stable_deref_trait` as a platform-dependent dependency
```
newAM, line 24: This should go under the `Unreleased` header.

newAM (changes requested):

Cargo.toml
rmsyn: I tried that, and it fails the build. Are you saying if I remove the `build.rs` stuff, those builds will succeed without the `aarch64` target?

I'll try.

newAM: Which build failed?

I tried this out on a branch (https://github.com/newAM/heapless/commit/42cc06d0631f334da998aed02508e6aa9205e97a) and the CI passed (https://github.com/newAM/heapless/actions/runs/6844625067)

It shouldn't need `build.rs` changes.

rmsyn: I'll try it again, but all the `arm` targets failed.

rmsyn: Literally no idea why those builds were failing, some weird heisenbug. The changes as they are now, are literally what I had in the first commit...

Anyway, builds pass now.

newAM (approved):
Thanks!

Issue #411, opened 2023-11-15
style and clippy


jordens:
CI does not run clippy and style/lints have apparently been a bit overlooked.
Since this is also true in macros (e.g. the `box_*!()`) this directly impacts downstream crates that care about clippy lints and style.
We should

* [x] add clippy (all features) to the CI and
* [x] address the clippy and style lints

newAM:
Can you provide an example for reproducing this?  This issue should be fixed, but I would like to double check before making a release.

jordens:
@newAM Thanks a bunch for quickly adressing this.
An example is here: 

https://github.com/quartiq/stabilizer/pull/807/files#diff-287c7b365cd67662db5fce6120a366c3e467d8d383e7624323f50b6ee2532249

https://github.com/quartiq/stabilizer/actions/runs/6847918046/job/18617087392#step:4:247

Note that making it camel case to satisfy the type naming leads to the complaint about the static value not being UPPER_CASE.
This pattern is exactly what the dicstring suggests. Maybe we should make sure the docstrings runs as a doctest.

newAM:
Thanks!

> Maybe we should make sure the docstrings runs as a doctest.

As far as I know there's no (easy) way to run doctests as clippy lints (https://github.com/rust-lang/rust/issues/56232)

jordens:
Ah. That's unfortunate.

Maybe just not calling the pool `P` would also show up in the logs at least (`P` conveniently is both ALL_CAPS as well as CamelCase ;)
Similarly we should add tests for the pattern that is mentioned in the docs (with the `const` initializer).

newAM:
> Maybe just not calling the pool `P` would also show up in the logs at least (`P` conveniently is both ALL_CAPS as well as CamelCase ;)

Yeah that's an unfortunate miss in testing.

I added a test/fix for it here: #417.  I feel like there should be a better way to express this though :thinking: 

> Similarly we should add tests for the pattern that is mentioned in the docs (with the `const` initializer).

Yeah, we need coverage for that too as part of a normal test.  I'm still thinking about the best way to handle this one.

newAM:
I think for the `const` initializer that clippy lint should simply be suppressed in the example code (#418).

From the documentation for the [`declare_interior_mutable_const` lint](https://rust-lang.github.io/rust-clippy/master/index.html#/declare_interior_mutable_const) it says:

[...]

PR #413, opened 2023-11-17
clippy lint in mpmc and linear_map

jordens:
- mpmc: use cmp() as per clippy
- linear_map: use binding ergonomics, adress clippy

c.f. #411


newAM (approved):
Thanks!

PR #417, opened 2023-11-17
Add test for box_pool with CamelCase types

newAM:
Part of #411


reitermarkus:
Nevermind, I just realized only the type is exposed. I thought the macro generates both a type and constant with the same name.

So this looks like fine as is.

newAM:
That is a good point though, we should also suppress the clippy lint for `SNAKE_CASE` and let the user do what they want.

Dirbaio (approved):

PR #425, opened 2023-12-15
Vec: add a "View" that can mutate the vec and erase the const generic.

sosthene-nitrokey:
Alternative approach to #424, with even better ergonomics and features, but slightly more confusing documentation.

```rust
use heapless::{Vec, VecView};

let mut vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
// No const-generic on the view, and `&Vec` coerces to `&VecView` through unsize coercion 
let view: &VecView<u8> = &vec;
assert_eq!(view, &[1, 2, 3, 4]);

// No const-generic on the view, and `&mut Vec` coerces to `&mut VecView` through unsize coercion 
let mut_view: &mut VecView<u8> = &mut vec;
mut_view.push(5);
assert_eq!(vec, [1, 2, 3, 4, 5]);
```

# Pros

This approach is more "scalable" in the sense that using it means that other structures that use a `Vec` internally can also have a `View` type that is simply the `!Sized` version of the same structure. This would be useful for the implementations of `String`,  `LinearMap`, `BinaryHeap`, and `heapless-bytes`.

This approach also enables implicit conversion between the owned type and the view without needing to make the owned type `Deref` to the view.

# Cons

The `Vec` and `VecView` types being aliases to a private type could be confusing.

Not totally sure about the semver compatibility of migrating to type alias for something that used to be a struct. Cargo semver-checks seems to complain.

---

Closes https://github.com/rust-embedded/heapless/pull/424.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Vec::spare_capacity_mut`.
 - Added `Extend` impls for `Deque`.
 - Added `Deque::make_contiguous`.
+- Added `VecView`, the `!Sized` version of `Vec`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Vec::spare_capacity_mut`.
 - Added `Extend` impls for `Deque`.
 - Added `Deque::make_contiguous`.
-- Added `VecView`, the `!Sized` version of `Vec`
+- Added `VecView`, the `!Sized` version of `Vec`.
 
 ### Changed
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,9 @@ pub use indexmap::{
 pub use indexset::{FnvIndexSet, IndexSet, Iter as IndexSetIter};
 pub use linear_map::LinearMap;
 pub use string::String;
+#[cfg(doc)]
+#[doc(hidden)]
+pub use vec::VecInner as _;
 pub use vec::{Vec, VecView};
 
 #[macro_use]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,7 +88,7 @@ pub use indexmap::{
 pub use indexset::{FnvIndexSet, IndexSet, Iter as IndexSetIter};
 pub use linear_map::LinearMap;
 pub use string::String;
-pub use vec::Vec;
+pub use vec::{Vec, VecView};
 
 #[macro_use]
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,10 @@ pub use indexmap::{
 pub use indexset::{FnvIndexSet, IndexSet, Iter as IndexSetIter};
 pub use linear_map::LinearMap;
 pub use string::String;
+
+// Workaround https://github.com/rust-lang/rust/issues/119015. This is required so that the methods on `VecView` and `Vec` are properly documented.
+// cfg(doc) prevents `VecInner` being part of the public API.
+// doc(hidden) prevents the `pub use vec::VecInner` from being visible in the documentation.
 #[cfg(doc)]
 #[doc(hidden)]
 pub use vec::VecInner as _;
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -64,10 +64,20 @@ pub struct VecInner<B: ?Sized + VecDrop> {
 /// }
 /// assert_eq!(*vec, [7, 1, 2, 3]);
 /// ```
+///
+/// In some cases, the const-generic might be cumbersome. `Vec` can coerce into a [`VecView`][] to remove the need for the const-generic:
+///
+/// ```rust
+/// use heapless::{Vec, VecView};
+///
+/// let vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+/// let view: &VecView<_> = &vec;
+/// ```
 pub type Vec<T, const N: usize> = VecInner<[MaybeUninit<T>; N]>;
+
 /// A Vec with dynamic capacity
 ///
-/// [`Vec`]() coerces to `VecView`. `VecView` is !Sized, meaning that it can only ever be used through pointer
+/// [`Vec`]() coerces to `VecView`. `VecView` is `!Sized`, meaning that it can only ever be used through pointer
 ///
 /// Unlike [`Vec`](), `VecView` does not have an `N` const-generic parameter.
 /// This has the ergonomic advantages of making it possible to use functions without needing to know at
@@ -889,11 +899,29 @@ impl<T, const N: usize> Vec<T, N> {
     }
 
     /// Get a reference to the Vec, erasing the `N` const-generic
+    ///
+    /// This can also be used through type coerction, since `Vec<T, N>` implements `Unsize<VecView<T>>`:
+    ///
+    /// ```rust
+    /// use heapless::{Vec, VecView};
+    ///
+    /// let vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+    /// let view: &VecView<_> = &vec;
+    /// ```
     pub const fn as_view(&self) -> &VecView<T> {
         self
     }
 
     /// Get a `mut` reference to the Vec, erasing the `N` const-generic
+    ///
+    /// This can also be used through type coerction, since `Vec<T, N>` implements `Unsize<VecView<T>>`:
+    ///
+    /// ```rust
+    /// use heapless::{Vec, VecView};
+    ///
+    /// let mut vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+    /// let view: &mut VecView<_> = &mut vec;
+    /// ```
     pub fn as_mut_view(&mut self) -> &mut VecView<T> {
         self
     }
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -65,7 +65,27 @@ pub struct VecInner<B: ?Sized + VecDrop> {
 /// assert_eq!(*vec, [7, 1, 2, 3]);
 /// ```
 pub type Vec<T, const N: usize> = VecInner<[MaybeUninit<T>; N]>;
-/// TODO: doc
+/// A Vec with dynamic capacity
+///
+/// [`Vec`]() coerces to `VecView`. `VecView` is !Sized, meaning that it can only ever be used through pointer
+///
+/// Unlike [`Vec`](), `VecView` does not have an `N` const-generic parameter.
+/// This has the ergonomic advantages of making it possible to use functions without needing to know at
+/// compile-time the size of the buffers used, for example for use in `dyn` traits.
+///
+/// `VecView<T>` is to `Vec<T, N>` what `[T]` is to `[T; N]`
+///
+/// ```rust
+/// use heapless::{Vec, VecView};
+///
+/// let mut vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+/// let view: &VecView<_> = &vec;
+/// assert_eq!(view, &[1, 2, 3, 4]);
+///
+/// let mut_view: &mut VecView<_> = &mut vec;
+/// mut_view.push(5);
+/// assert_eq!(vec, [1, 2, 3, 4, 5]);
+/// ```
 pub type VecView<T> = VecInner<[MaybeUninit<T>]>;
 
 impl<T> VecView<T> {
@@ -86,8 +106,8 @@ impl<T> VecView<T> {
     /// # Examples
     ///
     /// ```
-    /// use heapless::Vec;
-    /// let buffer: Vec<u8, 5> = Vec::from_slice(&[1, 2, 3, 5, 8]).unwrap();
+    /// use heapless::{Vec, VecView};
+    /// let buffer: &VecView<u8> = &Vec::<u8, 5>::from_slice(&[1, 2, 3, 5, 8]).unwrap();
     /// assert_eq!(buffer.as_slice(), &[1, 2, 3, 5, 8]);
     /// ```
     pub fn as_slice(&self) -> &[T] {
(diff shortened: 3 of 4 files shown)


sosthene-nitrokey:
I'm surprised by the miri failures, especially that hey don't happen in #424

sosthene-nitrokey:
I think I fixed the documentation issues. By having a 

```rust
#[cfg(doc)]
#[doc(hidden)]
pub use vec::VecInner as __private_api_do_not_use;
```

The documentation for the aliases is generated with all the methods.

The miri test fails were because not all methods were forwarded to the `View` implementation, but it seems to raise some very subtle behaviour with the `as_mut_ptr` method. It looks like there is some subtlety with stacked borrows when mix-matching methods on the `view` and methods on the owned type. I'm not really sure what this is about.

burrbull:
> The documentation for the aliases is generated with all the methods.

Yeah this is painful.

sosthene-nitrokey:
> Yeah this is painful.

What do you mean by that?

What I meant was that without the `#[doc(hidden)]` line the aliases where not documented at all by `rustdoc`. With the line, the aliases now properly all the traits they implement as well as the method that are available. Now the docs look like that:

image

Which I think is almost as good as it can get. The only "confusing" part is the second line showing it being an alias

burrbull:
> What do you mean by that?

Sorry. It was just a comment.

`svd2rust` has the same issue and I also have not found better solution.
Type aliases make code much clearer and faster to compile, but documentation is not good due to this issue.

src/lib.rs
```diff
@@ -89,7 +89,10 @@ pub use indexmap::{
 pub use indexset::{FnvIndexSet, IndexSet, Iter as IndexSetIter};
 pub use linear_map::LinearMap;
 pub use string::String;
-pub use vec::Vec;
+#[cfg(doc)]
+#[doc(hidden)]
+pub use vec::VecInner as _;
+pub use vec::{Vec, VecView};
```
sosthene-nitrokey, line 95: This is not required any-more with the latest nightly, but is needed to work with stable.

reitermarkus: By “this” you mean only the `doc(hidden)` workaround? If so, does it matter given `docs.rs` uses nightly anyways?

src/vec.rs
```diff
@@ -7,6 +7,35 @@ use core::{
...
+}
+
+impl<T> VecDrop for [MaybeUninit<T>] {
+    fn drop_with_len(&mut self, _len: usize) {
+        // Case of a view, drop does nothing
+    }
+}
+
+impl<T, const N: usize> VecDrop for [MaybeUninit<T>; N] {
+    fn drop_with_len(&mut self, len: usize) {
```
reitermarkus, line 22: This should be unsafe, given there is no bounds checking in the implementation.

src/lib.rs
sosthene-nitrokey: Yes this is what I mean. `docs.rs` uses nightly, but the workaround might still be useful for people using `cargo doc` locally (I often do so that the search covers all my dependencies).

src/vec.rs
sosthene-nitrokey: Correct, I made this unsafe.

src/lib.rs
reitermarkus: I guess it doesn't hurt, but let's add a comment why `doc(hidden)` is needed.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `is_full`, `recent_index`, `oldest`, and `oldest_index` to `HistoryBuffer`
 - Added infallible conversions from arrays to `Vec`.
 - Added `Vec::spare_capacity_mut`.
+- Added `VecView`, the `!Sized` version of `Vec`
```
reitermarkus, line 17:
```suggestion
- Added `VecView`, the `!Sized` version of `Vec`.
```

src/vec.rs
```diff
@@ -7,6 +7,35 @@ use core::{
...
+
+impl<T> VecDrop for [MaybeUninit<T>] {
+    unsafe fn drop_with_len(&mut self, _len: usize) {
+        // Case of a view, drop does nothing
+    }
+}
+
+impl<T, const N: usize> VecDrop for [MaybeUninit<T>; N] {
+    unsafe fn drop_with_len(&mut self, len: usize) {
+        // NOTE(unsafe) avoid bound checks in the slicing operation
```
reitermarkus, line 23: This only explains why unsafe code is used, not why it is safe. Also, the `unsafe` blocks in here are not needed anymore.

src/vec.rs
```diff
@@ -35,117 +64,41 @@ use core::{
...
-        for elem in self {
-            unsafe {
-                new.push_unchecked(elem.clone());
-            }
-        }
-        new
-    }
+/// A [`Vec`] with dynamic capacity
+///
+/// [`Vec`] coerces to `VecView`. `VecView` is `!Sized`, meaning that it can only ever be used through pointer
```
reitermarkus, line 80:
```suggestion
/// [`Vec`] coerces to `VecView`. `VecView` is `!Sized`, meaning it can only ever be used by reference.
```

src/vec.rs
```diff
@@ -35,117 +64,41 @@ use core::{
...
+///
+/// Unlike [`Vec`], `VecView` does not have an `N` const-generic parameter.
+/// This has the ergonomic advantages of making it possible to use functions without needing to know at
```
reitermarkus, line 83:
```suggestion
/// This has the ergonomic advantage of making it possible to use functions without needing to know at
```

src/vec.rs
```diff
@@ -1180,18 +1952,41 @@ where
...
+{
+    fn eq(&self, other: &VecView<A>) -> bool {
+        <[A]>::eq(other, *self)
     }
 }
 
 // Implements Eq if underlying data is Eq
 impl<T, const N: usize> Eq for Vec<T, N> where T: Eq {}
+// Not sure about this one? Should two distinct capacities really be Eq?
+// Anyways it derefs to &[u8] which implements Eq so I suppose it should
```
reitermarkus, line 1973: You can remove this comment. We already have `Eq` for `Vec` itself, so it makes sense to also have it for `VecView`. We should probably also have `Eq` for `Vec`s with different capacities.

sosthene-nitrokey: Ok

src/vec.rs
sosthene-nitrokey: Fixed

src/vec.rs
```diff
@@ -35,117 +64,41 @@ use core::{
...
-    }
+/// A [`Vec`] with dynamic capacity
+///
+/// [`Vec`] coerces to `VecView`. `VecView` is `!Sized`, meaning that it can only ever be used through pointer
+///
+/// Unlike [`Vec`], `VecView` does not have an `N` const-generic parameter.
+/// This has the ergonomic advantages of making it possible to use functions without needing to know at
+/// compile-time the size of the buffers used, for example for use in `dyn` traits.
+///
+/// `VecView<T>` is to `Vec<T, N>` what `[T]` is to `[T; N]`
```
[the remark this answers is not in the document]
sosthene-nitrokey: Fixed.

src/vec.rs
sosthene-nitrokey: Fixed

src/vec.rs
sosthene-nitrokey: Good point. Fixed in https://github.com/rust-embedded/heapless/pull/425/commits/fe3dd4d64af313533999e2014361fcd3c2db31cf

src/lib.rs
sosthene-nitrokey: I added a comment with a link to the rustdoc issue that this works around and with justification for the `#[doc(hidden)]` and the `#[cfg(doc)]`.

src/lib.rs
```diff
@@ -89,6 +89,10 @@ pub use indexmap::{
 pub use indexset::{FnvIndexSet, IndexSet, Iter as IndexSetIter};
 pub use linear_map::LinearMap;
 pub use string::String;
+
+// Workaround https://github.com/rust-lang/rust/issues/119015. This is required so that the methods on `VecView` and `Vec` are properly documented.
+// cfg(doc) prevents `VecInner` being part of the public API.
```
reitermarkus, line 94: Is `cfg(doc)` actually necessary? I don't think you could even import `heapless::_` if you wanted to anyways.

sosthene-nitrokey: I don't think it is but I'm not certain. I trust the behaviour of `cfg` better than that of `_`.

src/vec.rs
```diff
@@ -876,60 +807,730 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// ```
+    /// use heapless::Vec;
+    ///
+    /// // allocate the vector on the stack
+    /// let mut x: Vec<u8, 16> = Vec::new();
+    ///
+    /// // allocate the vector in a static variable
+    /// static mut X: Vec<u8, 16> = Vec::new();
+    /// ```
+    /// `Vec` `const` constructor; wrap the returned value in [`Vec`].
```
reitermarkus, line 851:
```suggestion
```

This wasn't added in this PR, but this sentence doesn't make any sense.

src/vec.rs
```diff
@@ -876,60 +807,730 @@ impl<T, const N: usize> Vec<T, N> {
...
+
+    /// Get a reference to the Vec, erasing the `N` const-generic
+    ///
+    /// This can also be used through type coerction, since `Vec<T, N>` implements `Unsize<VecView<T>>`:
+    ///
+    /// ```rust
+    /// use heapless::{Vec, VecView};
+    ///
+    /// let vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+    /// let view: &VecView<_> = &vec;
```
reitermarkus, line 942: This should document `as_view`, not type coersion.

src/vec.rs
```diff
@@ -876,60 +807,730 @@ impl<T, const N: usize> Vec<T, N> {
...
+
+    /// Get a `mut` reference to the Vec, erasing the `N` const-generic
+    ///
+    /// This can also be used through type coerction, since `Vec<T, N>` implements `Unsize<VecView<T>>`:
+    ///
+    /// ```rust
+    /// use heapless::{Vec, VecView};
+    ///
+    /// let mut vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+    /// let view: &mut VecView<_> = &mut vec;
```
reitermarkus, line 956: This should document `as_mut_view`, not type coersion.

src/vec.rs
```diff
@@ -876,60 +807,730 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// use heapless::{Vec, VecView};
+    ///
+    /// let vec: Vec<u8, 10> = Vec::from_slice(&[1, 2, 3, 4]).unwrap();
+    /// let view: &VecView<_> = &vec;
+    /// ```
+    pub const fn as_view(&self) -> &VecView<T> {
+        self
+    }
+
+    /// Get a `mut` reference to the Vec, erasing the `N` const-generic
```
reitermarkus, line 948:
```suggestion
    /// Get a mutable reference to the `Vec`, erasing the `N` const-generic.
```

src/vec.rs
```diff
@@ -876,60 +807,730 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// Equivalent to `&s[..]`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::Vec;
+    /// let buffer: Vec<u8, 5> = Vec::from_slice(&[1, 2, 3, 5, 8]).unwrap();
+    /// assert_eq!(buffer.as_slice(), &[1, 2, 3, 5, 8]);
+    /// ```
+    pub fn as_slice(&self) -> &[T] {
```
reitermarkus, line 973: This should already be available through `Deref`, right?

reitermarkus (changes requested):

src/vec.rs
sosthene-nitrokey: I'm not sure I like replacing `mut` with "mutable", since `mut` doesn't necessarily mean "mutable", but rather means "exclusive", and a non-mut reference is shared.

Not a hill I'm ready to die on though.

sosthene-nitrokey:
I made a detailed blog post explaining the motivations and possible approaches to this PR: https://sgued.fr/blog/heapless-howto/

src/vec.rs
reitermarkus: > since `mut` doesn't necessarily mean "mutable"

`mut` is an abbreviation for mutable, so it does mean that quite literally, in addition to meaning “exclusive” in the context of Rust references.

sosthene-nitrokey: Here is an example of `mut` being called an "exclusive" : https://google.github.io/comprehensive-rust/references/exclusive.html

Exclusive is more in line with the existence of interior mutability, but exclusive doesn't necessarily map to the `mut` keyword, because it's a shorthand for "mutable", so I prefer referring to the keyword directly, so there is no ambiguity at all, and everyone still understands it.

reitermarkus:
I have been thinking about the naming here.

Would it make sense to switch the naming to have `Vector` and `&Vec`, similar to `String` and `&str`. I think this would also make all APIs using `Vec` more similar to `std::collections::Vec`.

sosthene-nitrokey:
I'm not sure this is a great idea to rename `Vec` to `Vector` and have the `VecView` be `Vec`:

- Firstly this would be a significant breaking change requiring a lot of downstream churn when released. 
- Secondly, most people looking at the documentation will search for `Vec` when looking for an equivalent to std's `Vec`, and they most likely will first look for `Vec` and be confused that it can't be constructed like that, and I also assume that `VecView` will not be a API that is used in the majority of cases, so making the documentation be straightforward for the owned case. 
- What would the equivalent for `StringView` be? Naming it `str` feels like asking for trouble.

> I think this would also make all APIs using Vec more similar to std::vec::Vec given they both don't have const generic type parameters.

I think that `Vec` is closer to `std::collections::Vec` than `VecView`, not because of the type definitions/the generics, but rather because of what they are used for: An dynamic number of owned `T`. 

`VecView` for me looks closer to `[T]` than to `Vec`, or  [`dyn bytes::BufMut`](https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html)

reitermarkus:
> Firstly this would be a significant breaking change requiring a lot of downstream churn when released.

Yes, which is why I am asking this now so we don't immediately need a breaking change for `VecView` as well, which would make it more complex than just replacing `Vec` with `Vector`/`VecBuf`. It really comes down to how much we care about following the `std` API (https://github.com/rust-embedded/heapless/issues/442).

> What would the equivalent for `StringView` be? Naming it `str` feels like asking for trouble.

I guess `StringBuf` -> `&String` -> `&str`?

sosthene-nitrokey:
Given #442 it indeed makes sense to do the renaming you want, but I disagree with the main point of #442 and therefore I disagree with this renaming too.

reitermarkus (approved):
This is working as intended. I don't think the naming is a blocker, but this should be discussed further in https://github.com/rust-embedded/heapless/issues/442.

sosthene-nitrokey:
Once this is merged I'll make a similar PR for String and the other types in heapless that could benefit from this pattern.

reitermarkus:
Thanks, @sosthene-nitrokey!

PR #427, opened 2023-12-20
`Vec` improvements

### LINKED ISSUE #426: Feature: Add `Vec::spare_capacity_mut`
It'd be great to have spare_capacity_mut method, similar to [what `alloc::vec::Vec` has](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut). 

```rust
impl<T, const N: usize> Vec<T, N> {
    pub fn spare_capacity_mut(&mut self) -> &mut [MaybeUninit<T>] { /* ... */ }
}
```

Having access to spare capacity enables certain use-cases, for instance, I wanted to securely zeroize a vector, and to do that I need to zeroize the spare data as well.

YuhanLiin:
Add `spare_capacity_mut` and `From` implementation for `Vec`, as well as `vec!` macro.
Resolves #426 and #344

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -1241,6 +1241,39 @@ where
     }
 }
 
+#[macro_export]
+/// Shorthand for defining vectors using array syntax.
+///
+/// Create `Vec` with a list of elements and capacity:
+/// ```
+/// let v = heapless::vec![1, 2, 3; 7];
+/// assert_eq!(v.as_slice(), &[1, 2, 3]);
+/// assert_eq!(v.capacity(), 7);
+/// ```
+///
+/// Create `Vec` with a repeated element, length, and capacity:
+/// ```
+/// let v = heapless::vec!['a'; 3; 4];
+/// assert_eq!(v.as_slice(), &['a', 'a', 'a']);
+/// assert_eq!(v.capacity(), 4);
+/// ```
+///
+/// Unlike the `std` version of this macro, the repeat element must be `Copy` or a constant, like
+/// with repeat elements in array expressions.
+macro_rules! vec {
+    (; $cap:expr) => {
+        heapless::Vec::<_, $cap>::new()
+    };
+
+    ($($elem:expr),+ ; $cap:expr) => {
+        heapless::Vec::<_, $cap>::from([$($elem),+])
+    };
+
+    ($elem:expr ; $len:expr ; $cap:expr) => {
+        heapless::Vec::<_, $cap>::from([$elem; $len])
+    };
+}
+
 #[cfg(test)]
 mod tests {
     use crate::Vec;
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -821,6 +821,36 @@ impl<T, const N: usize> Vec<T, N> {
         // All item are processed. This can be optimized to `set_len` by LLVM.
         drop(g);
     }
+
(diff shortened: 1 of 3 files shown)


src/vec.rs
```diff
@@ -858,6 +888,17 @@ impl<T, const N: usize> Drop for Vec<T, N> {
     }
 }
 
+impl<T, const N: usize> From<[T; N]> for Vec<T, N> {
+    /// Converts array to `Vec` of same size and capacity without copying
+    fn from(buffer: [T; N]) -> Self {
+        Self {
+            // cast [T; N] into [MaybeUninit<T>; N]
+            buffer: unsafe { buffer.as_ptr().cast::<[MaybeUninit<T>; N]>().read() },
```
Dirbaio, line 896: this does a bitwise copy of the elements into the Vec, but will still drop them when `buffer` goes out of scope. This will cause use-after-frees if `T` is an owned type, like `Box`.

You can wrap the `buffer` into a `ManuallyDrop` to avoid this.

src/vec.rs
```diff
@@ -858,6 +888,17 @@ impl<T, const N: usize> Drop for Vec<T, N> {
     }
 }
 
+impl<T, const N: usize> From<[T; N]> for Vec<T, N> {
```
Dirbaio, line 891: it feels a bit strange that the array length becomes the capacity. Intuitively I'd say it should be possible to convert a `[u8; 10]` into a `Vec<u8; 20>` with 10 filled elements.

This would fail if `CAP < N`, and there's no way to add `where CAP >= N` in const generics, so tihs'd have to be TryFrom which perhaps is not that ergonomic?

sosthene-nitrokey: There is a way to generate a compile error in that case. This PR does this: https://github.com/rust-embedded/heapless/pull/352

YuhanLiin: I thought doing it this way would allow me to "move" the array into the Vec without copying, but it turns out that copying will happen either way, so might as well allow N < CAP.

src/vec.rs
```diff
@@ -1175,6 +1241,39 @@ where
...
+/// ```
+///
+/// Unlike the `std` version of this macro, the repeat element must be `Copy` or a constant, like
+/// with repeat elements in array expressions.
+macro_rules! vec {
+    (; $cap:expr) => {
+        heapless::Vec::<_, $cap>::new()
+    };
+
+    ($($elem:expr),+ ; $cap:expr) => {
```
Dirbaio, line 1268: this should be consistent with std. 

As it's written now, `vec![42; 5]` in std it would create a vec with five 42's, while in heapless it creates a vec with a single 42 with a capacity of 5.

IMO the macro should behave like this:

- `vec![1,2,3]`: 3 elements, capacity is inferred.
- `vec![42; 3]`: element repeated 3 times, capacity is inferred.

and then invent some extra syntax for specifying capacity explicitly that doesn't conflict with that. 

In #369 we had the same issue, we went with `format!(256; "foo {}", 42)` where `256` is the capacity, but that doesn't work well here...

Dirbaio:
in the end i've merged #352 for the "from array" conversion because it was a bit more complete (more tests). 

Could you open separate PRs for the two remaining features so they can be reviewed/discussed separately? (the `vec!` macro, and `spare_capacity_mut`).

YuhanLiin:
Continued in #437 and #438

PR #429, opened 2023-12-26
Implemented `GenericVec` and redefined `Vec` as a type alias for `GenericVec`.

JohnScience:
Implemented `AlignedVec` gated by the `aligned_vec` feature. `AlignedVec<T,A,N>` is similar to `Vec<T,N>` but also ensures that the buffer is aligned just like the type `A`. This can be useful for transmutation of the buffer to other types.

`AlignedVec` mimics all implementations of `Vec` and adds the following implementation:

```
impl<A, const N: usize> AlignedVec<u8, A, N> {
    /// Transmutes the aligned buffer to another type.
    /// The length of the `AlignedVec` does not participate
    /// in the transmutation.
    ///
    /// # Example
    ///
    /// ```
    /// let mut v = AlignedVec::<u8, /*alignas*/ u32, 4>::new();
    /// v.extend_from_slice(&[0, 0, 0, 0]).unwrap();
    /// let number = unsafe { v.transmute_buffer::<u32>() };
    /// assert_eq!(number, 0);
    /// ```
    ///
    /// # Safety
    ///
    /// The size of the buffer (not the type) must be the same as the size of T2.
    /// The alignment of the buffer must be the same as the alignment of T2.
    ///
    /// # Panics
    ///
    /// In debug mode, this function panics if the size of the buffer
    /// is not the same as the size of T2 or (as a safety measure) if the
    /// alignment of the buffer is not the same as the alignment of T2.
    pub unsafe fn transmute_buffer<T2>(self) -> T2
    where
        T2: Sized,
    {
        // transmute_unchecked is unstable
        // transmute wouldn't work because we can't guarantee that
        // the size of the buffer is the same as the size of T2

        // #[cfg(debug_assertions)] was used here to have a chance of constant implementation.
        // However, `pointer::align_offset` is not a constant function at the time of writing.
        #[cfg(debug_assertions)]
        if core::mem::size_of::<T2>() != self.len {
            panic!("AlignedVec::transmute_buffer: size mismatch")
        }
        let buffer_ptr = self.buffer.as_ptr();
        #[cfg(debug_assertions)]
        if buffer_ptr as usize % core::mem::align_of::<A>() != 0 {
            panic!("AlignedVec::transmute_buffer: alignment of the buffer is incorrect")
        }
        core::ptr::read(buffer_ptr as *const T2)
    }
}
```

This can be useful for binary protocols.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `format` macro.
 - Added `String::from_utf16`.
+- Added `is_full`, `recent_index`, `oldest`, and `oldest_index` to `HistoryBuffer`
 
 ### Changed
 
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -141,6 +141,12 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
         N
     }
 
+    /// Returns whether the buffer is full
+    #[inline]
+    pub fn is_full(&self) -> bool {
+        self.filled
+    }
+
     /// Writes an element to the buffer, overwriting the oldest value.
     pub fn write(&mut self, t: T) {
         if self.filled {
@@ -182,14 +188,70 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
     /// assert_eq!(x.recent(), Some(&10));
     /// ```
     pub fn recent(&self) -> Option<&T> {
+        self.recent_index()
+            .map(|i| unsafe { &*self.data[i].as_ptr() })
+    }
+
+    /// Returns index of the most recently written value in the underlying slice.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::HistoryBuffer;
+    ///
+    /// let mut x: HistoryBuffer<u8, 16> = HistoryBuffer::new();
+    /// x.write(4);
+    /// x.write(10);
+    /// assert_eq!(x.recent_index(), Some(1));
+    /// ```
+    pub fn recent_index(&self) -> Option<usize> {
         if self.write_at == 0 {
             if self.filled {
-                Some(unsafe { &*self.data[self.capacity() - 1].as_ptr() })
+                Some(self.capacity() - 1)
             } else {
                 None
             }
         } else {
-            Some(unsafe { &*self.data[self.write_at - 1].as_ptr() })
+            Some(self.write_at - 1)
+        }
+    }
+
+    /// Returns a reference to the oldest value in the buffer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::HistoryBuffer;
+    ///
+    /// let mut x: HistoryBuffer<u8, 16> = HistoryBuffer::new();
+    /// x.write(4);
+    /// x.write(10);
+    /// assert_eq!(x.oldest(), Some(&4));
+    /// ```
+    pub fn oldest(&self) -> Option<&T> {
+        self.oldest_index()
+            .map(|i| unsafe { &*self.data[i].as_ptr() })
+    }
+
+    /// Returns index of the oldest value in the underlying slice.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::HistoryBuffer;
+    ///
+    /// let mut x: HistoryBuffer<u8, 16> = HistoryBuffer::new();
+    /// x.write(4);
+    /// x.write(10);
+    /// assert_eq!(x.oldest_index(), Some(0));
+    /// ```
+    pub fn oldest_index(&self) -> Option<usize> {
+        if self.filled {
+            Some(self.write_at)
+        } else if self.write_at == 0 {
+            None
+        } else {
+            Some(0)
         }
     }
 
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -15,17 +15,26 @@ repository = "https://github.com/rust-embedded/heapless"
...
 [features]
+
 # Enable polyfilling of atomics via `portable-atomic`.
 # `portable-atomic` polyfills some functionality by default, but to get full atomics you must
 # enable one of its features to tell it how to do it. See `portable-atomic` documentation for details.
 portable-atomic = ["dep:portable-atomic"]
 
 # Enable polyfilling of atomics via portable-atomic, using critical section for locking
-portable-atomic-critical-section = ["dep:portable-atomic", "portable-atomic", "portable-atomic?/critical-section"]
+portable-atomic-critical-section = [
```
Dirbaio, line 25: please don't autoformat .toml files

Cargo.toml
```diff
@@ -39,6 +48,9 @@ defmt-03 = ["dep:defmt"]
 # Enable larger MPMC sizes.
 mpmc_large = []
 
+# Implement AlignedVec
+aligned_vec = []
```
Dirbaio, line 52: there's no need to gate it behind a feature, the only reason to do it is compile time and heapless already compiles really fast.

Dirbaio:
This is copypasting literally all `Vec` code, which I don't think is maintainable. We should find a way to share the code somehow, for example making the main `Vec` generic over a "storage".

JohnScience:
> This is copypasting literally all `Vec` code, which I don't think is maintainable. We should find a way to share the code somehow, for example making the main `Vec` generic over a "storage".

My concern is that they want to have different order of fields. `AlignedVec` wants length last so that it can be reused after transmuting the buffer. `Vec` wants length `first` for optimization reasons:

```rust
pub struct Vec<T, const N: usize> {
    // NOTE order is important for optimizations. the `len` first layout lets the compiler optimize
    // `new` to: reserve stack space and zero the first word. With the fields in the reverse order
    // the compiler optimizes `new` to `memclr`-ing the *entire* stack space, including the `buffer`
    // field which should be left uninitialized. Optimizations were last checked with Rust 1.60
    len: usize,

    buffer: [MaybeUninit<T>; N],
}
```

Dirbaio:
why does AlignedVec need len last? Is it for `transmute_buffer`? AFAICT that can still work even if len is first.

Also note `Vec` is `repr(Rust)` which lets the compiler whether to place len first or last, whatever it deems to be more optimized. I think it should be possible to do the same with `AlignedVec` like this::

```rust
pub struct AlignedVec<T, A, const N: usize> {
    len: usize,
    buffer: AlignedBuf<T, A, N>,
}

#[repr(C)]
pub struct AlignedBuf<T, A, const N: usize> {
    // this is a widely used hack to make the buffer aligned
    alignment_field: [A; 0],
    buffer: [MaybeUninit<T>; N],
}
```

and if you do that, it becomes the same as the regular `Vec` except the type of the buffer changes. Hence the idea of making it generic on a "storage" thing, which could use the regular or the aligned storage (or the regular storage could be just the aligned one to u8)

JohnScience:
> why does AlignedVec need len last? Is it for `transmute_buffer`? AFAICT that can still work even if len is first.
> 
> Also note `Vec` is `repr(Rust)` which lets the compiler whether to place len first or last, whatever it deems to be more optimized. I think it should be possible to do the same with `AlignedVec` like this::
> 
> ```rust
> pub struct AlignedVec<T, A, const N: usize> {
>     len: usize,
>     buffer: AlignedBuf<T, A, N>,
> }
> 
> #[repr(C)]
> pub struct AlignedBuf<T, A, const N: usize> {
>     // this is a widely used hack to make the buffer aligned
>     alignment_field: [A; 0],
>     buffer: [MaybeUninit<T>; N],
> }
> ```
> 
> and if you do that, it becomes the same as the regular `Vec` except the type of the buffer changes. Hence the idea of making it generic on a "storage" thing, which could use the regular or the aligned storage (or the regular storage could be just the aligned one to u8)

My reasoning was that having length first would prevent reusing the freed memory because it would be too far from the stack pointer after filling the buffer and, if the memory after freeing `length` field is reused anyway, it would cause cache misses for large structures. However, I didn't check it with godbolt, so I can easily be wrong.

JohnScience:
> why does AlignedVec need len last? Is it for `transmute_buffer`? AFAICT that can still work even if len is first.
> 
> Also note `Vec` is `repr(Rust)` which lets the compiler whether to place len first or last, whatever it deems to be more optimized. I think it should be possible to do the same with `AlignedVec` like this::
> 
> ```rust
> pub struct AlignedVec<T, A, const N: usize> {
>     len: usize,
>     buffer: AlignedBuf<T, A, N>,
> }
> 
> #[repr(C)]
> pub struct AlignedBuf<T, A, const N: usize> {
>     // this is a widely used hack to make the buffer aligned
>     alignment_field: [A; 0],
>     buffer: [MaybeUninit<T>; N],
> }
> ```
> 
> and if you do that, it becomes the same as the regular `Vec` except the type of the buffer changes. Hence the idea of making it generic on a "storage" thing, which could use the regular or the aligned storage (or the regular storage could be just the aligned one to u8)

I thought more about it and I have doubts that it's reasonable to make `Vec` generic over storage because it is itself the storage. I believe that the better approach is the definition of `Vec` and `AlignedVec` as types-aliases of semver-exempt `GenericVec`.

Currently, `AlignedVec` should be the same thing as `GenericVec`.

JohnScience:
I just looked at the diff. It seems that the `not-send.stderr` has to be changed too: https://www.diffchecker.com/I7lkNq8o/

JohnScience:
@Dirbaio Ready for the second review!

burrbull:
Why not to just make `A` a last generic with default value?:
```
struct Vec<T, const N: usize, A = u8> { .. }
```

JohnScience:
The PR has been revived at https://github.com/rust-embedded/heapless/pull/430

JohnScience:
@burrbull Your idea has been implemented.

PR #430, opened 2023-12-27
Generalized `Vec`, implemented `Vec::transmute_buffer`.

JohnScience:


JohnScience:
Ready for the review!

JohnScience:
@Dirbaio

src/vec.rs
```diff
@@ -28,20 +34,41 @@ use core::{cmp::Ordering, fmt, hash, iter::FromIterator, mem::MaybeUninit, ops,
...
-pub struct Vec<T, const N: usize> {
+pub struct Vec<T, const N: usize, A = u8> {
     // NOTE order is important for optimizations. the `len` first layout lets the compiler optimize
     // `new` to: reserve stack space and zero the first word. With the fields in the reverse order
     // the compiler optimizes `new` to `memclr`-ing the *entire* stack space, including the `buffer`
     // field which should be left uninitialized. Optimizations were last checked with Rust 1.60
     len: usize,
 
-    buffer: [MaybeUninit<T>; N],
+    storage: VecBuf<T, N, A>,
```
Dirbaio, line 44: could you keep the name `buffer` to keep the diff small?

src/vec.rs
```diff
@@ -28,20 +34,41 @@ use core::{cmp::Ordering, fmt, hash, iter::FromIterator, mem::MaybeUninit, ops,
...
     const INIT: [MaybeUninit<T>; N] = [Self::ELEM; N]; // important for optimization of `new`
 
+    const fn new() -> Self {
+        Self {
+            alignment_field: [],
+            buffer: Self::INIT,
+        }
+    }
+
+    const fn as_ptr(&self) -> *const T {
```
Dirbaio, line 58: you can implement `Deref` and `DerefMut` into `[T; N]` so all the usual slice methods are instantly available, without you having to wrap them.

Dirbaio:
i'm not 100% convinced we should have `transmute_buffer` in the public API, it's quite easy to misuse and a bit niche. With the alignment generic, it should be possible for user code to safely do any transmuting on its own.

src/vec.rs
JohnScience: > you can implement `Deref` and `DerefMut` into `[T; N]` so all the usual slice methods are instantly available, without you having to wrap them.

My main concern is that then the methods won't be `const`.

src/vec.rs
JohnScience: @Dirbaio Fixed.

JohnScience:
> i'm not 100% convinced we should have `transmute_buffer` in the public API, it's quite easy to misuse and a bit niche. With the alignment generic, it should be possible for user code to safely do any transmuting on its own.

In defense of `transmute buffer`, this method is the primary reason for addition of the alignment type parameter. 

In order to avoid the pollution of the namespace, this method is provided only for `T=u8`.

```
impl<const N: usize, A> Vec<u8, N, A> {
 // fn transmute_buffer
}
```

In order to make the transmutation in user code safe, it has to either expose this function or expose the buffer. Otherwise, it'd be impossible/hard to do safely. And I believe that hiding this implementation detail is quite important.

JohnScience:
Ready for review!

src/vec.rs
```diff
@@ -170,7 +214,7 @@ impl<T, const N: usize> Vec<T, N> {
     pub fn as_slice(&self) -> &[T] {
         // NOTE(unsafe) avoid bound checks in the slicing operation
         // &buffer[..self.len]
-        unsafe { slice::from_raw_parts(self.buffer.as_ptr() as *const T, self.len) }
+        unsafe { slice::from_raw_parts(self.buffer.as_ptr(), self.len) }
```
reitermarkus, line 217:
```suggestion
        unsafe { slice::from_raw_parts(self.as_ptr(), self.len) }
```

src/vec.rs
```diff
@@ -35,20 +41,56 @@ use core::{
...
+
+    const fn as_ptr(&self) -> *const T {
+        self.buffer.as_ptr() as *const T
+    }
+
+    fn as_mut_ptr(&mut self) -> *mut T {
+        self.buffer.as_mut_ptr() as *mut T
+    }
+
+    unsafe fn get_unchecked_mut(&mut self, index: usize) -> &mut MaybeUninit<T> {
```
reitermarkus, line 74: This is now already provided through `DerefMut`.

JohnScience:
@reitermarkus Fixed

src/vec.rs
reitermarkus: Does this actually have to be `const`?

src/vec.rs
```diff
@@ -859,25 +899,75 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// - The size of the output type must be equal to the size of the **buffer** (rather than [`Vec`]).
+    /// - The alignment of the `Vec` must be equal or be a multiple of the alignment of the output.
+    ///
+    /// # Panics
+    ///
+    /// In debug mode, this method panics...
+    ///
+    /// - If the buffer isn't full.
+    /// - If the size of the output type doesn't match the size of the buffer.
+    /// - If the alignment of the `Vec` isn't equal or a multiple of the alignment of the output.
```
reitermarkus, line 928: I don't think these panics should be documented, given they only happen in debug mode.

JohnScience:
@reitermarkus Fixed.

Dirbaio:
> I'll try elaborate on the choice here. In order to make the transmutation in user code safe, it has to either expose this function (transmute_buffer) or expose the buffer.

We already expose the buffer with `.as_ptr()` / `.as_mut_ptr()`, so it's already possible to do `transmute_buffer` from user code soundly.

src/vec.rs
```diff
@@ -859,25 +899,75 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::Vec;
+    ///
+    /// let mut v: Vec<u8, 2, u16> = Vec::new();
+    /// v.extend_from_slice(&[0, 0]).unwrap();
+    /// let number: u16 = unsafe { v.transmute_buffer() };
+    /// assert_eq!(number, 0u16);
+    /// ```
```
reitermarkus, line 931: I'd like to see a more elaborate example here. While the example shows how to use this method, it is a bit too trivial and doesn't give me a good idea why this is needed in practice. Especially the fact that the `Vec` needs to be full feels quite limiting. In the example, you could transmute to `[MaybeUninit<u16>; 1]` without it being full. Of course you wouldn't be able to read the partially uninitialized value, but with a longer buffer it may make sense.

I think ideally, we would have a safe method that returns `[MaybeUninit<A>; M]`, but there currently doesn't seem a way to calculate or even assert the size of `M` with const generics.

src/vec.rs
```diff
@@ -859,25 +899,75 @@ impl<T, const N: usize> Vec<T, N> {
...
+        };
+        #[cfg(debug_assertions)]
+        if N * core::mem::size_of::<u8>() != core::mem::size_of::<O>() {
+            panic!("Vec::transmute_buffer: size mismatch");
+        };
+        #[cfg(debug_assertions)]
+        if core::mem::align_of::<A>() % core::mem::align_of::<O>() != 0 {
+            panic!("Vec::transmute_buffer: alignment mismatch");
+        };
+        let inner_buf_ref: &[MaybeUninit<u8>; N] = &self.buffer;
```
reitermarkus, line 947: I think `self` needs to be wrapped in a `ManuallyDrop` before this.

JohnScience:
> > I'll try elaborate on the choice here. In order to make the transmutation in user code safe, it has to either expose this function (transmute_buffer) or expose the buffer.
> 
> We already expose the buffer with `.as_ptr()` / `.as_mut_ptr()`, so it's already possible to do `transmute_buffer` from user code soundly.

Another reason why I'm not particularly happy with leaving this to the user is that with the stabilization of [`std::intrinsics::transmute_unchecked`](https://doc.rust-lang.org/std/intrinsics/fn.transmute_unchecked.html) we will be able to no longer rely on the optimization of the unnecessary copy created with [`std::mem::transmute_copy`](https://doc.rust-lang.org/std/mem/fn.transmute_copy.html).

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `format` macro.
 - Added `String::from_utf16`.
 - Added `is_full`, `recent_index`, `oldest`, and `oldest_index` to `HistoryBuffer`
+- Added an optional generic parameter to `Vec` to specify the alignment of the backing array.
+- Added `transmute_buffer` unsafe method for `Vec<u8, N, A>`.
```
reitermarkus, line 16: This can be removed again.

reitermarkus (approved):

reitermarkus:
@JohnScience, can you rebase this?

reitermarkus:
Closing as outdated. This will need a redo anyways with all the changes to `Vec` in the meantime.

PR #433, opened 2023-12-29
HistoryBuffer: add clear_leak method

### LINKED ISSUE #432: Add a cheap `HistoryBuffer::empty` method
Currently there's no easy way to drain a `HistoryBuffer` without zeroing it.

A simple function like this would come in handy in performance critical situations:

```rust
fn empty(&mut self) {
  self.write_at = 0;
  self.filled = false;
}
```

newAM:
Closes #432.

### DIFF
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -72,8 +72,63 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
 
     /// Clears the buffer, replacing every element with the default value of
     /// type `T`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::HistoryBuffer;
+    ///
+    /// let mut buf: HistoryBuffer<u8, 16> = HistoryBuffer::new();
+    ///
+    /// buf.write(3);
+    /// buf.write(5);
+    /// buf.extend(&[4, 4]);
+    /// assert!(!buf.is_empty());
+    ///
+    /// buf.clear();
+    /// assert!(buf.is_empty());
+    /// ```
     pub fn clear(&mut self) {
         *self = Self::new();
+    }
+
+    /// Clears the buffer **without running the destructor on values**.
+    ///
+    /// Normally, one would use [`clear`] instead to correctly drop the contents
+    /// and thus not leak memory.
+    ///
+    /// This method is sound, but it can leak memory because it will not call
+    /// the destructor for the values.  See [`core::mem::forget`] for more
+    /// information on Rust's safety guarantees.
+    ///
+    /// # Example
+    ///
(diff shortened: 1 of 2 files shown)


newAM:
CI failure fixed in #434

Dirbaio:
I think it should be called `clear` for consistency, that's what it's called  in Vec, String, and the maps.

newAM:
Good catch, I missed that skimming through the `std` docs.

Turns out we also have a clear method already, it's just not as efficient as it could be since it "zeros" the memory with whatever the default value of `T` is.  Unless there's someone doing unsafe memory things this shouldn't be a breaking change to improve it.

src/histbuf.rs
```diff
@@ -70,10 +70,26 @@ impl<T, const N: usize> HistoryBuffer<T, N> {
...
+    /// buf.extend(&[4, 4]);
+    /// assert!(!buf.is_empty());
+    ///
+    /// buf.clear();
+    /// assert!(buf.is_empty());
+    /// ```
     pub fn clear(&mut self) {
-        *self = Self::new();
+        self.write_at = 0;
+        self.filled = false;
```
Dirbaio, line 92: this should drop all the items that were previously in the container. (`*self = ...` did it implicitly because it dropped self before the assignment)

Dirbaio:
I had actually not noticed we already had a `clear` :rofl:

src/histbuf.rs
newAM: :man_facepalming: thanks!  I added a test for this separately here: #435

newAM:
I ended up not changing `clear`.  I could not find a performant way to correctly drop everything.  I instead added separate `clear_leak` to clear the buffer while potentially leaking memory.  I did not make this unsafe because technically leaking memory isn't part of Rust's guarantees.

Dirbaio:
`clear` and `clear_leak` seem to generate the same asm code for types that don't have Drop: https://godbolt.org/z/azEff9b8W

`new()` initializes the buffer to uninit, the optimizer sees most of the written bytes are uninit and the only thing changed are `write_at` and `filled`, and writes just those. So I'm not sure what's up with #432, i've asked there whether they actually saw a perf issue, or it was just theoretical from reading the code.

Dirbaio:
I don't think we should have `clear_leak` if it's as fast as `clear` for types without Drop, because for types *with* Drop it's a footgun.

newAM:
Yup, that makes sense to me.  I committed the classic sin of not using optimizations when doing my own performance profiling :facepalm:

reitermarkus:
Closing since `clear` already generates the same code.

PR #436, opened 2024-01-02
Add missing entries to list of implemented data structures in crate docs.

timonvo:
Adds `Deque`, `HistoryBuffer` and `SortedLinkedList` to the list of implemented data structures in the crate docs. The current list of items seems to be only a subset what's actually been implemented.


Dirbaio (approved):
thanks!

PR #438, opened 2024-01-10
Add vec macros

### LINKED ISSUE #344: Feature request create a vec macro

YuhanLiin:
Add `vec` and `vec_with_cap` macros
Resolves #344

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -131,6 +131,14 @@ impl<T, const N: usize> Vec<T, N> {
         }
     }
 
+    /// Constructs a new vector with fixed capacity of `N`, initializing it with the provided
+    /// array.
+    ///
+    /// The length of the provided array must be `N`.
+    pub fn from_array_same_cap(src: [T; N]) -> Self {
+        Self::from_array::<N>(src)
+    }
+
     /// Clones a vec into a new vec
     pub(crate) fn clone(&self) -> Self
     where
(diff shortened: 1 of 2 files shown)


src/vec.rs
```diff
@@ -131,6 +131,14 @@ impl<T, const N: usize> Vec<T, N> {
         }
     }
 
+    /// Constructs a new vector with fixed capacity of `N`, initializing it with the provided
+    /// array.
+    ///
+    /// The length of the provided array must be `N`.
+    pub fn from_array_same_cap(src: [T; N]) -> Self {
```
reitermarkus, line 138: `N` can be constructed in the macro, so this isn't needed.

reitermarkus:
I think instead of `vec_with_cap`, an API like 

```rust
vec![1, 2, 3]    // Vec::<_, 3>::from([1, 2, 3])
vec![1; 3]       // Vec::<_, 3>::from([1; 3])
vec![1, 2, 3; 6] // Vec::<_, 6>::from([1, 2, 3])
vec![1; 3; 6]    // Vec::<_, 6>::from([1; 3])
```

would work.

birkenfeld:
That's ambiguous though in the case of `vec![1; 3]`.

reitermarkus:
Good point. In that case, maybe `:` as separator would work, e.g. `vec![1: 3]` vs `vec![1; 3: 3]`. This also slightly matches the general `:` syntax for specifying types.

Then again, I think 

```rust
let vec: Vec<_, 6> = vec![1; 3];
```

should work when using `Vec::from_array`, right? So supporting capacity in the macro may not be necessary, to avoid deviating from how the macro works in `std`.

YuhanLiin:
If we don't go with the `vec[1; 3: 3]` version with explicit capacity, then we won't be able to use the macro inline in a lot of places, such as `somefunction(vec![0; 3])`. It'll need to be in a declaration statement with an explicit type like `let v: Vec<_, 3> = vec![...]`, otherwise the compiler might not know the capacity, leading to ambiguity.

PR #439, opened 2024-01-11
Fix tsan.rs CI errors

YuhanLiin:
Fix the current CI error


newAM (approved):
Thanks!

I did a force push just to trigger CI - I don't think github detects when the target branch changes.

PR #443, opened 2024-01-21
Make `Queue::split` `const`.

reitermarkus:
This allows constructing a producer-consumer pair directly in a `static`, e.g.

```rust
static PC: Mutex<RefCell<(Option<Producer<'static, (), 4>>, Option<Consumer<'static, (), 4>>)>> = {
  static mut Q: Queue<(), 4> = Queue::new();
  let (p, c) = unsafe { Q.split() };
  Mutex::new(RefCell::new((Some(p), Some(c))))
};
```


newAM (dismissed):
Can you rebase and update this?

`const_mut_refs` was stabilized in 1.83.0, shouldn't need the separate const/non-const implementations anymore :rocket:

newAM (approved):
Thanks!

PR #444, opened 2024-01-23
Add `Vec::drain` and `String::drain`.

reitermarkus:
Closes https://github.com/rust-embedded/heapless/issues/334 and https://github.com/rust-embedded/heapless/pull/365.

### DIFF
--- a/src/vec/drain.rs
+++ b/src/vec/drain.rs
@@ -1,0 +1,222 @@
+use core::{
+    fmt,
+    iter::FusedIterator,
+    mem::{self, size_of},
+    ptr::{self, NonNull},
+    slice,
+};
+
+use super::VecView;
+
+/// A draining iterator for [`Vec`](super::Vec).
+///
+/// This `struct` is created by [`Vec::drain`](super::Vec::drain).
+/// See its documentation for more.
+///
+/// # Example
+///
+/// ```
+/// use heapless::{vec, Vec};
+///
+/// let mut v = Vec::<_, 4>::from_array([0, 1, 2]);
+/// let iter: vec::Drain<'_, _> = v.drain(..);
+/// ```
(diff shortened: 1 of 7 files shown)


Dirbaio:
could you add tests?

src/vec/drain.rs
```diff
@@ -0,0 +1,163 @@
...
+/// let iter: vec::Drain<'_, _, 3> = v.drain(..);
+/// ```
+pub struct Drain<'a, T: 'a, const N: usize> {
+    /// Index of tail to preserve
+    pub(super) tail_start: usize,
+    /// Length of tail
+    pub(super) tail_len: usize,
+    /// Current remaining range to remove
+    pub(super) iter: slice::Iter<'a, T>,
+    pub(super) vec: NonNull<Vec<T, N>>,
```
sosthene-nitrokey, line 31: If this was built on top of #425 this would allow removing the `const N: usize` generic, and replace the pointer with `NonNull<VecView<T>>`. #425 does not offer the `View` functionality for `String` to keep the PR small, but it would be easy to add it.

It is worth considering because merging this PR and releasing before merging #425 would make it a breaking change to remove  the const generic, so there would need to be 2 implementation of `drain`, one for the view and one for the owned type.

reitermarkus: Yes, I agree it makes sense to base this on `VecView` instead.

reitermarkus:
I can't quite figure out the `miri` error. Note that the implementation is almost 1:1 copied from `std`, so I'm not sure if it's a false positive.

Dirbaio:
rebased it for you on top of the new VecView impl.

Dirbaio (approved):

reitermarkus:
@Dirbaio, thanks!

PR #445, opened 2024-01-23
Only deny warnings in CI.

reitermarkus:
We already pass `--deny warnings` to `clippy`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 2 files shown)


newAM (changes requested):
I think this is a good idea, but clippy alone doesn't have entirely the same coverage as everything else.

At a minimum `env: {"RUSTFLAGS": "-D warnings"}` should be added to the `check` job, and maybe the `cpass` and `miri` jobs too.

reitermarkus:
Still passes with `RUSTFLAGS` applied to everything, so should be good to go.

newAM (approved):

PR #448, opened 2024-01-24
Adds `Deque::make_contiguous`

ansg191:
Adds `Deque::make_contiguous` based on [`VecDeque::make_contiguous`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.make_contiguous) in the std library.

Implementation derived from [here](https://github.com/rust-lang/rust/blob/0b7730105fb7e008002f6ba29f87a55699f67744/library/alloc/src/collections/vec_deque/mod.rs#L2160).

This method provides a contiguous mutable slice to a `Deque`, allowing slice methods like `sort` to be used.

### DIFF
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -207,6 +207,25 @@ impl<T, const N: usize> Deque<T, N> {
     ///
     /// [`as_slices`]: Deque::as_slices
     /// [`as_mut_slices`]: Deque::as_mut_slices
+    ///
+    /// # Examples
+    /// Sorting a deque:
+    /// ```
+    /// use heapless::Deque;
+    ///
+    /// let mut buf = Deque::<_, 4>::new();
+    /// buf.push_back(2).unwrap();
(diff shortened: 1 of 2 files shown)


src/deque.rs
```diff
@@ -866,4 +1043,55 @@ mod tests {
...
+
+        // Deque contains: 0, 1, 2, 3
+        assert_eq!(q.pop_front(), Some(0));
+        assert_eq!(q.pop_front(), Some(1));
+
+        // Deque contains: ., ., 2, 3
+        q.push_back(4).unwrap();
+
+        // Deque contains: 4, ., 2, 3
+        assert_eq!(q.as_slices(), (&[2, 3][..], &[4][..]));
```
reitermarkus, line 1065: Can you use e.g. `[2, 3].as_slice()` in the tests? That may be a bit more readable.

reitermarkus (approved):

reitermarkus:
Thanks!

PR #451, opened 2024-02-07
Use macros to implement traits for `Vec` and `VecView`.

reitermarkus:
Reduce duplicated code by implementing traits for `Vec` and `VecView` using macros.

Also add `inline` to functions that redirect to `VecView`.


reitermarkus:
@rust-embedded/libs, please review.

newAM (approved):

PR #453, opened 2024-02-08
VecView: Implement serialize

sosthene-nitrokey:
I kept the initial implementation on `Vec`, but is there any reason why it doesn't use the implementation of `Serialize` for `[T]`?

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -2,7 +2,7 @@ use core::hash::{BuildHasher, Hash};
 
 use crate::{
     binary_heap::Kind as BinaryHeapKind, BinaryHeap, Deque, IndexMap, IndexSet, LinearMap, String,
-    Vec,
+    Vec, VecView,
 };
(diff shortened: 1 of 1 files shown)


sosthene-nitrokey:
This probably does not need a changelog entry since `VecView` has not yet been released.

sosthene-nitrokey:
The error seems to be caused by the Rust 1.76 release changing the error message. Maybe the Rust version should be pinned for CI.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Extend` impls for `Deque`.
 - Added `Deque::make_contiguous`.
 - Added `VecView`, the `!Sized` version of `Vec`.
+- Implement `Serialize` for `VecView`.
```
reitermarkus, line 20:
```suggestion
```

Dirbaio:
done in #486 1a08c7ac27634e5d405acaa2f35bd31e746c96da

PR #455, opened 2024-02-19
Fix CI.

reitermarkus:
- `FromIterator` is already in the prelude.
- Use `static_assertions` instead of `trybuild`.


Dirbaio (approved):
TIL `assert_not_impl_any`. seems much nicer than the cfail tests that keep breaking. Thanks!

Issue #270, opened 2022-02-14
Should examples be marked as no_run instead of ignore?


TDHolmes:
Coming out of #269 , I'm wondering if it makes sense to change all examples from `ignore` to just `no_run`. This would ensure they all still compile. I can take a look at this, but wanted to file a radar so I don't forget / *_if_* I forget at least I wrote it down 😅

japaric:
the examples marked as `ignore` depend on crates like `cortex-m`. having those target-specific crates as dev-dependencies could make the CI more complex (e.g. if we want to avoid building them on x86). I think it would be preferable to tweak the code to not depend on the `cortex-m` crates while still indicating "this function is the entry point", "this function is a signal/interrupt handler", etc.

PR #457, opened 2024-02-19
Don't ignore doc tests.

reitermarkus:
Closes https://github.com/rust-embedded/heapless/issues/270.


reitermarkus:
This doesn't actually do anything, since doc tests require `std`.

Issue #421, opened 2023-11-29
`heapless::pool` support for `x86_64`


ophilli:
I'm working on a no_std project targeting some armv7 CPUs. I would like to use `heapless::pool` but I am having issues because it is not available on my `x86_64` host at test / lint time.

https://github.com/rust-embedded/heapless/blob/c7c731d3587a46a35d7bc316434eee7bab9c4180/src/lib.rs#L119-L120

In [the docs](https://docs.rs/heapless/0.8.0/heapless/pool/index.html#target-support) there is a mention that it is available on ARM architectures which instruction set include the LDREX, CLREX and STREX instructions, and 32-bit x86, but it is unclear why x86_64 is no longer supported. It seems that x86_64 support was removed in https://github.com/rust-embedded/heapless/pull/315 last year.

What is limiting `pool` support such that it is available on `x86` but not `x86_64`?

newAM:
It's a good question.  I tried to compile it on `x86_64` with the `x86` implementation to see what happens and 6 tests fail with `signal: 11, SIGSEV: invalid memory reference`.

I assume to support `x86_64` we just need a backend implementation for the [Treiber stack](https://en.wikipedia.org/wiki/Treiber_stack).  This isn't something I have the time to investigate furthur at the moment sadly.

PR #458, opened 2024-02-19
Implement `pool` for any 32/64-bit architecture that supports the corresponding atomics.

reitermarkus:
Closes https://github.com/rust-embedded/heapless/issues/421.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Extend` impls for `Deque`.
 - Added `Deque::make_contiguous`.
 - Added `VecView`, the `!Sized` version of `Vec`.
+- Added pool implementations for 64-bit architectures.
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,8 @@ defmt-03 = ["dep:defmt"]
 # Enable larger MPMC sizes.
 mpmc_large = []
 
+nightly = []
+
 [dependencies]
 portable-atomic = { version = "1.0", optional = true }
 hash32 = "0.3.0"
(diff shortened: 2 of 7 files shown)


Dirbaio:
do we know why the current code is impl'd only for x86? It could be because the treiber stack is only sound in x86 thanks to its stronger ordering guarantees. Or it could be because `cfg(target_has_atomic)` didn't exist back then and the original author thought "meh x86 is the only major 32bit arch with 64bit atomics so let's just hardcode it".

Would be good to know before we lift the restriction. I'm not familiar at all with these fancier atomics stuff.

jamesmunns:
> do we know why the current code is impl'd only for x86? It could be because the treiber stack is only sound in x86 thanks to its stronger ordering guarantees. Or it could be because `cfg(target_has_atomic)` didn't exist back then and the original author thought "meh x86 is the only major 32bit arch with 64bit atomics so let's just hardcode it".
> 
> Would be good to know before we lift the restriction. I'm not familiar at all with these fancier atomics stuff.

I think the trick here is that the trieber stack is susceptible to the ABA problem: https://en.wikipedia.org/wiki/Treiber_stack#Correctness

On platforms with TRUE LL/SC atomics (e.g. Arm), this can be implemented correctly (tho might require custom asm? Unsure the state of things today, but see https://github.com/rust-embedded/heapless/issues/180 and changes made in https://github.com/rust-embedded/heapless/pull/315).

However it looks like this trick (again, I'm out of the loop) is to allow this in cases where the platform supports atomics that are LARGER than pointer widths, so you can CAS a pointer AND a unique tag to avoid the ABA problem, or at least make it sufficiently rare.

Dirbaio (approved):
reading more about it, I think i'm somewhat comfortable with the x64 impl being sound if it uses AtomicU128. Shame it's nightly-only but it's better than nothing :)

reitermarkus:
> I think the trick here is that the trieber stack is susceptible to the ABA problem: > https://en.wikipedia.org/wiki/Treiber_stack#Correctness

Thanks for the link, I understand now what the problem is and how using a value twice the size of the pointer width fixes it.

I have added an example explaining how it works, which is hopefully easy to follow.

I guess technically it's still possible to be incorrect if one threads pops/pushes the top of the stack `usize::MAX` times between another's `load` and `cas`, but that may be practically impossible.

> do we know why the current code is impl'd only for x86?

I would expect because `AtomicI128` is still unstable, i.e. this only works on `nightly` or when using `portable-atomic`.

Dirbaio (dismissed):

Dirbaio (approved):

PR #459, opened 2024-02-26
Add DequeView, similar to VecView and StringView

sosthene-nitrokey:
Ideally we would be able to merge the iterators of `&(mut) DequeView` and `&(mut) Deque`, but this can't be done without a breaking change.


sosthene-nitrokey:
Superseded by #490

PR #460, opened 2024-02-28
Add `uDisplay` for `String` and inline `ufmt` functions

lmbollen:
Most likely the compiler already inlines the function calls, but afaik adding the annotation makes it more likely.

I could not build it locally due to:
```
   Compiling heapless v0.8.0 (/home/lucas/repos/heapless)
error[E0658]: use of unstable library feature 'build_hasher_simple_hash_one'
    --> src/indexmap.rs:1260:28
     |
1260 |     HashValue(build_hasher.hash_one(key) as u16)
     |                            ^^^^^^^^
     |
     = note: see issue #86161 <https://github.com/rust-lang/rust/issues/86161> for more information

For more information about this error, try `rustc --explain E0658`.
```
But the implementation seems trivial and has been tested elsewhere.
I did not add `uDisplay` for `Vec` because the formatting is not trivial.

We might want to consider adding `uDebug`, but don't know what the rusty way to do so would be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implemented `PartialEq` and `Eq` for `Deque`.
 - Added `truncate` to `IndexMap`.
 - Added `get_index` and `get_index_mut` to `IndexMap`.
+- Added `String::uDisplay`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - `ufmt-impl` is now `ufmt`
   - `cas` is removed, atomic polyfilling is now opt-in via the `portable-atomic` feature.
 - `Vec::as_mut_slice` is now a public method.
+- `ufmt` functions are annotated with `inline`.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,13 +47,13 @@ hash32 = "0.3.0"
 serde = { version = "1", optional = true, default-features = false }
 ufmt-write = { version = "0.1", optional = true }
 defmt = { version = "1.0.1", optional = true }
+ufmt = "0.2"
 
 # for the pool module
 [target.'cfg(any(target_arch = "arm", target_pointer_width = "32", target_pointer_width = "64"))'.dependencies]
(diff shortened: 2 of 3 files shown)


reitermarkus:
Apparently not yet: https://github.com/japaric/ufmt/issues/52

reitermarkus:
[`inline(always)` is almost never needed](https://std-dev-guide.rust-lang.org/policy/inline.html#what-about-inlinealways), so let's use only `inline`.

Dirbaio:
can you rebase and fix CI?

lmbollen:
> can you rebase and fix CI?

Done, also changed `#[inline(always)]` to `#[inline]`

Cargo.toml
```diff
@@ -42,11 +42,12 @@ mpmc_large = []
 nightly = []
 
 [dependencies]
-portable-atomic = { version = "1.0", optional = true }
+defmt = { version = ">=0.2.0,<0.4", optional = true }
```
Dirbaio, line 45: please don't autoformat Cargo.toml

CHANGELOG.md
```diff
@@ -74,6 +75,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - `ufmt-impl` is now `ufmt`
   - `cas` is removed, atomic polyfilling is now opt-in via the `portable-atomic` feature.
 - `Vec::as_mut_slice` is now a public method.
+- `ufmt` functions are annotated with `inline(always)`.
```
Dirbaio, line 78: this no longer applies

src/ufmt.rs
```diff
@@ -1,15 +1,28 @@
 use crate::{storage::Storage, string::String, vec::VecInner};
```
sosthene-nitrokey, line 1:
```suggestion
use crate::{storage::Storage, string::StringInner, vec::VecInner};
```

src/ufmt.rs
```diff
@@ -1,15 +1,28 @@
...
+use ufmt::uDisplay;
 use ufmt_write::uWrite;
 
+impl<const SIZE: usize> uDisplay for String<SIZE> {
```
sosthene-nitrokey, line 5:
```suggestion
impl<S: Storage> uDisplay for StringInner<S> {
```

sosthene-nitrokey (changes requested):
This should be generic over the `StringInner` and the storage used.

reitermarkus (approved):

PR #465, opened 2024-04-05
VecView: fix memory leak of drop

sosthene-nitrokey:
In the initial implementation, I implemented `Drop` for `Vec` but I implemented it as a `no-op` for `VecView`, thinking that it should only be dropped through `Vec`.

This is incorrect, `VecView` can be responsible for dropping its content when its owned, for example in a `Box`.

This PR adds a test that fails with the current implementation, showing the memory leak, and fixes it by implementing `Drop` properly for `VecView`

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -15,12 +15,6 @@ pub trait VecDrop {
 }
 
 impl<T> VecDrop for [MaybeUninit<T>] {
-    unsafe fn drop_with_len(&mut self, _len: usize) {
-        // Case of a view, drop does nothing
-    }
-}
-
-impl<T, const N: usize> VecDrop for [MaybeUninit<T>; N] {
     unsafe fn drop_with_len(&mut self, len: usize) {
         // NOTE(unsafe) avoid bound checks in the slicing operation
         // &mut buffer[..len]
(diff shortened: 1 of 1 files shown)


src/vec.rs
```diff
@@ -5,34 +5,37 @@ use core::{
...
 };
 
-/// Workaround forbidden specialization of Drop
-pub trait VecDrop {
-    // SAFETY: drop_with_len will be called to call drop in place the first `len` elements of the buffer.
-    // Only the Owned buffer (`[MaybeUninit<T>; N]`) must drop the items
-    // and the view (`[MaybeUninit<T>]`) drops nothing.
-    // `drop_with_len `assumes that the buffer can contain `len` elements.
-    unsafe fn drop_with_len(&mut self, len: usize);
+pub trait VecBuffer {
```
reitermarkus, line 8: This trait should be sealed.

sosthene-nitrokey: It is sealed because `vec.rs` is already a private module. `Vec` and `VecView` are `pub use vec::{Vec, Vecview}`.

I can add one more layer of sealing in case `vec.rs` is made a public module.

sosthene-nitrokey: I'm also not sure how to get the workaround https://github.com/rust-lang/rust/issues/119015 to work with 2 layers sealing, but it's also no necessary anymore on stable 1.77.0, so we might as well remove it.

reitermarkus: Okay, fine as is then.

reitermarkus (approved):

PR #473, opened 2024-05-19
 const fn new_with_len<const LEN: usize>() to heapless::Vec

Bruce0203:


reitermarkus:
This is unsound. What is the use case for this?

burrbull:
Agree. This is totally wrong.

PR #479, opened 2024-06-26
LinearMap: add `LinearMapView`, similar to `VecView`

sosthene-nitrokey:
While implementing I realized that `&mut LinearMap` does not implement `IntoIter`. Should I add it ?  

I currently did not add it to `LinearMapView` either for symmetry, but that can lead to confusion. For example, the following doesn't compile:

```rust
use heapless::{LinearMap, LinearMapView};

let mut map: LinearMap<_, _, 8> = LinearMap::new();
let map_view: &mut LinearMapView<_, _> = &mut map;

for (key, val) in map_view {
    println!("key: {} val: {}", key, val);
}
```

The fix is to use instead 

```rust
for (key, val) in &*map_view {
```


sosthene-nitrokey:
replaced by #491

PR #480, opened 2024-06-27
`HistoryBuffer`: add `HistoryBufferView`, similar to `VecView`

sosthene-nitrokey:
The `view` uses `iter::Chain<Slice::Iter, Slice::Iter>` as the iterator. This is conceptually simpler, and would also allow exposing the other guarantees of the iterator (Reverse, TrustedLen, Fused) if we want to. This also means the iterator does not have the `const N: usize` in its type.

We could change `HistoryBuffer::oldest_ordered` to use the same iterator, but that would be a breaking change.


sosthene-nitrokey:
replaced by #493

PR #481, opened 2024-06-27
BinaryHeap: add `BinaryHeapView`, similar to `VecView`

sosthene-nitrokey:
I left the `PeekMut` implementation as_is and duplicated it into `PeekMutView`, since it would be a breaking change to remove the `N` const generic on it. But I think it should be removed with the next breaking change release.


sosthene-nitrokey:
Replaced by #495

PR #483, opened 2024-06-27
Fix memory leak of `MpMcQueue`

### LINKED ISSUE #482: mpmc: dropping a queue does not drop its content, which can cause a memory leak
Implementing `Drop` is technically a breaking change.

sosthene-nitrokey:
Fix #482

### DIFF
--- a/src/mpmc.rs
+++ b/src/mpmc.rs
@@ -194,6 +194,13 @@ impl<T, const N: usize> Default for MpMcQueue<T, N> {
     }
 }
 
+impl<T, const N: usize> Drop for MpMcQueue<T, N> {
+    fn drop(&mut self) {
+        // drop all contents currently in the queue
+        while self.dequeue().is_some() {}
+    }
+}
+
 unsafe impl<T, const N: usize> Sync for MpMcQueue<T, N> where T: Send {}
 
 struct Cell<T> {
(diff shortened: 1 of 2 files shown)


src/mpmc.rs
```diff
@@ -194,6 +194,13 @@ impl<T, const N: usize> Default for MpMcQueue<T, N> {
     }
 }
 
+impl<T, const N: usize> Drop for MpMcQueue<T, N> {
+    fn drop(&mut self) {
+        // drop all contents currently in the queue
+        while self.dequeue().is_some() {}
```
Dirbaio, line 200: This is copying the elements to the stack before dropping them, and running the complex dequeue() logic. 

it'd be more efficient to read enqueue_pos/dequeue_pos, iterate, then .assume_init_drop() all the items in place. No copies, no atomic ops. This is sound because on drop we have `&mut self` so we know no other thread can be concurrently accessing the queue at this point.

sosthene-nitrokey: I would like to implement that but I'm not 100% certain I understand the layout of the Queue which is not documented, and which items are actually valid. I'm certain that the `dequeue` approach is  correct so this is the approach I took.

Dirbaio: ah yes, it's true it's not that trivial. We can optimize later if someone complains.

Dirbaio (approved):
thank you!

src/mpmc.rs
sosthene-nitrokey: Let's open an Issue for that

PR #485, opened 2024-06-27
`spsc::Queue`: add `spsc::QueueView`, similar to `VecView`

sosthene-nitrokey:
Similarly to https://github.com/rust-embedded/heapless/pull/480,  there are structs which keep the `const` generic `N` for backwards compatibility (`Iter`, `IterMut`, `Producer`, `Consumer`).


sosthene-nitrokey:
Replaced by #496

PR #452, opened 2024-02-08
String: add `StringView`, similar to `VecView`

sosthene-nitrokey:
Follow up of #425 

This time we need a dedicated `sealed` module for `StringInner`, because of the `pub mod string` in `lib.rs`. Maybe `VecInner` should have the same treatment, in case some time in the future the `vec` module is made public.

Issue #353, opened 2023-03-02
Feature: Support SliceVec, backed by a &mut [MaybeUninit<T>] instead of [MaybeUninit<T>; N]


kupiakos:
So, a dynamically sized `Vec`, but still with a max capacity. This would remain heapless, as you can construct the slice on the stack and pass it in. There are some particular advantages to this approach in an embedded environment, the biggest of which being the lack of monomorphization required to support multiple capacities. A `SliceVec<'a, T>` can be produced by borrowing a `Vec<T, N>`, integrating into the library well.

When used like this, it can be directly used to keep track of how many bytes of an uninit buffer have been written to and provide a safe method to retrieve the initialized bytes. This is particularly useful for zero-copy deserialization.

If I were send a PR to add this, would it be accepted? This is a blocker for my team switching to `heapless` off of [`FixedVec`](https://docs.rs/fixedvec/latest/fixedvec/struct.FixedVec.html), which has less support and fewer types to work with.

kupiakos:
So, after some deliberation, my initial proposal isn't quite right. You cannot simply produce a `SliceVec<'a, T>` by borrowing on a `&'a mut Vec<T, N>`, since the length isn't borrowed. I'm considering this shape of function:

```rust
/// Read up to `N` bytes from some response into `out`.
fn read_from_response<const N: usize>(&mut self, out: &mut heapless::Vec<u8, N>) -> Result<(), Err>;
```

We want this to be able to work safely with different `N`, but avoid monomorphization for different `N`. We can do that with an inner function and the right inlining, but it'd be much simpler for us to express this as a type.

However, I also want to be able to work with an externally provided buffer, as parts of our system design require it. Right now we're just `unsafe` and I'm trying to reduce our amount of unsafety with the right primitives. It turns out safely filling in an uninitialized buffer is perfectly represented by a `Vec` that borrows from an `&'a mut [MaybeUninit<T>]` instead of storing its own array inline.

So, I want a `Vec` that borrows from an external buffer, but I also want one that can be easily borrowed from a `Vec<T, N>`. These are at odds.

So, the best design I have is two separate types, and I'd like to hear your thoughts. Both avoid monomorphization on `N`. I have a basic demo playground [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=25826a61beecfa38e65f95d8ebe2decc).

[...]

PR #486, opened 2024-06-27
vec: remove code duplication due to VecView.

Dirbaio:
I'm a bit concerned about the code duplication introduced by #425. Most `Vec`
methods forward to their `VecView` counterparts, but the function signature
and the doc comments are still duplicated. This means every time we do a doc
fix or add a new method we have to remember to do it on both sides.

It has worked fine for for experimenting and proving the concept. However,
now we're at a stage where it's proven and we have to commit to
add it to all the other containers: #452 #459 #479 #480 #481 #485.
Each of these PRs adds 400-600 lines to the codebase, mostly duplicated
signatures and docs.

IMO we should find a way of adding `View` types without so much duplication
first, *then* add it to all containers.

This PR implements a possible solution I've found that removes all the duplication:

- Added a `Storage` trait with two impls, for sized and unsized.
- Vec is now generic over `Storage` instead of over the buffer type. The buffer type is now a `Buffer` associated type.
- `Storage` is sealed. The `Buffer` associated type is private. The user cannot learn the `Buffer` type for the 
  `Storage` impls, they can just use it with `VecInner` and it makes it behave differently magically.
- Most methods are implemented for `VecInner<S>` for any `S: Storage`. This makes them
  available on both `Vec` and `VecView` without duplicating anything.

Advantages:
- No duplication.
- No macros.
- Trait impls are generic over `Storage` now, so you can e.g. compare a `Vec` with a `VecView`. Before you couldn't, you could only do `Vec-Vec` or `VecView-VecView`.
- Users can now write code or trait impls generic over the storage.

Docs work fine on latest stable, it doesn't seem to need any workarounds:
- docs for `Vec` show the `Vec`-only methods and the storage-independent methods.
- docs for `VecView` show the `VecView`-only methods (if we had some, we don't) and the storage-independent methods.
- docs for `Vecinner` show all methods.

I also propose making `mod vec` public, including `Storage, VecInner`. The reasons are:
- `vec::IntoIter` was already in the public API, but was previously unnameable.
- Exposing `Storage, VecInner` can be useful for users that want to write code or trait impls generic over the storage.
- Now that `Storage` is a sealed trait and the `Buffer` type is hidden, it's not leaking implementation details anymore. There's no `[MaybeUninit<T>]` in the public API, we can change it at any time without it being a breaking change.

I think we also should remove the `Vec, VecView` reexports in the crate root, but this is left for a future PR.

cc @sosthene-nitrokey

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -13,7 +13,15 @@ pub(crate) trait SealedStorage {
 /// - [`OwnedStorage`]: stores the data in an array `[T; N]` whose size is known at compile time.
 /// - [`ViewStorage`]: stores the data in an unsized `[T]`.
 ///
-/// This allows containers to be generic over either sized or unsized storage.
+/// This allows containers to be generic over either sized or unsized storage. For example,
+/// the [`vec`](crate::vec) module contains a [`VecInner`](crate::vec::VecInner) struct
+/// that's generic on [`Storage`], and two type aliases for convenience:
+///
+/// - [`Vec<T, N>`](crate::vec::Vec) = `VecInner<T, OwnedStorage<N>>`
+/// - [`VecView<T>`](crate::vec::VecView) = `VecInner<T, ViewStorage>`
+///
+/// `Vec` can be unsized into `VecView`, either by unsizing coercions such as `&mut Vec -> &mut VecView` or
+/// `Box<Vec> -> Box<VecView>`, or explicitly with [`.as_view()`](crate::vec::Vec::as_view) or [`.as_mut_view()`](crate::vec::Vec::as_mut_view).
 ///
 /// This trait is sealed, so you cannot implement it for your own types. You can only use
 /// the implementations provided by this crate.
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,0 +1,35 @@
+//! `Storage` trait defining how data is stored in a container.
+
+use core::borrow::{Borrow, BorrowMut};
+
+pub(crate) trait SealedStorage {
+    type Buffer<T>: ?Sized + Borrow<[T]> + BorrowMut<[T]>;
+}
+
+/// Trait defining how data for a container is stored.
+///
+/// There's two implementations available:
+///
+/// - [`OwnedStorage`]: stores the data in an array `[T; N]` whose size is known at compile time.
+/// - [`ViewStorage`]: stores the data in an unsized `[T]`.
+///
+/// This allows containers to be generic over either sized or unsized storage.
+///
+/// This trait is sealed, so you cannot implement it for your own types. You can only use
+/// the implementations provided by this crate.
+#[allow(private_bounds)]
+pub trait Storage: SealedStorage {}
+
+/// Implementation of [`Storage`] that stores the data in an array `[T; N]` whose size is known at compile time.
+pub enum OwnedStorage<const N: usize> {}
+impl<const N: usize> Storage for OwnedStorage<N> {}
+impl<const N: usize> SealedStorage for OwnedStorage<N> {
+    type Buffer<T> = [T; N];
+}
+
+/// Implementation of [`Storage`] that stores the data in an unsized `[T]`.
+pub enum ViewStorage {}
+impl Storage for ViewStorage {}
+impl SealedStorage for ViewStorage {
+    type Buffer<T> = [T];
+}
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -341,17 +341,27 @@ impl<T, S: Storage> VecInner<T, S> {
     where
         T: Clone,
     {
-        if self.len + other.len() > self.capacity() {
-            // won't fit in the `Vec`; don't modify anything and return an error
-            Err(())
-        } else {
-            for elem in other {
-                unsafe {
-                    self.push_unchecked(elem.clone());
+        pub fn extend_from_slice_inner<T>(
+            len: &mut usize,
+            buf: &mut [MaybeUninit<T>],
+            other: &[T],
+        ) -> Result<(), ()>
+        where
+            T: Clone,
+        {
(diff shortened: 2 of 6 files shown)


Dirbaio:
Tested code size on company firmware:

```
   text    data     bss     dec     hex filename
 378136    1796  182436  562368   894c0 heapless git v0.8.0
 378168    1796  182436  562400   894e0 heapless git main
 378232    1796  182436  562464   89520 heapless git vecview-dedup
```

slight 0.02% bloat, but imo it's within the margin of error.

sosthene-nitrokey:
I thought about this approach initially but didn't go for it because:

- had concerns about the documentation generated, but there were bugs with Rust doc that have been fixed since then, so it may be actually better now
- The initial reasoning behind this PR was to reduce binary size by reducing the need for monomorphization for each value `N`. I'll re-run the comparison again but if I remember correctly this lead to 2% size reduction in our firmware, i'm testing this again.

> - Users can now write code or trait impls generic over the storage.

I'm not  sure I see the benefits of that. My idea was that we would end up using `VecView` pretty much everywhere we can, and when we can't it would be because it's `!Sized`, and thus not 

On the other hand, I agree with the benefits of this approach:

- Less code (thus more maintainable)
- The `Storage` trait could also be implemented for a `VecSlice<'a>` type, that would be backed by a `&'a [MaybeUninit<T>]`, thus providing also the implementation for https://github.com/rust-embedded/heapless/issues/353

sosthene-nitrokey:
Ok, on our firmware (we have 4 versions) we see the following sizes (in bytes):

# Current main

```
1397551
2007561
501460
576036
```

# Heapless 0.8

```
1398803
2014626
502048
576516
```

# This PR

```
1397543
2010261
501312
575940
```

The biggest difference we can see is from heapless 0.8 to heapless main, with a difference of 0.3% smaller firmware size. The difference is lower that what I measured previously, but I'm not sure why (I tested with heapless 0.7 at the time).

src/vec.rs
```diff
@@ -1,43 +1,53 @@
...
-        vec
-    }
-    fn as_mut_vecview(vec: &mut VecInner<Self>) -> &mut VecView<Self::T> {
-        vec
-    }
+/// Implementation of [`Storage`] for [`VecView`].
+pub enum VecViewStorage {}
+impl Storage for VecViewStorage {}
+impl SealedStorage for VecViewStorage {
+    type Buffer<T> = [MaybeUninit<T>];
```
sosthene-nitrokey, line 41: I think It would make sense to for these structs be moved in a common module for the crate, and rename the structs to `Owned` and `View`, so that the same structs can be used for other containers ?

I think it's clearer to use the same structs for all containers.

sosthene-nitrokey:
I would instead use:

```rust
trait SealedStorage {
    type Buffer<T>: ?Sized + Borrow<[T]> + BorrowMut<[T]>;
}
```

And have `VecInner` be defined to be:

```rust
pub struct VecInner<T, S: Storage> 
    len: usize,
    buffer: S::Buffer<MaybeUninit<T>>,
}
```

That would make the `Storage` more generic, and make it usable for all container structures. For example `mpmc` would be defined as:

```rust
pub struct MpMcQueueInner<S: Storage> {
    pub(super) dequeue_pos: AtomicTargetSize,
    pub(super) enqueue_pos: AtomicTargetSize,
    pub(super) buffer: UnsafeCell<S::Buffer<Cell<T>>>,
}
 ```
 
 Which can't be done if the `Buffer` always introduces a `MaybeUninit`.

File: src/vec.rs (hunk 1 above)
Comment (Dirbaio) [reply src/vec.rs]: good point, done!

**Comment by Dirbaio** [comment]
> I'm not sure I see the benefits of that. My idea was that we would end up using VecView pretty much everywhere we can, and when we can't it would be because it's !Sized, and thus not

yeah it's a rather weak advantage, but I can't see why *not* allow it.

> I would instead use: ... `type Buffer<T>: ?Sized + Borrow<[T]> + BorrowMut<[T]>`

very true, not all containers use MaybeUninit. Changed!

File: src/storage.rs [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+///
+/// - [`OwnedStorage`]: stores the data in an array `[T; N]` whose size is known at compile time.
+/// - [`ViewStorage`]: stores the data in an unsized `[T]`.
+///
+/// This allows containers to be generic over either sized or unsized storage.
+///
+/// This trait is sealed, so you cannot implement it for your own types. You can only use
+/// the implementations provided by this crate.
+#[allow(private_bounds)]
+pub trait Storage: SealedStorage {}
```
sosthene-nitrokey, line 21: I would add an intra-doc link to the documentation of `VecView` and `Vec::as_view` to explain what these traits are used for.

sosthene-nitrokey (approved):
That is indeed much cleaner for maintenance than the current duplicated documentation.
Since the size benefits are minor, I'm all for this PR.

I just wish it were possible to have the examples in the documentation of `VecView` to actually use `VecView` and not `Vec` but well, we can live with that.

sosthene-nitrokey (changes requested):
The implementation for `Serialize` for `Vec` should be over `VecInner` and generic over the `Storage`

sosthene-nitrokey:
Using `Storage` for other implementations becomes so efficient with this !

For example `String`: https://github.com/rust-embedded/heapless/pull/487/commits/c053490c9836fcf1d6d060ad1011fa533ffc31c7

src/storage.rs
Dirbaio: done!

Dirbaio:
> The implementation for Serialize for Vec should be over VecInner and generic over the Storage

done. I was also missing defmt, ufmt, done that too.

newAM (approved):

PR #484, opened 2024-06-27
MpMcQueue: add `MpMcQueueView`, similar to `VecView`

sosthene-nitrokey:
This PR is built on top of #483

PR #488, opened 2024-06-28
 `MpMcQueue`: add `MpMcQueueView`, similar to `VecView` on top of #486

sosthene-nitrokey:
This would supersede #484. We need to pay attention to the `Drop` impl depending on the order we merge PRs

This adds a couple of methods to the `Storage` sealed trait to deal with the fact that `Borrow` cannot be used in concurrent code:

- One to safely get the length of the buffer from a pointer to it
- One to get a pointer to the beginning of the slice

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -4,6 +4,10 @@ use core::borrow::{Borrow, BorrowMut};
 
 pub(crate) trait SealedStorage {
     type Buffer<T>: ?Sized + Borrow<[T]> + BorrowMut<[T]>;
+    /// Obtain the length of the buffer
+    fn len<T>(this: *const Self::Buffer<T>) -> usize;
+    /// Obtain access to the first element of the buffer
+    fn as_ptr<T>(this: *mut Self::Buffer<T>) -> *mut T;
 }
 
 /// Trait defining how data for a container is stored.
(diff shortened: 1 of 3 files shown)


src/storage.rs
```diff
@@ -33,11 +39,30 @@ pub enum OwnedStorage<const N: usize> {}
...
+    }
 }
 
 /// Implementation of [`Storage`] that stores the data in an unsized `[T]`.
 pub enum ViewStorage {}
 impl Storage for ViewStorage {}
 impl SealedStorage for ViewStorage {
     type Buffer<T> = [T];
+    fn len<T>(this: *const Self::Buffer<T>) -> usize {
+        // We get the len of the buffer. There is no reverse method of `core::ptr::from_raw_parts`, so we have to work around it.
```
Dirbaio, line 56: Rust 1.79 has stabilized `[T]::len()`, so you can do `this.len()`.

sosthene-nitrokey: Oh, that's good!

sosthene-nitrokey: Done in 22efa13bf24aafe5bba28751443fca7454e2069d

Dirbaio (approved):
thanks!

PR #491, opened 2024-07-01
LinearMap: add LinearMapView, similar to VecView on top of #486

sosthene-nitrokey:


Dirbaio:
can you update the serialize impl too?

Dirbaio (approved):
thanks!

Dirbaio (approved):

PR #493, opened 2024-07-01
Implement HistoryBufferView on top of #486

sosthene-nitrokey:
`OldestOrdered` was already part of the public API.

When merging we will need to add its removal to #494 so that there is only `OldestOrederedView` (which can be renamed to just `OldestOrdered`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `StringView`, the `!Sized` version of `String`.
 - Added `MpMcQueueView`, the `!Sized` version of `MpMcQueue`.
 - Added `LinearMapView`, the `!Sized` version of `LinearMap`.
+- Added `HistoryBufferView`, the `!Sized` version of `HistoryBuffer`.
 
 ### Changed
 
--- a/src/histbuf.rs
+++ b/src/histbuf.rs
@@ -1,8 +1,58 @@
+//! A "history buffer", similar to a write-only ring buffer of fixed length.
+//!
+//! This buffer keeps a fixed number of elements.  On write, the oldest element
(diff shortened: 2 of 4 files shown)


Dirbaio:
> OldestOrdered was already part of the public API.

oof :smiling_face_with_tear: 

maybe it's "less bad" to make `OldestOrdered` the "new" way already, and add a type alias at the root for compat that just ignores the const generic. This way in 0.9 we can just remove the type alias.

```rust
mod histbuf {
    pub struct OldestOrdered<'a, T> {
        inner: core::iter::Chain<core::slice::Iter<'a, T>, core::slice::Iter<'a, T>>,
    }
}

#[deprecated = "use `heapless::histbuf::OldestOrdered` instead."]
pub type OldestOrdered<'a, T, const N: usize> = OldestOrderedInner<'a, T>;
```

i'm not sure if there's any problem in "ignoring" `N` that way, it seems to work in quick testing.

sosthene-nitrokey:
> i'm not sure if there's any problem in "ignoring" N that way, it seems to work in quick testing.

I did not know that was possible!

sosthene-nitrokey:
> i'm not sure if there's any problem in "ignoring" N that way, it seems to work in quick testing.

I think I found a possible issue, there will be some cases where the compiler won't be able to infer `N`. For exemple the following compiles now, but wouldn't compile with the new alias approach.

```rust
fn testing<const N: usize>(iter: OldestOrdered<'_, u32, N> {}

fn main() {
   let a = HistoryBuffer::<u32, 3>::new();
   testing(a.oldest_ordered);
}
```

Dirbaio:
aw, that's a shame :(. Nice find. Let's do the PhantomData then.

Dirbaio (approved):

PR #498, opened 2024-07-02
Keep `Vec::capacity` const

sosthene-nitrokey:
In 0.8.0 it was const, but this was removed in https://github.com/rust-embedded/heapless/pull/486, which would be an unexpected breaking change.

The other container types did not make the `capacity` method const, and therefore can kept with the normal name and the generic implementation.

No need for changelog for this change, since it affects b previous PR that has not been released.

### DIFF
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -291,6 +291,13 @@ impl<T, const N: usize> Vec<T, N> {
     {
         self.as_mut_view().drain(range)
     }
+
+    /// Returns the maximum number of elements the vector can hold.
+    ///
+    /// This method is not available on a `VecView`, use [`storage_len`](VecInner::storage_capacity) instead
+    pub const fn capacity(&self) -> usize {
+        self.buffer.len()
+    }
 }
 
 impl<T> VecView<T> {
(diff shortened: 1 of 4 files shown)


Dirbaio (approved):
oh oops, nice find! thanks

src/vec/mod.rs
```diff
@@ -408,7 +415,7 @@ impl<T, S: Storage> VecInner<T, S> {
     }
 
     /// Returns the maximum number of elements the vector can hold.
-    pub fn capacity(&self) -> usize {
+    pub fn storage_capacity(&self) -> usize {
```
reitermarkus, line 418: @sosthene-nitrokey, why can't this use the same name?

sosthene-nitrokey: > so for Vec<T, N> both apply, so it'd give ambiguity errors if the user wrote vec.capacity()

It wouldn't even allow  getting to that point. The compiler would complain about duplicate definitions with the same name.

We could implement `fn capacity` on `Vec` and `VecView` independently, but then it would make it unusable in a generic context over the storage.

PR #500, opened 2024-07-03
Implement `Vec::drain`, `as_(mut_)view` on the `*Inner` types, generic over `Storage`

sosthene-nitrokey:
This makes `drain` usable in the context of an implementation generic over the storage
This also makes it possible to get a `*View` in this same context

This could also be a way to "de-monomorphize" all the implementations of the `Inner` structs:

We could make all implementation define an inner function that is over a `View` type, and start the
implementation with `this = self.as_(mut_)view`, maybe for binary size and compilation time benefits. For example the implementation of `VecInner::push` could be:

```rust
    /// Appends an `item` to the back of the collection
    ///
    /// Returns back the `item` if the vector is full.
    pub fn push(&mut self, item: T) -> Result<(), T> {
        fn inner<T>(this: &mut VecView<T>, item: T) -> Result<(), T> {
            if this.len < this.storage_capacity() {
                unsafe { this.push_unchecked(item) }
                Ok(())
            } else {
                Err(item)
            }
        }
        inner(self.as_mut_view(), item)
    }
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,6 +181,11 @@ pub mod storage;
 pub mod string;
 pub mod vec;
 
+// Workaround a compiler ICE in rust 1.83 to 1.86
+// https://github.com/rust-lang/rust/issues/138979#issuecomment-2760839948
+#[expect(dead_code)]
+fn dead_code_ice_workaround() {}
+
 #[cfg(feature = "serde")]
 mod de;
 #[cfg(feature = "serde")]
--- a/src/binary_heap.rs
+++ b/src/binary_heap.rs
@@ -184,14 +184,16 @@ impl<T, K, const N: usize> BinaryHeap<T, K, N> {
     pub fn into_vec(self) -> Vec<T, N> {
         self.data
     }
+}
 
+impl<T, K, S: VecStorage<T>> BinaryHeapInner<T, K, S> {
     /// Get a reference to the `BinaryHeap`, erasing the `N` const-generic.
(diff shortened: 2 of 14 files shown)


sosthene-nitrokey:
I checked with nightly, we can't do that with just a `type Buffer<T>: Unsize<[T]`, because there is no implementation of `Unsize<[T]>` for `[T]`, that's a shame. So even if `Unsize` gets stabilized, we'll have to keep all the manual conversion methods.

sosthene-nitrokey:
To get it working with `LinearMap` and `String` (which have constraints on the the inner type of the Vec), I had to create new traits for them. I wonder if that would not be beneficial for the other types too.

sosthene-nitrokey:
Ok, I have a workaround for the compiler ICE, that doesn't appear to have any downsides. Added it and documented it.

src/lib.rs
```diff
@@ -181,6 +181,11 @@ pub mod storage;
 pub mod string;
 pub mod vec;
 
+// Workaround a compiler ICE in rust 1.83 to 1.86
```
reitermarkus, line 184:
```suggestion
// FIXME: Workaround a compiler ICE in rust 1.83 to 1.86
```

reitermarkus: Can be done in a follow-up PR. Let's get this in before it conflicts with anything else.

reitermarkus (approved):

reitermarkus:
Thanks, @sosthene-nitrokey!

PR #502, opened 2024-07-04
Re-introduce covariance for containers

### LINKED ISSUE #501: The deduped versions of the `*View` types break variance:
The following compiles with `v0.8.0` and https://github.com/rust-embedded/heapless/pull/425, but not with latest main:

```rust
fn test_variance<'a: 'b, 'b>(x: Vec<&'a (), 42>) -> Vec<&'b (), 42> {
    x
}
```

@Dirbaio 

I think this is caused the intermediary `Generic Associated Type`, but I could not find where this is documented.

sosthene-nitrokey:
`MpMc` and `spsc::Queue` are not migrated to the new pattern because they were already invariant due to the `UnsafeCell`.

Fix #501

### DIFF
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -1,8 +1,8 @@
 //! A fixed capacity [`Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html).
 
 use core::borrow;
+use core::marker::PhantomData;
 use core::{
-    borrow::{Borrow, BorrowMut},
     cmp::Ordering,
     fmt, hash,
     mem::{self, ManuallyDrop, MaybeUninit},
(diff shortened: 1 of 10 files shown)


sosthene-nitrokey:
Rebased. CI is failing only because of CHANGELOG.

newAM (approved):
Thanks!

This looks logical, but I'm not confident enough to merge this on my review alone, will wait for another review from @rust-embedded/libs to review.

src/vec/mod.rs
```diff
@@ -11,18 +11,88 @@ use core::{
...
 pub use drain::Drain;
 
-/// Base struct for [`Vec`] and [`VecView`], generic over the [`Storage`].
+/// Base struct for [`Vec`] and [`VecView`], generic over the [`VecStorage`].
 ///
 /// In most cases you should use [`Vec`] or [`VecView`] directly. Only use this
 /// struct if you want to write code that's generic over both.
-pub struct VecInner<T, S: Storage> {
+pub struct VecInner<T, S: VecStorage<T> + ?Sized> {
+    // This phantomdata is required because otherwise rustc thinks that `T` is not used
```
jannic, line 92: Minor nit: Is this comment helpful? After all, that's usually the reason to use PhantomData.

sosthene-nitrokey: Indeed, it was not obvious to me initially that it would be required because `T` is used in the generic bound, but the comment does not do anything to explain why it's needed.

I removed it. Thanks.

jannic:
It doesn't mean much because I'm not a member of the libs team, and also don't know much about the internals of heapless. 

But I was curious and spent some time to understand both the issue and the proposed solution. The key information not obvious from other parts of the documentation was https://rustc-dev-guide.rust-lang.org/variance.html#variance-and-associated-types. After I found that, I'm convinced your proposal is a good solution. It's unfortunate that it complicates the code a bit, but it's probably worth it.

I tried to measure the effect on binary sizes and compiled the embassy uart_r503 for heapless 0.8.0 and for this branch, and the version from this branch was actually 20 bytes smaller than with 0.8.0.
Notably, I was unable to compare it with the current main branch, because that didn't compile: It failed with some entirely unhelpful error message about lifetimes. :-) I guess this shows that the fix is indeed useful.

therealprof (approved):
Looking great to me.

PR #504, opened 2024-07-18
Implement a generic length parameter for Vec<T, N>

GnomedDev:
Currently, `Vec<T, N>` always uses `usize` for the length field, which leads to silly situations like `Vec<u8, 4>` being `8` bytes large on 32 bit platforms. This PR introduces a generic to provide this length field as well as a default for many common length values to avoid much user code changing.

I am okay to revert sections of this PR if wanted, such as
- ~~The change of `usize` to `LenT` in many arguments and return types, if they are considered too breaking.~~ Done, `usize` is still used for indexing and as the `len` return type.
- Adding the length generic to `VecView`, this can be changed to only store `usize` at the cost of only `Vec<T, N, usize>` being able to unsize (or requiring a `Vec::cast_len_type` call before unsizing, at least).

This lays the ground work for followup PRs to implement a similar length generic on the other containers, but I'm starting with just `Vec` to test the waters.

### DIFF
--- a/src/len_type.rs
+++ b/src/len_type.rs
@@ -3,34 +3,8 @@ use core::{
     ops::{Add, AddAssign, Sub, SubAssign},
 };
 
-mod private {
-    pub trait Sealed {}
-
-    impl Sealed for u8 {}
-    impl Sealed for u16 {}
-    #[cfg(any(target_pointer_width = "32", target_pointer_width = "64"))]
-    impl Sealed for u32 {}
-
-    impl Sealed for usize {}
-}
-
-macro_rules! impl_lentype {
-    ($($(#[$meta:meta])* $LenT:ty),*) => {$(
-        $(#[$meta])*
-        impl LenType for $LenT {
-            const ZERO: Self = 0;
-            const ONE: Self = 1;
-            const MAX: usize = Self::MAX as _;
-        }
-    )*}
-}
-
-/// A sealed trait representing a valid type to use as a length for a container.
-///
-/// This cannot be implemented in user code, and is restricted to `u8`, `u16`, `u32`, and `usize`.
-pub trait LenType:
-    private::Sealed
-    + Send
+pub trait Sealed:
+    Send
     + Sync
     + Copy
     + Display
@@ -61,6 +35,25 @@ pub trait LenType:
         self.try_into().unwrap()
     }
 }
+
+macro_rules! impl_lentype {
+    ($($(#[$meta:meta])* $LenT:ty),*) => {$(
+        $(#[$meta])*
+        impl Sealed for $LenT {
+            const ZERO: Self = 0;
+            const ONE: Self = 1;
+            const MAX: usize = Self::MAX as _;
+        }
+
+        $(#[$meta])*
+        impl LenType for $LenT {}
+    )*}
+}
+
+/// A sealed trait representing a valid type to use as a length for a container.
+///
+/// This cannot be implemented in user code, and is restricted to `u8`, `u16`, `u32`, and `usize`.
+pub trait LenType: Sealed {}
 
 impl_lentype!(
     u8,
--- a/src/len_type.rs
+++ b/src/len_type.rs
@@ -26,11 +26,13 @@ pub trait Sealed:
     const MAX: usize;
 
     /// An infallible conversion from `usize` to `LenT`.
+    #[inline]
     fn from_usize(val: usize) -> Self {
         val.try_into().unwrap()
     }
 
     /// An infallible conversion from `LenT` to `usize`.
+    #[inline]
     fn into_usize(self) -> usize {
         self.try_into().unwrap()
     }
--- a/src/len_type.rs
+++ b/src/len_type.rs
@@ -1,0 +1,112 @@
+use core::{
+    fmt::{Debug, Display},
+    ops::{Add, AddAssign, Sub, SubAssign},
+};
+
+mod private {
+    pub trait Sealed {}
+
+    impl Sealed for u8 {}
+    impl Sealed for u16 {}
+    #[cfg(any(target_pointer_width = "32", target_pointer_width = "64"))]
+    impl Sealed for u32 {}
+
+    impl Sealed for usize {}
+}
+
+macro_rules! impl_lentype {
+    ($($(#[$meta:meta])* $LenT:ty),*) => {$(
+        $(#[$meta])*
+        impl LenType for $LenT {
+            const ZERO: Self = 0;
+            const ONE: Self = 1;
+            const MAX: usize = Self::MAX as _;
+        }
+    )*}
+}
+
+/// A sealed trait representing a valid type to use as a length for a container.
+///
+/// This cannot be implemented in user code, and is restricted to `u8`, `u16`, `u32`, and `usize`.
+pub trait LenType:
+    private::Sealed
+    + Send
+    + Sync
+    + Copy
+    + Display
+    + Debug
+    + PartialEq
+    + Add<Output = Self>
+    + AddAssign
+    + Sub<Output = Self>
+    + SubAssign
+    + PartialOrd
+    + TryFrom<usize, Error: Debug>
+    + TryInto<usize, Error: Debug>
+{
+    /// The zero value of the integer type.
+    const ZERO: Self;
+    /// The one value of the integer type.
+    const ONE: Self;
+    /// The maximum value of this type, as a `usize`.
+    const MAX: usize;
+
+    /// An infallible conversion from `usize` to `LenT`.
+    fn from_usize(val: usize) -> Self {
+        val.try_into().unwrap()
+    }
+
+    /// An infallible conversion from `LenT` to `usize`.
+    fn into_usize(self) -> usize {
+        self.try_into().unwrap()
+    }
+}
+
+impl_lentype!(
+    u8,
+    u16,
+    #[cfg(any(target_pointer_width = "32", target_pointer_width = "64"))]
+    u32,
+    usize
+);
+
+macro_rules! impl_lentodefault {
+    ($LenT:ty: $($len:literal),*) => {$(
+        impl SmallestLenType for Const<$len> {
+            type Type = $LenT;
+        }
+    )*};
+}
+
+/// A struct to create individual types for mapping with [`SmallestLenType`].
+///
+/// See the documentation of [`DefaultLenType`] for a detailed explanation.
+pub struct Const<const N: usize>;
+
+/// A trait to map [`Const`] to it's respective [`LenType`].
+///
+/// See the documentation of [`DefaultLenType`] for a detailed explanation.
+#[diagnostic::on_unimplemented(
+    message = "Length `N` does not have a default `LenType` mapping",
+    note = "Provide the `LenType` explicitly, such as `usize`"
+)]
+pub trait SmallestLenType {
+    type Type: LenType;
+}
+
+/// A type alias to perform the `const N: usize` -> `LenType` mapping.
+///
+/// This is impossible to perform directly, but it is possible to write a `const N: usize` -> related `Type` mapping via a const generic argument,
+/// then map from that to an unrelated type via a trait with associated types.
+///
+/// [`Const`] is the "related type" in the above explaination, [`SmallestLenType`] is the mapping trait.
+pub type DefaultLenType<const N: usize> = <Const<N> as SmallestLenType>::Type;
+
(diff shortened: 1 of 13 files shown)


src/vec/mod.rs
```diff
@@ -528,13 +568,16 @@ impl<T, S: Storage> VecInner<T, S> {
...
+            .borrow_mut()
+            .get_unchecked_mut(self.len.into_usize()) = MaybeUninit::new(item);
 
-        self.len += 1;
+        self.len += LenT::ONE;
     }
 
     /// Shortens the vector, keeping the first `len` elements and dropping the rest.
-    pub fn truncate(&mut self, len: usize) {
+    pub fn truncate(&mut self, len: LenT) {
```
diondokter, line 580: Same here and many other functions, I think it'd be weird if all these API's don't use usize.

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+impl_lentodefault!(u16: 256, 300, 400, 500, 512, 600, 700, 800, 900, 1000, 1024, 2000, 2048, 4000, 4096, 8000, 8192, 16000, 16384, 32000, 32768, 65000, 65535);
+impl_lentodefault!(u32: 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824, 2147483648);
```
diondokter, line 95: Oh this is quite fun!
I wonder if there's a way to cover all numbers without breaking the compiler :P

diondokter (commented):
Not a full review from me, I've not read every line. This is a *fun* change and really cool. I do have some concerns though

src/vec/mod.rs
GnomedDev: Okay, so revert the changes made to arguments and return types as said in the PR description. Got it.

src/len_type.rs
GnomedDev: This doesn't actually cause too much stress on the compiler, it's surprising how fast the trait resolver is. From my knowledge (and I have done quite a bit of advanced type system logic, see https://github.com/GnomedDev/aformat) this is the only possible stable way to do this.

Since these containers probably shouldn't be used for N values above u8::MAX, let alone u16::MAX, I'm sure this will be fine, especially with the `diagnostic::on_unimplemented` making it super easy for users to fix.
image

src/vec/mod.rs
GnomedDev: Done.

t-moe:
May I ask what your motivation is for this change? Do you have any real-world use cases for it?

This seems like a premature optimization to me, and it might confuse users who are new to Rust, especially when they need to specify the full type of Vec.

If this change is necessary, a better approach might be to add LenT only to VecInner and let Vec use it with LenT=usize. This would keep the API simple while still allowing advanced users to leverage VecInner when needed.

GnomedDev:
I have been interested in performing this kind of optimization to a stack array for a while, and in-fact have a PR open to `arrayvec` to implement this, but the case that pushed me over the edge was `embedded-tls`'s usage of many 1 byte large `T`s in nested `Vec`'s in enums leading to `160` bytes of data to be take up `328` bytes of storage.

Users who are new to Rust will most likely not hit this, as they will not be generic over the length type and therefore will get the exact same experience due to the default. If they do stray into being generic or using large N values, it is explained in documentation and in the error message what to do (pick the smallest integer type for your N, or usize).

src/vec/mod.rs
```diff
@@ -152,25 +162,40 @@ impl<T, const N: usize> Vec<T, N> {
...
+    /// This can be useful if you need to pass a `Vec<T, N, u8>` into a `Vec<T, N, usize>` for example.
+    ///
+    /// This will check at compile time if the `N` value will fit into `NewLenT`, and error if not.
+    pub fn cast_len_type<NewLenT: LenType>(self) -> Vec<T, N, NewLenT> {
+        const { check_capacity_fits::<NewLenT, N>() }
+        let this = ManuallyDrop::new(self);
+
+        Vec {
+            len: NewLenT::from_usize(this.len()),
+            buffer: unsafe { ptr::read(&this.buffer) },
```
reitermarkus, line 195: Please document why this is safe.

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub const fn check_capacity_fits<LenT: LenType, const N: usize>() {
+    assert!(LenT::MAX >= N, "The capacity is larger than LenT can hold, increase the size of `LenT` or reduce the capacity")
```
reitermarkus, line 98:
```suggestion
    assert!(LenT::MAX >= N, "The capacity is larger than `LenT` can hold, increase the size of `LenT` or reduce the capacity")
```

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+        impl LenToDefault for Const<$len> {
+            type Default = $LenT;
+        }
+    )*};
+}
+
+pub struct Const<const N: usize>;
+
+#[diagnostic::on_unimplemented(
+    message = "Length `N` does not have a default LenType mapping",
```
reitermarkus, line 84:
```suggestion
    message = "Length `N` does not have a default `LenType` mapping",
```

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+    + SubAssign
+    + PartialOrd
+    + TryFrom<usize, Error: Debug>
+    + TryInto<usize, Error: Debug>
+{
+    /// The zero value of the integer type.
+    const ZERO: Self;
+    /// The one value of the integer type.
+    const ONE: Self;
+    /// The maxiumum value of this type, as a usize.
```
reitermarkus, line 51:
```suggestion
    /// The maximum value of this type, as a usize.
```

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+
+macro_rules! impl_lentodefault {
+    ($LenT:ty: $($len:literal),*) => {$(
+        impl LenToDefault for Const<$len> {
+            type Default = $LenT;
+        }
+    )*};
+}
+
+pub struct Const<const N: usize>;
```
reitermarkus, line 81:
```suggestion
#[non_exhaustive]
pub struct Const<const N: usize>;
```

Also needs to be documented.

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[diagnostic::on_unimplemented(
+    message = "Length `N` does not have a default LenType mapping",
+    note = "Provide the `LenType` explicitly, such as `usize`"
+)]
+pub trait LenToDefault {
```
reitermarkus, line 87: I'd remove `Default` from the naming here. Maybe something like `SmallestLenType`?

src/len_type.rs
```diff
@@ -0,0 +1,99 @@
...
+    type Default: LenType;
+}
+
+pub type DefaultLenType<const N: usize> = <Const<N> as LenToDefault>::Default;
```
reitermarkus, line 91: Needs to be documented.

src/string/mod.rs
```diff
@@ -210,9 +211,9 @@ impl<const N: usize> String<N> {
     /// # Ok::<(), core::str::Utf8Error>(())
     /// ```
     #[inline]
-    pub fn from_utf8(vec: Vec<u8, N>) -> Result<Self, Utf8Error> {
+    pub fn from_utf8<LenT: LenType>(vec: Vec<u8, N, LenT>) -> Result<Self, Utf8Error> {
         core::str::from_utf8(&vec)?;
-        Ok(Self { vec })
+        Ok(unsafe { Self::from_utf8_unchecked(vec) })
```
reitermarkus, line 216: Document why this is safe.

src/len_type.rs
GnomedDev: This is not publicly accessible, as the surrounding module is private and it is not re-exported. I have added documentation anyway though.

GnomedDev:
Sorted out all of your review comments, @reitermarkus!

YuhanLiin:
Interestingly, I made a similar MR in #204 2 years ago, but it was rejected for complexity.

GnomedDev:
It doesn't look like it was rejected for complexity, although that PR is significantly more complex due to typenum bounds, it was just closed because of the migration to const generics.

src/len_type.rs
Sunshine40: In fact, if you'd like to go further about space efficiency, you can use less than a whole byte for cases where capacity < 255 by using enums with fewer variants, so that niche optimization could encode enum discriminants into the unused bit patterns.

On the other hand, with element type `T`'s alignment > 1, it's often easier to find space in the padding bytes than to squeeze space from the actual `LenType`, but with the current default behavior of the type system ([the contents of padding bytes are unspecified rather than fixed](https://internals.rust-lang.org/t/missed-layout-optimization/20045/2)), I don't know the idiomatic way to let the container make its padding bytes specified.

Sunshine40: A possible approach would be to make `(T, ZeroPad)` a `LenType` with the same behavior as `<T as LenType>`

GnomedDev: This could be a future PR, but I'm trying to get this initial massive improvement before we start working on niche opt stuff.

reitermarkus:
@GnomedDev, sorry for the long wait. Can you have a look at rebasing this?

GnomedDev:
I'll take a look at rebasing it in the next week, if I don't get around to it after that time feel free to close or take over this PR.

GnomedDev:
It looks like this PR is going to have to be practically rewritten due to the VecStorage change... but I'll see if I can do that.

GnomedDev:
@reitermarkus Done, please note that I just did this from scratch instead of rebasing as it was conflicting so much, so keep that in mind while reviewing.

src/len_type.rs
```diff
@@ -0,0 +1,112 @@
...
+    const MAX: usize;
+
+    /// An infallible conversion from `usize` to `LenT`.
+    fn from_usize(val: usize) -> Self {
+        val.try_into().unwrap()
+    }
+
+    /// An infallible conversion from `LenT` to `usize`.
+    fn into_usize(self) -> usize {
+        self.try_into().unwrap()
```
max-heller, line 61: Could this be implemented in the macro expansion as `self as usize` to avoid the possibility of a panicking codepath given that `u8`, `u16`, `usize`, and `u32` (on 32/64bit targets) all fit in a `usize`?

GnomedDev: I would expect rustc to know that already and not emit a panic branch in release mode, I prefer to avoid `as` conversions when possible.

src/len_type.rs
```diff
@@ -0,0 +1,112 @@
...
+            const ONE: Self = 1;
+            const MAX: usize = Self::MAX as _;
+        }
+    )*}
+}
+
+/// A sealed trait representing a valid type to use as a length for a container.
+///
+/// This cannot be implemented in user code, and is restricted to `u8`, `u16`, `u32`, and `usize`.
+pub trait LenType:
```
reitermarkus, line 31: Leave this trait empty, and move all the implementation details into `Sealed`.

src/len_type.rs
reitermarkus: Yes, this should be optimized in release mode, but maybe add an `#[inline]` here as well.

src/len_type.rs
```diff
@@ -0,0 +1,112 @@
...
+    + Display
+    + Debug
+    + PartialEq
+    + Add<Output = Self>
+    + AddAssign
+    + Sub<Output = Self>
+    + SubAssign
+    + PartialOrd
+    + TryFrom<usize, Error: Debug>
+    + TryInto<usize, Error: Debug>
```
reitermarkus, line 45: Also move all of these to the `Sealed` trait.

src/string/mod.rs
```diff
@@ -229,9 +232,11 @@ impl<const N: usize> String<N> {
     /// # Ok::<(), core::str::Utf8Error>(())
     /// ```
     #[inline]
-    pub fn from_utf8(vec: Vec<u8, N>) -> Result<Self, Utf8Error> {
+    pub fn from_utf8<LenT: LenType>(vec: Vec<u8, N, LenT>) -> Result<Self, Utf8Error> {
```
reitermarkus, line 235: This can just use `usize`, unless you plan on also making `StringInner` generic.

reitermarkus: Ah, nevermind, the `Vec` can be generic even if `StringInner` isn't.

src/vec/drain.rs
```diff
@@ -19,25 +21,25 @@ use super::VecView;
 /// use heapless::{vec, Vec};
 ///
 /// let mut v = Vec::<_, 4>::from_array([0, 1, 2]);
-/// let iter: vec::Drain<'_, _> = v.drain(..);
+/// let iter: vec::Drain<'_, _, _> = v.drain(..);
```
reitermarkus, line 24: This example is kind of useless, might as well remove it given `Vec::drain` has a better one.

GnomedDev: True, removed.

src/len_type.rs
GnomedDev: Done.

src/len_type.rs
GnomedDev: Done.

src/len_type.rs
GnomedDev: Done.

src/len_type.rs
```diff
@@ -0,0 +1,107 @@
+use core::{
+    fmt::{Debug, Display},
+    ops::{Add, AddAssign, Sub, SubAssign},
+};
+
+pub trait Sealed:
```
reitermarkus, line 6:
```suggestion
trait Sealed:
```

reitermarkus (dismissed):

reitermarkus:
@GnomedDev, needs another rebase, but should be good to go after that.

> Adding the length generic to `VecView`, this can be changed to only store `usize` at the cost of only `Vec<T, N, usize>` being able to unsize (or requiring a `Vec::cast_len_type` call before unsizing, at least).

I'm quite split on this: If `VecView` only supports `usize`, it's easier to use, but then `Vec` will have to use `usize` by default, or it gets confusing, so the space-saving becomes opt-in. I sightly prefer the space-saving by default, since it applies to the most common sizes with `N < 256`, and keeping `Vec` and `VecView` more consistent.

GnomedDev:
@reitermarkus I've rebased, happy for this to finally get merged in.

reitermarkus (approved):

reitermarkus:
Thanks!

PR #508, opened 2024-08-08
Add `Deque::{swap, swap_unchecked, swap_remove_front, swap_remove_back}`

honzasp:
These methods are equivalent to their `std` counterparts (except `swap_unchecked()`, which is a natural extension).


honzasp:
I fixed the Miri error in CI, there remain some errors from Clippy, which are however not caused by this PR: it seems that a new version of Clippy flags new errors.

jannic:
The clippy warnings have been fixed in the mean time (#512).
Could you rebase your PR?

jannic:
We discussed this in today's matrix meeting and the response was positive. So @honzasp, if you rebase this, I'm sure it will be approved soon.

honzasp:
@jannic Rebased and all checks passed :tada:

Dirbaio (approved):
thnks!

PR #513, opened 2024-10-07
Added `is_full` function to `BinaryHeap`, `IndexMap`, `IndexSet` and `LinearMap`

sourcebox:
This function is already implemented for some data structures, so having it everywhere makes the API more consistent.


sourcebox:
Clippy check is failing because of errors not related to this PR.

Dirbaio (approved):
thanks~

PR #525, opened 2024-12-28
feat(indexmap): added `get_index`, `get_index_mut` and `get_index_of`

pamburus:
This adds `get_index`, `get_index_mut` and `get_index_of` methods to `IndexMap`.

See corresponding [get_index](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.get_index), [get_index_mut](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.get_index_mut) and [get_index_of](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.get_index_of) in the [indexmap](https://docs.rs/indexmap/latest/indexmap/index.html) crate.

Related pull requests:
* #526

PR #526, opened 2024-12-28
feat(indexmap): added `truncate`

pamburus:
This adds `truncate` method to `IndexMap`.

See corresponding [truncate](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.truncate) in the [indexmap](https://docs.rs/indexmap/latest/indexmap/index.html) crate.

Related pull requests:
* #525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Deque::{swap, swap_unchecked, swap_remove_front, swap_remove_back}`.
 - Make `String::from_utf8_unchecked` const.
 - Implemented `PartialEq` and `Eq` for `Deque`.
+- Added `truncate` to `IndexMap`.
 
 ### Changed
 
--- a/src/indexmap.rs
+++ b/src/indexmap.rs
@@ -1169,9 +1169,9 @@ where
         self.core.retain_in_order(move |k, v| f(k, v));
     }
 
-    /// Shortens the map, keeping the first len elements and dropping the rest.
+    /// Shortens the map, keeping the first `len` elements and dropping the rest.
     ///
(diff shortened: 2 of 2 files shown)


Dirbaio:
Thanks for the PR. Could you add tests?

src/indexmap.rs
```diff
@@ -1168,6 +1168,33 @@ where
         self.core.retain_in_order(move |k, v| f(k, v));
     }
 
+    /// Shortens the map, keeping the first len elements and dropping the rest.
```
reitermarkus, line 1171:
```suggestion
    /// Shortens the map, keeping the first `len` elements and dropping the rest.
```

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `SortedLinkedListView`, the `!Sized` version of `SortedLinkedList`.
 - Added implementation of `Borrow` and `BorrowMut` for `String` and `Vec`.
 - Added `Deque::{swap, swap_unchecked, swap_remove_front, swap_remove_back}`.
+- Added `truncate` to `IndexMap`
```
reitermarkus, line 42:
```suggestion
- Added `truncate` to `IndexMap`.
```

src/indexmap.rs
```diff
@@ -1168,6 +1168,33 @@ where
...
+    ///
+    /// If len is greater than the map’s current length, this has no effect.
```
reitermarkus, line 1173:
```suggestion
    /// If `len` is greater than the map's current length, this has no effect.
```

reitermarkus (approved):

reitermarkus:
There is a doc test, I think that's enough. I changed the order to 3, 2, 1 so it is more obvious what should happen.

PR #533, opened 2025-03-09
Remove allow(clippy::result_unit_err)

newAM:
Unit error types do not implement core::error::Error which leads to some ergonomic issues.

### DIFF
--- a/src/string/mod.rs
+++ b/src/string/mod.rs
@@ -12,6 +12,7 @@ use core::{
 };
 
 use crate::vec::{OwnedVecStorage, Vec, VecInner, VecStorage, ViewVecStorage};
+use crate::CapacityError;
 
 mod drain;
 pub use drain::Drain;
@@ -24,17 +25,25 @@ pub use drain::Drain;
 #[derive(Debug)]
 pub enum FromUtf16Error {
     /// The capacity of the `String` is too small for the given operation.
-    Capacity,
+    Capacity(CapacityError),
     /// Error decoding UTF-16.
-    DecodeUtf16Error(DecodeUtf16Error),
+    DecodeUtf16(DecodeUtf16Error),
 }
 
 impl fmt::Display for FromUtf16Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Capacity => "insufficient capacity".fmt(f),
-            Self::DecodeUtf16Error(e) => write!(f, "invalid UTF-16: {}", e),
+            Self::Capacity(err) => write!(f, "{err}"),
+            Self::DecodeUtf16(err) => write!(f, "invalid UTF-16: {err}"),
         }
+    }
+}
+
+impl core::error::Error for FromUtf16Error {}
+
+impl From<CapacityError> for FromUtf16Error {
+    fn from(e: CapacityError) -> Self {
+        Self::Capacity(e)
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/string/mod.rs
```diff
@@ -28,20 +28,28 @@ pub use drain::Drain;
 #[derive(Debug)]
 pub enum FromUtf16Error {
     /// The capacity of the `String` is too small for the given operation.
-    Capacity,
+    Capacity(CapacityError),
     /// Error decoding UTF-16.
     DecodeUtf16Error(DecodeUtf16Error),
```
reitermarkus, line 33: Given `Capacity` is not called `CapacityError`, this variant should probably be named just `DecodeUtf16`.

src/string/mod.rs
```diff
@@ -28,20 +28,28 @@ pub use drain::Drain;
...
 }
 
 impl fmt::Display for FromUtf16Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Capacity => "insufficient capacity".fmt(f),
+            Self::Capacity(e) => write!(f, "{}", e),
             Self::DecodeUtf16Error(e) => write!(f, "invalid UTF-16: {}", e),
```
reitermarkus, line 40:
```suggestion
            Self::Capacity(err) => write!(f, "{err}"),
            Self::DecodeUtf16Error(err) => write!(f, "invalid UTF-16: {err}"),
```

src/vec/mod.rs
```diff
@@ -16,6 +16,18 @@ use crate::storage::{OwnedStorage, Storage, ViewStorage};
 mod drain;
 pub use drain::Drain;
 
+/// The error type for fallible [`Vec`] and [`String`](crate::String) methods.
+#[derive(Debug)]
+pub struct CapacityError;
```
reitermarkus, line 21: This should be at the top level of the crate, given it's shared by all types.

reitermarkus: Also, make it `#[non_exhaustive]` just in case.

src/vec/mod.rs
```diff
@@ -563,13 +573,12 @@ impl<T, S: Storage> VecInner<T, S> {
...
     /// See also [`resize_default`](Self::resize_default).
-    #[allow(clippy::result_unit_err)]
-    pub fn resize(&mut self, new_len: usize, value: T) -> Result<(), ()>
+    pub fn resize(&mut self, new_len: usize, value: T) -> Result<(), CapacityError>
     where
         T: Clone,
     {
         if new_len > self.storage_capacity() {
-            return Err(());
+            return Err(CapacityError {});
```
reitermarkus, line 581:
```suggestion
            return Err(CapacityError);
```

src/string/mod.rs
newAM: Done!  That's consistent with `std::io::ErrorKind` naming.  Makes sense to me.

src/string/mod.rs
newAM: Done!

src/vec/mod.rs
newAM: Moved to the top level.

I didn't use an enum because you can't create a value of that type.

newAM: I did not know that, thanks, fixed!

reitermarkus (approved):

PR #534, opened 2025-03-09
LinearMap: remove allow(clippy::map_identity)

newAM:
This was previously required due to a false positive that has since been resolved.


pvdrz:
I was thinking about including this in #542 but noticed that this was already open. Given that #542 was merged I think this should be merged as well.

reitermarkus (approved):

PR #536, opened 2025-03-09
BinaryHeap: fix missing safety doc

newAM:


src/binary_heap.rs
```diff
@@ -389,7 +389,25 @@ where
...
+    /// # Example
+    ///
+    /// ```
+    /// use heapless::binary_heap::{BinaryHeap, Max};
+    ///
+    /// let mut heap: BinaryHeap<_, Max, 8> = BinaryHeap::new();
+    /// heap.push(42).unwrap();
+    ///
+    /// if !heap.is_empty() {
+    ///     // SAFETY: the heap is not empty
```
reitermarkus, line 406: I'd skip the `is_empty` check here since you wouldn't use it in a real use case for this, and argue with

```suggestion
    ///     // SAFETY: We just pushed a number onto the heap, so it cannot be empty.
```

src/binary_heap.rs
```diff
@@ -389,7 +389,25 @@ where
...
+    ///
+    /// The binary heap must not be empty.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use heapless::binary_heap::{BinaryHeap, Max};
+    ///
+    /// let mut heap: BinaryHeap<_, Max, 8> = BinaryHeap::new();
+    /// heap.push(42).unwrap();
```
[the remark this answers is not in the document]
newAM: Good catch, thanks!

src/binary_heap.rs
newAM: Good point, fixed!

src/binary_heap.rs
```diff
@@ -420,7 +437,25 @@ where
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use heapless::binary_heap::{BinaryHeap, Max};
+    ///
+    /// let mut heap: BinaryHeap<_, Max, 8> = BinaryHeap::new();
+    ///
+    /// if !heap.is_full() {
+    ///     // SAFETY: the heap is not full
```
reitermarkus, line 453: Same here for the `is_full` check, sorry if that was too implicit. I'd argue with

```suggestion
    ///     // SAFETY: We just created an empty heap of size 8, so it cannot be full.
```

reitermarkus (approved):
Good to merge once the `is_full` is removed.

src/binary_heap.rs
newAM: Done!

reitermarkus (approved):

PR #539, opened 2025-03-20
Remove `#[allow(clippy::result_unit_err)]`

pvdrz:
- **Remove unit errors from `String` and `Vec`**
- **Remove unit errors from `SortedLinkedList`**
- **Update changelog**

Fixes #533

There is an unsolved question regarding trait implementations with associated error types like `TryFrom` where in some cases `TryFrom::Err` is still `()`. In that case, we still have the issue of `()` not implementing `Error` but the only way to fix it would be to introduce a new error type on each case. Is that acceptable?


newAM:
Covered by #533 and #532 as well

pvdrz:
how did I miss this :clown_face:

Issue #494, opened 2024-07-01
Tracking issue for planned breaking changes:


sosthene-nitrokey:
- [x] Rework re-exports in lib.rs https://github.com/rust-embedded/heapless/pull/490#issuecomment-2200015830
- [x] Rename `Deque::storage_len` to `len`: https://github.com/rust-embedded/heapless/pull/490#issue-2383361459
- [x] Remove the `Storage` generic on `history_buf::OldestOrdered`. The view version can become the version that is always used: https://github.com/rust-embedded/heapless/pull/570
- [x] `Queue::capacity` and `Queue::storage_capacity` should be changed
- [x] `spsc`: The iterators, consumers and producers should be made to only have the `View` variants: https://github.com/rust-embedded/heapless/pull/571
- [x] `SortedLinkLinst`: `Iter` and `FindMut` should be made to only exist in the `View` version: https://github.com/rust-embedded/heapless/pull/540
- [x] `Vec`: rename `storage_capacity` to `capacity` and remove the `const` version https://github.com/rust-embedded/heapless/pull/538
- [x] `SortedLinkedList` removing unit error type: #532
- [x] `Vec` and `String` removing unit error type: #533

sgued:
This can be closed as everything is done and will be included in 0.9.1.

PR #540, opened 2025-03-20
`SortedLinkedList`: Keep only the view versions of `Iter` and `FindMut`

pvdrz:
- **Remove `sorted_linked_list::Iter` and `sorted_linked_list::IterInner`**
- **Remove `sorted_linked_list::FindMut` and `sorted_linked_list::FindMutInner`**
- **Introduce private `AsView` trait**
- **Update changelog**

See #494

### DIFF
--- a/src/sorted_linked_list.rs
+++ b/src/sorted_linked_list.rs
@@ -581,11 +581,12 @@ where
     }
 }
 
-impl<T, Idx, K> SortedLinkedListView<T, Idx, K>
+impl<T, Idx, K, S> SortedLinkedListInner<T, Idx, K, S>
 where
     T: Ord,
     Idx: SortedLinkedListIndex,
     K: Kind,
+    S: SortedLinkedListStorage<T, Idx> + ?Sized,
 {
     /// Get an iterator over the sorted list.
     ///
@@ -606,7 +607,7 @@ where
     /// ```
     pub fn iter(&self) -> IterView<'_, T, Idx, K> {
         IterView {
-            list: self,
+            list: S::as_view(self),
             index: self.head,
         }
     }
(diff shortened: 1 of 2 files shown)


src/sorted_linked_list.rs
```diff
@@ -272,18 +272,56 @@ impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, { u8::MAX as usize - 1 });
 impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, { u16::MAX as usize - 1 });
 impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, { usize::MAX - 1 });
 
+mod as_view {
+    use super::{SortedLinkedList, SortedLinkedListIndex, SortedLinkedListView};
+
+    pub trait AsView<T, Idx, K>
```
reitermarkus, line 278: This is now a sealed trait visible in public docs. Could these methods be added to `SortedLinkedListSealedStorage` instead?

src/sorted_linked_list.rs
```diff
@@ -726,7 +750,7 @@ where
     /// ll.push(3).unwrap();
     ///
     /// // Find a value and update it
-    /// let mut find = ll.find_mut(|v| *v == 2).unwrap();
+    /// let mut find = ll.as_mut_view().find_mut(|v| *v == 2).unwrap();
```
reitermarkus, line 753: Shouldn't need this here.

src/sorted_linked_list.rs
pvdrz: Yeah I think they could but it would be confusing as you'd need the whole list and not just the storage, so both `as_view` and `as_view_mut` would take `&SortedLinkedListInner<T, Self>` instead of just `&self`

reitermarkus: Seems like https://github.com/rust-embedded/heapless/pull/500 does exactly that for all types.

pvdrz: I'll rebase on top of that and see how to fix this then

reitermarkus:
Thanks!

PR #541, opened 2025-03-21
Add some conversion options for alloc vecs

rilipco:
This PR introudces a new feature `alloc` which will make `heapless` depend on the `alloc` crate and adds two trait implementations on `heapless::Vec` and `alloc::vec::Vec` to allow conversion between them.

Conversion from `heapless::Vec` to `alloc::vec::Vec` is fallible (due to OOM errors) and implemented via `TryFrom`.
Conversion from `alloc::vec::Vec` to `heapless::Vec` is fallible (due to capacity limit) and implemented via `TryFrom`.


newAM:
> Conversion from `heapless::Vec` to `alloc::vec::Vec` is infallible and implemented via `From`.

On embedded platforms people may want to handle allocation errors locally, such as using [try_with_capacity](https://doc.rust-lang.org/alloc/vec/struct.Vec.html#method.try_with_capacity) (sadly not stable at the moment).

I'm hesitant to implement an infallible implementation because in #381 I found that implementing infallible conversions prevents fallible ones from being implemented later.

newAM:
Would you be willing to remove the conversion from `heapless::Vec` to `alloc::vec::Vec` for now?

I'm happy to merge from `heapless::Vec` to `alloc::vec::Vec`!

newAM:
I realized a fallible `heapless::Vec` to `alloc::vec::Vec` is possible with the current APIs.

`alloc::vec::Vec::new` will not allocate until elements are pushed onto it.  That plus `try_reserve` should allow for a fallible conversion implementation.

src/vec/mod.rs
```diff
@@ -1218,6 +1218,53 @@ impl<T, S: VecStorage<T> + ?Sized> Drop for VecInner<T, S> {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if the `alloc::vec::Vec` fails to allocate memory.
+    fn try_from(vec: Vec<T, N>) -> Result<Self, Self::Error> {
+        let mut alloc_vec = alloc::vec::Vec::new();
+
+        // Allocate enough space for the elements, return an error if the
+        // allocation fails.
+        alloc_vec.try_reserve(vec.len()).map_err(|_| ())?;
```
reitermarkus, line 1258: Maybe use [`try_reserve_exact`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.try_reserve_exact) here?

src/vec/mod.rs
```diff
@@ -1218,6 +1218,53 @@ impl<T, S: VecStorage<T> + ?Sized> Drop for VecInner<T, S> {
...
+        }
+
+        Ok(vec)
+    }
+}
+
+#[cfg(feature = "alloc")]
+/// Converts the given `Vec<T, N>` into an `alloc::vec::Vec<T>`.
+impl<T, const N: usize> TryFrom<Vec<T, N>> for alloc::vec::Vec<T> {
+    type Error = ();
```
reitermarkus, line 1246: This should use [`TryReserveError`](https://doc.rust-lang.org/std/collections/struct.TryReserveError.html).

reitermarkus (approved):

PR #543, opened 2025-03-24
Remove `sealed` module

pvdrz:
All the assertions exposed by this module can be written using `const` for a while.


pvdrz:
This probably needs the skip changelog check label

newAM (changes requested):
These asserts were compile time before, I don't think we should replace them with runtime equivalents.

Dirbaio:
I think you can do `const { assert!(N > 0) }` to make it fail at compile time.

pvdrz:
oh sorry. I totally overlooked that fact. I force-pushed to include all the `const{ .. }` blocks as @Dirbaio suggested.

newAM (approved):
Thanks!

PR #546, opened 2025-04-02
defmt: 0.3 -> 1

newAM:


Cargo.toml
```diff
@@ -33,8 +33,8 @@ serde = ["dep:serde"]
 # implement ufmt traits.
 ufmt = ["dep:ufmt-write"]
 
-# Implement defmt::Format from defmt v0.3
-defmt-03 = ["dep:defmt"]
+# Implement defmt::Format from defmt v1
+defmt-1 = ["dep:defmt"]
```
reitermarkus, line 37: Given `serde` is at version 1, should probably just be called `defmt`.

newAM: The `defmt` authors have said in matrix there will never be a v2.  I think it's safe to remove the suffix, I'll do that later today.

ufmt should probably be suffixed, I'll add that in another PR.

reitermarkus (approved):

Cargo.toml
reitermarkus: Went ahead and removed the suffix.

PR #549, opened 2025-04-07
Add `CString` cont.

### LINKED ISSUE #330: Is there interest for `heapless::CString`?
I believe there's an interesting use-case for `heapless::Vec` to represent always null-terminated strings for FFI purposes

I've done something like this before in [unixstring](https://github.com/vrmiguel/unixstring) so I believe I could make a PR for this if the maintainers would like this functionality

```rust
// How it could look

let mut path = heapless::CString<128>;

path.append("/home/")?;
path.append(get_username())?;
path.append("/.config/my_app")?;

let config_dir_exists = nix::unistd::access(path.as_cstr(), AccessFlags::F_OK).is_ok();

### PR DESCRIPTION
**Pull request by raviqqe**
This PR is a continued work of #342, the heapless implementation of [`alloc::ffi::CString`](https://doc.rust-lang.org/stable/alloc/ffi/struct.CString.html). Credits to @vrmiguel.

Closes #330.

# Notable changes from the previous PR

- Implement `Deref<Target = CStr>` for `CString`.
  - As well as `Default`, `AsRef`, `Borrow`, `PartialEq`, `Eq`, `PartialOrd`, and `Ord`
- Remove the `len` and `as_str_unchecked` associated functions that should be implemented in the upstream of `core::ffi::CStr`.
- Rename the `push_bytes` associated function `extend_from_bytes`.
  - This change is to align the nominology with `heapless::Vec`.

# Future work

- `heapless::String` interoperability
   - `CString::into_string`
- `heapless::Vec` interoperability
   - `CString::from_vec_with_nul`
   - `CString::into_bytes`
   - `CString::into_bytes_with_nul`

### DIFF
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -118,6 +118,7 @@ impl<const N: usize> CString<N> {
     }
 
     /// Converts the [`CString`] to a [`CStr`] slice.
+    #[inline]
     pub fn as_c_str(&self) -> &CStr {
         debug_assert!(CStr::from_bytes_with_nul(&self.inner).is_ok());
 
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -13,7 +13,7 @@ use core::{
 /// A fixed capacity [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html).
 ///
 /// It stores up to `N - 1` non-nul characters with a trailing nul terminator.
-#[derive(Clone, Default, Hash)]
+#[derive(Clone, Hash)]
 pub struct CString<const N: usize> {
     inner: Vec<u8, N>,
 }
@@ -273,6 +273,13 @@ impl<const N: usize> CString<N> {
     }
 }
 
+impl<const N: usize> Default for CString<N> {
+    #[inline]
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl<const N: usize> Borrow<CStr> for CString<N> {
     #[inline]
     fn borrow(&self) -> &CStr {
@@ -444,6 +451,10 @@ mod tests {
             c_string.capacity_with_bytes(b"defg\0"),
             Some(INITIAL_BYTES.len() + 5)
         );
+    }
+    #[test]
+    fn default() {
+        assert_eq!(CString::<1>::default().as_c_str(), <&CStr>::default());
     }
 
     #[test]
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -292,6 +292,14 @@ impl<const N: usize> Deref for CString<N> {
     }
 }
 
+impl<const N: usize, const M: usize> PartialEq<CString<M>> for CString<N> {
+    fn eq(&self, rhs: &CString<M>) -> bool {
+        self.as_c_str() == rhs.as_c_str()
+    }
+}
+
+impl<const N: usize> Eq for CString<N> {}
+
 /// An error to extend [`CString`] with bytes.
 #[derive(Debug)]
 pub enum ExtendError {
@@ -329,6 +337,11 @@ mod tests {
         assert_eq!(empty.as_c_str(), <&CStr>::default());
         assert_eq!(empty.as_bytes(), &[]);
         assert_eq!(empty.to_str(), Ok(""));
+    }
+
+    #[test]
+    fn create_with_capacity_error() {
+        assert!(CString::<1>::from_bytes_with_nul(b"a\0").is_err())
     }
 
     #[test]
@@ -428,5 +441,37 @@ mod tests {
         let mut string = CString::<4>::new();
         string.extend_from_bytes(b"foo").unwrap();
         assert_eq!(Borrow::<CStr>::borrow(&string), c"foo");
+    }
+
+    #[test]
+    fn equal() {
+        // Empty strings
+        assert!(CString::<1>::new() == CString::<1>::new());
+        assert!(CString::<1>::new() == CString::<2>::new());
+        assert!(CString::<1>::from_bytes_with_nul(b"\0").unwrap() == CString::<3>::new());
+
+        // Single character
+        assert!(
+            CString::<2>::from_bytes_with_nul(b"a\0").unwrap()
+                == CString::<2>::from_bytes_with_nul(b"a\0").unwrap()
+        );
+        assert!(
+            CString::<2>::from_bytes_with_nul(b"a\0").unwrap()
+                == CString::<3>::from_bytes_with_nul(b"a\0").unwrap()
+        );
+        assert!(
+            CString::<2>::from_bytes_with_nul(b"a\0").unwrap()
+                != CString::<2>::from_bytes_with_nul(b"b\0").unwrap()
+        );
+
+        // Multiple characters
+        assert!(
+            CString::<4>::from_bytes_with_nul(b"abc\0").unwrap()
+                == CString::<4>::from_bytes_with_nul(b"abc\0").unwrap()
+        );
+        assert!(
+            CString::<3>::from_bytes_with_nul(b"ab\0").unwrap()
+                != CString::<4>::from_bytes_with_nul(b"abc\0").unwrap()
+        );
     }
 }
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -224,7 +224,7 @@ impl<const N: usize> CString<N> {
     ///
     /// If `additional` is not nul-terminated, the [`CString`] is left non nul-terminated, which is
     /// an invalid state. Caller must ensure that either `additional` has a terminating nul byte
-    /// or ensure to append a trailing null terminator.
+    /// or ensure to append a trailing nul terminator.
     unsafe fn extend_from_bytes_unchecked(
         &mut self,
         additional: &[u8],
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -8,7 +8,7 @@ use core::{
 
 /// A fixed capacity [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html).
 ///
-/// It stores up to N-1 elements with a byte reserved for the terminating nul byte.
+/// It stores up to `N - 1` elements with a byte reserved for the trailing nul terminator.
 #[derive(Clone, Default)]
 pub struct CString<const N: usize> {
     vec: Vec<u8, N>,
@@ -21,7 +21,7 @@ pub struct CString<const N: usize> {
 /// which should be the average slice size for this crate due to its use case.
 /// But ideally we'd use at least `BurntSushi`'s fallback implementation here.
 fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
-    haystack.iter().position(|&b| b == needle)
+    haystack.iter().position(|&byte| byte == needle)
 }
 
 impl<const N: usize> CString<N> {
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -29,7 +29,7 @@ impl<const N: usize> CString<N> {
     /// // including the nul terminator.
     /// let empty = CString::<10>::new();
     ///
-    /// assert_eq!(empty.as_c_str(), <&CStr>::default());
+    /// assert_eq!(empty.as_c_str(), c"");
     /// assert_eq!(empty.to_str(), Ok(""));
     /// ```
     pub fn new() -> Self {
--- a/src/c_string.rs
+++ b/src/c_string.rs
@@ -88,7 +88,7 @@ impl<const N: usize> CString<N> {
     pub fn from_bytes_with_nul(bytes: &[u8]) -> Result<Self, ExtendError> {
         let mut string = Self::new();
 
-        string.push_bytes(bytes)?;
+        string.extend_from_bytes(bytes)?;
 
         Ok(string)
     }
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: src/c_string.rs [hunk 1]
```diff
@@ -0,0 +1,461 @@
...
+    /// // Fixed-size CString that can store up to 10 elements
+    /// // (counting the null terminator)
+    /// let empty = CString::<10>::new();
+    ///
+    /// assert_eq!(empty.as_c_str(), <&CStr>::default());
+    /// assert_eq!(empty.to_str(), Ok(""));
+    /// ```
+    pub fn new() -> Self {
+        // TODO Resolve this.
+        // crate::sealed::greater_than_0::<CAP>();
```
raviqqe, line 45: Can we return `Result<Self, CapacityError>`? Is there any conventional way for static capacity check in this crate?

src/c_string.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// use heapless::CString;
+    ///
+    /// let mut cstr = CString::<10>::new();
+    ///
+    /// cstr.push_bytes(b"hey").unwrap();
+    /// cstr.push_bytes(b" there\0").unwrap();
+    ///
+    /// assert_eq!(cstr.to_str(), Ok("hey there"));
+    /// ```
+    pub fn push_bytes(&mut self, bytes: &[u8]) -> Result<(), CapacityError> {
```
raviqqe, line 212: Should we rename this `extend_from_bytes` or `extend_from_slice`?

src/c_string.rs
```diff
@@ -0,0 +1,469 @@
...
+    ///
+    /// // Insert 5 bytes to it
+    /// cstr.push_bytes(b"/etc/").unwrap();
+    /// // Length is 6 (5 bytes inserted plus null terminator)
+    /// assert_eq!(cstr.len(), 6);
+    /// cstr.push_bytes(b"dconf").unwrap();
+    ///
+    /// assert_eq!(cstr.to_str(), Ok("/etc/dconf"));
+    /// ```
+    pub fn len(&self) -> usize {
```
raviqqe, line 160: Not sure about this as `std::ffi::CString` didn't have this. To me, it looks like they didn't add it intentionally in order to avoid confusion around the null terminator...

raviqqe: I removed it for now. In the future, if the Rust's `std library team adds it to `CStr`, it will be available via `Deref`.

raviqqe:
FYI, I also removed `to_str` implementation because it's available via `Deref<Target = CStr>`.

src/c_string.rs
raviqqe: A more general question is if we need these utility functions implemented for `heapless::CString` which should technically be implemented for the upstream `core::ffi::CStr`.

raviqqe:
@reitermarkus Can you start reviewing this although it's not complete yet?

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+    /// // Fixed-size CString that can store up to 10 elements
+    /// // (counting the nul terminator)
+    /// let empty = CString::<10>::new();
+    ///
+    /// assert_eq!(empty.as_c_str(), <&CStr>::default());
+    /// assert_eq!(empty.to_str(), Ok(""));
+    /// ```
+    pub fn new() -> Self {
+        // TODO Resolve this.
+        // crate::sealed::greater_than_0::<N>();
```
reitermarkus, line 44: Inline `const { .. }` is stable.

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        let mut vec = Vec::new();
+
+        // Safety: will not fail since we const-assert that N > 0.
```
reitermarkus, line 48:
```suggestion
        // SAFETY: We just asserted that `N > 0`.
```

src/c_string.rs
vrmiguel: Makes sense to drop `len`, since `CStr::count_bytes` will be available from Deref

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+    ops::Deref,
+    str,
+};
+
+/// A fixed capacity [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html).
+///
+/// It stores up to `N - 1` elements with a byte reserved for the trailing nul terminator.
+#[derive(Clone, Default)]
+pub struct CString<const N: usize> {
+    vec: Vec<u8, N>,
```
reitermarkus, line 15:
```suggestion
    inner: Vec<u8, N>,
```

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+use crate::{vec::Vec, CapacityError};
+use core::{
+    ffi::{c_char, CStr},
+    ops::Deref,
+    str,
+};
+
+/// A fixed capacity [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html).
+///
+/// It stores up to `N - 1` elements with a byte reserved for the trailing nul terminator.
```
reitermarkus, line 12: Which “elements” are we talking about here? This should be reworded to refer to the length of the string.

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+            Some(nul_pos) if nul_pos + 1 == bytes.len() => {
+                // Safety: inserted bytes are nul-terminated so
+                //         the CString will be left in a valid state
+                unsafe { self.extend_slice(bytes) }?;
+
+                Ok(())
+            }
+            Some(_nul_pos) => {
+                // Found an interior nul byte
+                Err(CapacityError)
```
vrmiguel, line 197: I guess this would have to be a new `InteriorNulByteError` or etc

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+                self.vec.push(0).unwrap();
+
+                Ok(())
+            }
+        }
+    }
+
+    /// Removes the nul byte terminator from the inner buffer.
+    ///
+    /// # Safety: caller must ensure to re-add the terminator after this function is called
```
reitermarkus, line 220: Newline after heading.

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+    #[inline]
+    unsafe fn pop_terminator(&mut self) {
+        debug_assert_eq!(self.vec.last(), Some(&0));
+
+        unsafe { self.vec.pop_unchecked() };
```
reitermarkus, line 225: Needs a safety comment.

src/c_string.rs
```diff
@@ -0,0 +1,413 @@
...
+    /// ```rust
+    /// use heapless::CString;
+    ///
+    /// let mut cstr = CString::<10>::new();
+    /// cstr.push_bytes(b"heapless").unwrap();
+    ///
+    /// assert_eq!(unsafe { cstr.as_str_unchecked() }, "heapless",);
+    /// ```
+    pub unsafe fn as_str_unchecked(&self) -> &str {
+        str::from_utf8_unchecked(self.inner_without_nul())
```
reitermarkus, line 248: Needs nested `unsafe` and a safety comment.

src/ufmt.rs
```diff
@@ -32,6 +33,14 @@ impl<S: VecStorage<u8> + ?Sized> uWrite for VecInner<u8, S> {
     }
 }
 
+impl<const N: usize> uWrite for CString<N> {
+    type Error = CapacityError;
+    #[inline]
```
reitermarkus, line 38:
```suggestion

    #[inline]
```

src/c_string.rs
raviqqe: I removed this associated function for now for the same reason as `len()`.

src/c_string.rs
raviqqe: Updated "elements" as "non-nul characters".

src/c_string.rs
raviqqe: TIL!

src/c_string.rs
raviqqe: I renamed it `extend_from_bytes` for now. `extend_from_slice` is a bit ambiguous because it might mean either byte or `CStr` slices.

src/c_string.rs
```diff
@@ -0,0 +1,597 @@
...
+        let Some(capacity) = self.capacity_with_bytes(bytes) else {
+            return Ok(());
+        };
+
+        if capacity > N {
+            // Cannot store these bytes due to an insufficient capacity.
+            return Err(CapacityError.into());
+        }
+
+        match memchr(0, bytes) {
```
vrmiguel, line 184: I was wondering if we should use `CStr::from_bytes_with_nul` here directly (it uses the `memchr` crate)

```rust
    match CStr::from_bytes_with_nul(bytes) {
        Ok(_cstr) => {
            unsafe { self.extend_from_bytes_unchecked(bytes) }?;
            Ok(())
        },
        Err(FromBytesWithNulError::InteriorNul { position }) => {
            Err(ExtendError::InteriorNulByte(position))
        },
        Err(FromBytesWithNulError::NotNulTerminated) => {
            unsafe {
                self.extend_from_bytes_unchecked(bytes).unwrap();
                self.inner.push_unchecked(0);
            };
            Ok(())
        }
    }
}
```

reitermarkus: Makes sense.

vrmiguel (approved):
Awesome stuff @raviqqe !

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+    ///
+    /// assert_eq!(copied.to_str(), Ok("Hello, world!"));
+    /// ```
+    pub unsafe fn from_raw(ptr: *const c_char) -> Result<Self, ExtendError> {
+        // SAFETY: The given pointer to a string is assumed to be nul-terminated.
+        Self::from_bytes_with_nul(unsafe { CStr::from_ptr(ptr).to_bytes_with_nul() })
+    }
+
+    /// Converts the [`CString`] to a [`CStr`] slice.
+    pub fn as_c_str(&self) -> &CStr {
```
reitermarkus, line 121:
```suggestion
    #[inline]
    pub fn as_c_str(&self) -> &CStr {
```

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+        debug_assert!(CStr::from_bytes_with_nul(&self.inner).is_ok());
```
reitermarkus, line 122: No need for this.

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+    /// use heapless::CString;
+    ///
+    /// let mut c_string = CString::<5>::new();
+    /// c_string.extend_from_bytes(b"abc").unwrap();
+    ///
+    /// assert_eq!(c_string.as_bytes(), b"abc");
+    /// ```
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        debug_assert_eq!(self.inner.last().copied(), Some(0));
```
reitermarkus, line 264: No need for this.

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+    /// # Safety
+    ///
+    /// If `additional` is not nul-terminated, the [`CString`] is left non nul-terminated, which is
+    /// an invalid state. Caller must ensure that either `additional` has a terminating nul byte
+    /// or ensure to append a trailing nul terminator.
+    unsafe fn extend_from_bytes_unchecked(
+        &mut self,
+        additional: &[u8],
+    ) -> Result<(), CapacityError> {
+        self.pop_terminator();
```
reitermarkus, line 228: Add a nested `unsafe` and document why it is safe.

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+
+        self.inner.extend_from_slice(additional)
```
reitermarkus, line 230: Add a nested `unsafe` and document why it is safe.

src/c_string.rs
```diff
@@ -0,0 +1,594 @@
...
+impl<const N: usize> Deref for CString<N> {
+    type Target = CStr;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.as_c_str()
+    }
+}
+
+impl<const N: usize, T: AsRef<CStr>> PartialEq<T> for CString<N> {
```
reitermarkus, line 300: This doesn't exist in `std`. Only `PartialEq<M> for CString<N>` should be enough.

reitermarkus (changes requested):

src/c_string.rs
raviqqe: This is a safe method, I think...

reitermarkus: Right, `inner` is a `Vec`.

src/c_string.rs
```diff
@@ -0,0 +1,552 @@
...
+    error::Error,
+    ffi::{c_char, CStr, FromBytesWithNulError},
+    fmt,
+    ops::Deref,
+};
+
+/// A fixed capacity [`CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html).
+///
+/// It stores up to `N - 1` non-nul characters with a trailing nul terminator.
+#[derive(Clone, Hash)]
```
reitermarkus, line 16: Please also add a `Debug` implementation.

reitermarkus: The one from `std` should work as it simply uses `Deref`.

raviqqe: hmmm, this code:

```rust
            panic!(
                "DEBUG: {:?}",
                &CString::<2>::from_bytes_with_nul(b"a\0").unwrap()
            );
```

results in:

```
DEBUG: CString { inner: [97, 0] }
```

raviqqe: I ended up with a manual implementation for now.

reitermarkus: Yes, I meant a manual implementation, like the one on `std`.

reitermarkus (approved):

reitermarkus:
Thanks!

Issue #551, opened 2025-04-11
Implement generic LenType in other containers


GnomedDev:
After #504 `Vec` now has a generic LenType which significantly improves space efficiency, but the other containers depending on `Vec` do not as to avoid the original PR becoming super wide-reaching and impossible to merge.  

This is a tracking issue for adding LenType generics to the other containers.
- [ ] `StringInner`
- [ ] `LinearMapInner`
- [ ] `BinaryHeapInner`

PR #552, opened 2025-04-11
Add generic LenType to String

GnomedDev:
Part of #551.


src/string/drain.rs
```diff
@@ -8,40 +10,40 @@ use super::StringView;
...
+impl<LenT: LenType> Drop for Drain<'_, LenT> {
     fn drop(&mut self) {
         unsafe {
             // Use `Vec::drain`. “Reaffirm” the bounds checks to avoid
             // panic code being inserted again.
             let self_vec = (*self.string).as_mut_vec();
-            if self.start <= self.end && self.end <= self_vec.len() {
-                self_vec.drain(self.start..self.end);
+            if self.start <= self.end && self.end <= self_vec.length() {
+                self_vec.drain(self.start.into_usize()..self.end.into_usize());
```
reitermarkus, line 40: Convert them before the `if`, so you don't need the `length` method.

GnomedDev: Done.

reitermarkus (approved):

PR #553, opened 2025-04-11
Migrate SortedLinkedList to LenType

GnomedDev:
Part of #551.

The previous `SortedLinkedListIndex` seemed to store `Option<NonMaxLenT>` and required unsafe calls to get at the raw value without checking `::MAX`.... but none of those unsafe calls had safety comments and removing the unsafety shouldn't make the external interface unsound.

### DIFF
--- a/src/len_type.rs
+++ b/src/len_type.rs
@@ -41,7 +41,7 @@ pub trait Sealed:
 
     /// Converts `LenT` into `Some(usize)`, unless it's `Self::MAX`, where it returns `None`.
     #[inline]
-    fn option(self) -> Option<usize> {
+    fn to_non_max(self) -> Option<usize> {
         if self == Self::MAX {
             None
         } else {
--- a/src/len_type.rs
+++ b/src/len_type.rs
@@ -22,8 +22,10 @@ pub trait Sealed:
     const ZERO: Self;
     /// The one value of the integer type.
     const ONE: Self;
+    /// The maximum value of this type.
+    const MAX: Self;
     /// The maximum value of this type, as a `usize`.
-    const MAX: usize;
+    const MAX_USIZE: usize;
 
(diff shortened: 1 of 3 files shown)


src/len_type.rs
```diff
@@ -36,6 +38,16 @@ pub trait Sealed:
     fn into_usize(self) -> usize {
         self.try_into().unwrap()
     }
+
+    /// Converts `LenT` into `Some(usize)`, unless it's `Self::MAX`, where it returns `None`.
+    #[inline]
+    fn option(self) -> Option<usize> {
```
reitermarkus, line 44: Is this still used anywhere?

src/sorted_linked_list.rs
```diff
@@ -317,22 +264,22 @@ macro_rules! impl_index_and_const_new {
     };
 }
 
-impl_index_and_const_new!(LinkedIndexU8, u8, new_u8, { u8::MAX as usize - 1 });
-impl_index_and_const_new!(LinkedIndexU16, u16, new_u16, { u16::MAX as usize - 1 });
-impl_index_and_const_new!(LinkedIndexUsize, usize, new_usize, { usize::MAX - 1 });
+impl_const_new!(u8, new_u8);
+impl_const_new!(u16, new_u16);
+impl_const_new!(usize, new_usize);
```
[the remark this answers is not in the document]
GnomedDev: Not without losing `const`, sadly, because the lack of const traits.

src/len_type.rs
reitermarkus: Please rename this while you're at it, since `option` is a meaningless name.

Maybe `to_non_max`?

GnomedDev: Done.

reitermarkus (approved):

reitermarkus:
Thanks!

PR #555, opened 2025-04-21
Optimize size of heapless::Vec<_, 0> to 0 bytes

tomkris:
If Vec has 0 capacity, it can store only 0 elements, so its len can be assumed to be always 0.

I'm updating `DefaultLenType` to have a special case: `DefaultLenType<0>` will be represented by zero sized type which always has value 0.

This allows reducing size of `heapless::Vec<u8, 0>` from 8 bytes to 0 bytes.


reitermarkus:
@tomkris, can you please rebase this and add a test?

tomkris:
Rebased, added unit tests.

reitermarkus (approved):

reitermarkus:
Thanks!

Issue #557, opened 2025-04-24
Enhancement/Feature: port upstream IndexMap methods such as shift_remove


prutschman:
In #321 it was mentioned that there were upstream methods from `IndexMap` that hadn't been ported yet, `shift_remove` in that particular case. I don't know if this is important enough to be considered for 1.0, but I thought I'd re-surface the issue for discussion.

`retain` being in-order does mean there are workarounds; I'm currently simulating `shift_remove_index(0)` with a combination of `.first` and `.retain(|key| key != first_key)`

PR #558, opened 2025-04-24
WIP/RFC: shift_remove and friends

prutschman:
Per #557, I took a stab at implementing the `shift_remove_*` family of functions from upstream `IndexMap`. I'm sure there's better ways to do the things I did; there are some todo's around exactly how to do the fixup after removing from `entries`. I factored out the re-calculation logic from `.retain`, but recalculating everything is probably overly conservative.

Also, you may not even want these API functions. No worries if so! But if you're interested, I'm happy to make changes or re-do things.

### DIFF
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -302,43 +302,61 @@ where
     where
         F: FnMut(&mut K, &mut V) -> bool,
     {
-        const INIT: Option<Pos> = None;
-
         self.entries
             .retain_mut(|entry| keep(&mut entry.key, &mut entry.value));
 
         if self.entries.len() < self.indices.len() {
-            for index in self.indices.iter_mut() {
-                *index = INIT;
-            }
+            self.after_removal()
+        }
+    }
 
-            for (index, entry) in self.entries.iter().enumerate() {
-                let mut probe = entry.hash.desired_pos(Self::mask());
-                let mut dist = 0;
+    fn shift_remove_index(&mut self, index: usize) -> Option<(K, V)>
+    {
+        if index >= self.entries.len() {
+            return None;
+        }
 
-                probe_loop!(probe < self.indices.len(), {
-                    let pos = &mut self.indices[probe];
+        let bucket = self.entries.remove(index);
 
-                    if let Some(pos) = *pos {
-                        let entry_hash = pos.hash();
+        self.after_removal();
 
-                        // robin hood: steal the spot if it's better for us
-                        let their_dist = entry_hash.probe_distance(Self::mask(), probe);
-                        if their_dist < dist {
-                            Self::insert_phase_2(
-                                &mut self.indices,
-                                probe,
-                                Pos::new(index, entry.hash),
-                            );
-                            break;
-                        }
-                    } else {
-                        *pos = Some(Pos::new(index, entry.hash));
+        Some((bucket.key, bucket.value))
+    }
+
+    fn after_removal(&mut self)
+    {
+        const INIT: Option<Pos> = None;
+
+        for index in self.indices.iter_mut() {
+            *index = INIT;
+        }
+
+        for (index, entry) in self.entries.iter().enumerate() {
+            let mut probe = entry.hash.desired_pos(Self::mask());
+            let mut dist = 0;
+
+            probe_loop!(probe < self.indices.len(), {
+                let pos = &mut self.indices[probe];
+
+                if let Some(pos) = *pos {
+                    let entry_hash = pos.hash();
+
+                    // robin hood: steal the spot if it's better for us
+                    let their_dist = entry_hash.probe_distance(Self::mask(), probe);
+                    if their_dist < dist {
+                        Self::insert_phase_2(
+                            &mut self.indices,
+                            probe,
+                            Pos::new(index, entry.hash),
+                        );
                         break;
                     }
-                    dist += 1;
-                });
-            }
+                } else {
+                    *pos = Some(Pos::new(index, entry.hash));
+                    break;
+                }
+                dist += 1;
+            });
         }
     }
 
@@ -1229,6 +1247,38 @@ where
     {
         self.find(key)
             .map(|(probe, found)| self.core.remove_found(probe, found).1)
+    }
+
+    /// Remove the key-value pair at position `index` and return them.
+    ///
+    /// Like `Vec::remove`, the pair is removed by shifting all remaining items. This maintains
+    /// the remaining elements' insertion order, but is a more expensive operation
+    ///
+    /// Return `None` if `index` is not in 0..len().
+    ///
+    /// Computes in *O*(n) time (average).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::index_map::FnvIndexMap;
+    ///
+    /// let mut map = FnvIndexMap::<_, _, 8>::new();
+    /// map.insert(3, "a").unwrap();
+    /// map.insert(2, "b").unwrap();
+    /// map.insert(1, "c").unwrap();
+    /// let removed = map.shift_remove_index(1);
+    /// assert_eq!(removed, Some((2, "b")));
+    /// assert_eq!(map.len(), 2);
+    ///
+    /// let mut iter = map.iter();
+    /// assert_eq!(iter.next(), Some((&3, &"a")));
+    /// assert_eq!(iter.next(), Some((&1, &"c")));
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    pub fn shift_remove_index(&mut self, index: usize) -> Option<(K,V)>
+    {
+        self.core.shift_remove_index(index)
     }
 
     /// Retains only the elements specified by the predicate.
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -322,7 +322,18 @@ where
 
         Some((bucket.key, bucket.value))
     }
+    
+    fn shift_remove_found(&mut self, _probe: usize, found: usize) -> (K, V)
+    {
+        let entry = self.entries.remove(found);
 
+        self.after_removal(); /* Todo: pass probe if this starts taking an index parameter */
+
+        (entry.key, entry.value)
+    }
+
+    // Todo: Should this take in a parameter to allow it to only process the moved
+    // elements?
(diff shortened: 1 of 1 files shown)


src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+    pub fn shift_remove_index(&mut self, index: usize) -> Option<(K, V)> {
+        self.core.shift_remove_index(index)
+    }
+
+    /// Remove the key-value pair equivalent to `key` and return it and
+    /// the index it had.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation
```
reitermarkus, line 1293:
```suggestion
    /// insertion order, but is a more expensive operation.
```

src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+            let (k, v) = self.core.shift_remove_found(probe, found);
+            (found, k, v)
+        })
+    }
+
+    /// Remove and return the key-value pair equivalent to `key`.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation
```
reitermarkus, line 1332:
```suggestion
    /// insertion order, but is a more expensive operation.
```

src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+    {
+        self.shift_remove_full(key).map(|(_idx, k, v)| (k, v))
+    }
+
+    /// Remove the key-value pair equivalent to `key` and return
+    /// its value.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all of the
+    /// elements that follow it, preserving their relative order.
+    /// **This perturbs the index of all of those elements!**
```
reitermarkus, line 1369: Why is this only mentioned here?

src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+    /// Remove and return the key-value pair equivalent to `key`.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation
+    ///
+    /// Return `None` if `key` is not in map.
+    ///
+    /// Computes in **O(n)** time (average).
+    /// # Examples
```
reitermarkus, line 1337:
```suggestion
    /// Computes in *O*(n) time (average).
    ///
    /// # Examples
```

src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+    /// the index it had.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation
+    ///
+    /// Return `None` if `key` is not in map.
+    ///
+    /// Computes in **O(n)** time (average).
+    /// # Examples
```
reitermarkus, line 1298:
```suggestion
    /// Computes in *O*(n) time (average).
    ///
    /// # Examples
```

src/index_map.rs
```diff
@@ -1231,6 +1253,152 @@ where
...
+    /// Remove the key-value pair equivalent to `key` and return
+    /// its value.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all of the
+    /// elements that follow it, preserving their relative order.
+    /// **This perturbs the index of all of those elements!**
+    ///
+    /// Return `None` if `key` is not in map.
+    ///
+    /// Computes in **O(n)** time (average).
```
reitermarkus, line 1373:
```suggestion
    /// Computes in *O*(n) time (average).
```

src/index_map.rs
```diff
@@ -302,43 +302,65 @@ where
...
-                                &mut self.indices,
-                                probe,
-                                Pos::new(index, entry.hash),
-                            );
-                            break;
-                        }
-                    } else {
-                        *pos = Some(Pos::new(index, entry.hash));
+    // Todo: Should this take in a parameter to allow it to only process the moved
+    // elements?
```
reitermarkus, line 334: Yes. Technically it's *O*(n) in both cases, but one is *O*(`len()`) while the other is *O*(`len() / 2`).

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-
+- Added `shift_remove`, `shift_remove_entry`, `shift_remove_full`, `shift_remove_index` to `IndexMap`
```
reitermarkus, line 11:
```suggestion
- Added `shift_remove`, `shift_remove_entry`, `shift_remove_full`, `shift_remove_index` to `IndexMap`.
```

src/index_map.rs
prutschman: I based the docs initially on the upstream docs, which made a point of mentioning it. I'd stripped it out most places, since it seemed too obvious to state, but I missed that one. I'll harmonize the comments; do you have a preference for leaving it in or taking it out?

src/index_map.rs
prutschman: Done (see subsequent push). I don't fully understand the hashing logic, so I may have missed a subtlety.

prutschman: This is wrong, fix incoming.

src/index_map.rs
```diff
@@ -1231,6 +1251,154 @@ where
             .map(|(probe, found)| self.core.remove_found(probe, found).1)
     }
 
+    /// Remove the key-value pair at position `index` and return them.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation
+    ///
+    /// Return `None` if `index` is not in `0..len()`.
```
reitermarkus, line 1260:
```suggestion
    /// Returns `None` if `index` is not in `0..len()`.
```

src/index_map.rs
```diff
@@ -1231,6 +1251,154 @@ where
             .map(|(probe, found)| self.core.remove_found(probe, found).1)
     }
 
+    /// Remove the key-value pair at position `index` and return them.
```
reitermarkus, line 1254:
```suggestion
    /// Removes the key-value pair at position `index` and returns it.
```

src/index_map.rs
reitermarkus: Yeah, I guess the name of the method and the *O*(n) complexity make it obvious that the indices will also shift. So take it out.

src/index_map.rs
```diff
@@ -1231,6 +1251,154 @@ where
...
+    /// let mut iter = map.iter();
+    /// assert_eq!(iter.next(), Some((&3, &"a")));
+    /// assert_eq!(iter.next(), Some((&1, &"c")));
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    pub fn shift_remove_index(&mut self, index: usize) -> Option<(K, V)> {
+        self.core.shift_remove_index(index)
+    }
+
+    /// Remove the key-value pair equivalent to `key` and return it and
```
reitermarkus, line 1286:
```suggestion
    /// Removes the key-value pair equivalent to `key` and returns it and
```

src/index_map.rs
```diff
@@ -1231,6 +1251,154 @@ where
...
+    /// the index it had.
+    ///
+    /// Like [`Vec::remove`], the pair is removed by shifting all
+    /// remaining items. This maintains the remaining elements' relative
+    /// insertion order, but is a more expensive operation.
+    ///
+    /// Return `None` if `key` is not in map.
```
reitermarkus, line 1293:
```suggestion
    /// Returns `None` if `key` is not in map.
```

src/index_map.rs
reitermarkus: Please also add another test for the wrong logic.

prutschman: I've added a test for the wrong logic. However, I'm having a hard time figuring out what the correct logic should be.

I can replace only the affected entries' slots in `indices` with None, and and restrict the recalculation to only the affected entries. But, that fails too, because we would also need to rewrite "unaffected" entries that would have "preferred" to have a slot value that has since been vacated. (There's no longer a mismatched entry there, so when probed, it shows no value present, and the lookup fails.)

I can revert the "optimization" of `after_removal` to again re-process everything, and the test then passes. But we lose the putative optimization opportunity.

src/index_map.rs
prutschman: It's gone.

src/index_map.rs
prutschman: Fixed

src/index_map.rs
prutschman: Fixed, and found another like it to fix too.

src/index_map.rs
prutschman: As a test, I'm going to try the re-hashing logic from `remove_found`. It doesn't have the "robin hood" logic in it, but when I tried just leaving that logic intact I got an infinite loop. (Or I can abandon the efficiency quest for now.)

prutschman: I put the logic from `retain` back inline; I don't need to solve the problem of arbitrary removals, just a single removal. I switched to having just CoreMap::shift_remove_found, which now mirrors remove_found, plus an extra step to "shift" the appropriate index values after fixing up the removed slot. All IndexMap::shift_* operations are implemented in terms of shift_remove_found now. There's now moderate code duplication between `remove_found` and `shift_remove_found`, however.

prutschman:
The removal logic is still bad. I'm adding additional tests that verify that the map is fully internally consistent: every entry can be found, every index points to an entry consistent with the hash value, and the number of entries and non-empty indices is the same.

prutschman (commented):
Adding `IndexMap::assert_internally_consistent` as a `#[cfg(test)]`-only method was useful in troubleshooting the shift logic, but it's kind of intrusive. I'm open to other approaches. All the added Debug impls are just there to support assert_eq! instead of assert!, which maybe isn't needed.

PR #560, opened 2025-04-25
Fix #559: IndexMap::truncate leaves map in inconsistent state

### LINKED ISSUE #559: IndexMap::truncate leaves the map in an inconsistent state.
`IndexMap::truncate` truncates the entries, but doesn't clear the corresponding indices. As a result, attempts to find a key that was eliminated by the truncation will trigger a debug_assert. To see this, you can add a line to the doctest:

```
    /// ```
    /// use heapless::index_map::FnvIndexMap;
    ///
    /// let mut map = FnvIndexMap::<_, _, 8>::new();
    /// map.insert(3, "a").unwrap();
    /// map.insert(2, "b").unwrap();
    /// map.insert(1, "c").unwrap();
    /// map.truncate(2);
    /// assert_eq!(map.len(), 2);
    /// assert_eq!(map.get(&1), None);              // This is the new line that panics
    ///
    /// let mut iter = map.iter();
    /// assert_eq!(iter.next(), Some((&3, &"a")));
    /// assert_eq!(iter.next(), Some((&2, &"b")));
    /// assert_eq!(iter.next(), None);
    /// ```
```

This could be fixed either by changing find logic (impacting every lookup) or by re-writing the indices on truncate (which would make it O(n) instead of O(1)). Happy to make a PR for either.

prutschman:
Issue #559: `IndexMap::truncate` cleared the `entries`, but left dangling `indices`. This caused a `debug_assert` to fire if you searched for a key that had been truncated. Fixed by scanning the `indices` to remove the dangling ones, at the expense of making `truncate` O(n) instead of O(1). Added an assert to the doctest to catch this problem.

Fixes https://github.com/rust-embedded/heapless/issues/559.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed bug in `IndexMap::truncate` that left the map in an inconsistent state.
 - Fixed clippy lints.
 - Fixed `{arc,box,object}_pool!` emitting clippy lints.
 - Fixed the list of implemented data structures in the crate docs, by adding `Deque`,
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -1245,7 +1245,7 @@ where
     ///
     /// If `len` is greater than the map's current length, this has no effect.
     ///
-    /// Computes in *O*(1) time (average).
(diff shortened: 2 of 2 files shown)


reitermarkus (approved):

reitermarkus:
Thanks!

PR #563, opened 2025-04-26
hash32: 0.3.0 -> 1.0.0

newAM:
hash32 is not out yet, this branch is for testing before the hash32 release.

See <https://github.com/rust-embedded-community/hash32/issues/17>


zeenix:
hash32 1.0 was release a few months ago so that's not blocker anymore but this will be a breaking change so it has to be coordinated with all other breaking changes we'll want before 1.0.

newAM:
Forgot I had this still open, I'll update it later today so it's ready to be merged for 1.0

zeenix (approved):
:+1:

zeenix:
@newAM Awesome! I converted to draft to avoid accidental merging. I hope you don't mind. :)

amy-kwan:
Hi there! I opened a pull request recently to update the hash32 dependency, but closed it as I realized this PR already exists. I was wondering, are there any plans in merging this change in the near future? 

For context, `heapless` on AIX currently brings in the `byteorder` crate (through `hash32`) in the dependency tree and we aren't able to build `byteorder` on AIX due to an assertion described in https://github.com/BurntSushi/byteorder/issues/221.

zeenix:
> I was wondering, are there any plans in merging this change in the near future?

Did you read the discussion above? There is a good reason it's in draft state.

amy-kwan:
> > I was wondering, are there any plans in merging this change in the near future?
> 
> Did you read the discussion above? There is a good reason it's in draft state.

Thanks @zeenix - Yes, my apologies, I should rephrase my question. I was more so curious on the current state of things with respect to the other breaking changes and if we are getting to resolving those, which would affect when heapless v1.0 would be released.

Issue #372, opened 2023-07-19
Feature gating types


rjsberry:
I've used _heapless_ in a number of applications now and I often find I only use a subset of its functionality (e.g. just `Vec` and `String`).

It would be nice if we could gate types behind feature flags. For simpler use cases this could eliminate all build and runtime dependencies for simpler. This keeps the dependency graph lean which can be an important factor when evaluating crates.

PR #565, opened 2025-04-27
Add features for queues.

reitermarkus:
Partially addresses https://github.com/rust-embedded/heapless/issues/372.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `mpmc` and `spsc` features.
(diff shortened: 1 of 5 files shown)


Dirbaio:
Have you measured how much compile time this saves?

IMO features make sense if they prevent pulling in big dependencies, but in this case they just prevent compiling a few hundred lines of code. In this case the compile time difference is probably negligible, and not worth the dev-experience downgrade due to users having to enable features.

reitermarkus:
No, I didn't measure compile time.

Should we just close https://github.com/rust-embedded/heapless/issues/372 then? The only non-optional dependency we have is `hash32`, which is pretty tiny as well, with no dependencies itself.

Issue #568, opened 2025-04-29
`LenType` causes regression regarding supported `Vec` sizes and usage in libraries.


sosthene-nitrokey:
The addition of `LenType` to vecs and the `DefaultLenType` means that arbitrary length of `Vec` are much less usable.

# Not all lengths "just work" with `Vec` like they did previously

For example, the  following: `Vec<u8, 587>` does not compile anymore, and needs to be worked around with `Vec<u8, u16, 587>`.

This itself is I think a significant downgrade in readability (the benefits of `LenType` are in most cases I think not worth making code less readable, since the overhead of the Vec itself will often be negligible compared to the buffer of the Vec).

# In a library setting it forces the library writer to be explicit of the `LenType`:

```rust
fn some_func<const N: usize>() -> Vec<u8, N> {
    todo!()
}
```

Fails to compile with

```
```

sosthene-nitrokey:
And actually I thought the next release would have been 0.8.1, not a breaking release. In my PR I made efforts to make sure that there were no breaking changes, and documented the breaking changes that were pending, but they were not implemented once it was decided to go with a 0.9.0 instead of 0.8.1 release: https://github.com/rust-embedded/heapless/issues/494

I saw the PR for the 0.9 release I should have remembered this.

newAM:
This is a fairly serious issue that is going to break a lot of code.

I have yanked 0.9.0 from crates.io for now to save people the effort of attempting to update their code.

reitermarkus:
I wouldn't really call this a regression since `LenType` is working as designed, and is a breaking change and a bit less ergonomic, as was to be expected.


> `Vec<u8, 587>` does not compile anymore, and needs to be worked around with `Vec<u8, u16, 587>`.

We can add more sizes, e.g. all up to 1024 to avoid needing the explicit type in most common cases.


> In a library setting it forces the library writer to be explicit of the `LenType`:

Is that a bad thing? A library author should know whether `usize` is needed or e.g. `u8` is sufficient for a particular function.


> It makes compatibility with `View` types much harder.

Not that much harder:

```rust
let vec: &mut VecView<_, _> = &mut Vec::<u8, 8>::new();
```


Note this isn't an argument for or against anything, just a clarification that it isn't as dramatically broken as this issue suggests.

sosthene-nitrokey:
[...]

PR #569, opened 2025-04-29
Make LenType opt-in

sosthene-nitrokey:
See https://github.com/rust-embedded/heapless/issues/568

This makes `LenType` all default to `usize` and prepares for a 0.9.1 release (assuming you chose to yank 0.9.0, as I suggested)

Regarding `SortedLinkedList`:

Before https://github.com/rust-embedded/heapless/pull/553/files it was already using and `Idx` type for the same optimization. The PR only added the `DefaultLenType`, but all previous code needed to make the type explicit. I think it might be fine to leave it there and not default to usize (but at the cost of having everyone needing to re-order the generics in their types).


sosthene-nitrokey:
We can also add back some new type aliases that use the `DefaultLenType` mechanisms, next to the main `Vec` and `String` types. But I didn't want this to be blocked by bikeshedding on the name of those.

sosthene-nitrokey:
Wait, maybe a release should not be made with just that. 0.9.0 was missing multiple planned breaking changes: https://github.com/rust-embedded/heapless/issues/494

sosthene-nitrokey:
To consider merging before this: 

- [ ] https://github.com/rust-embedded/heapless/pull/571
- [ ] https://github.com/rust-embedded/heapless/pull/570

zeenix:
> Wait, maybe a release should not be made with just that. 0.9.0 was missing multiple wanted breaking changes: https://github.com/rust-embedded/heapless/issues/494

I think the idea was that 0.9.0 has waited already **very** long and any more breaking changes should go into the next breaking release (perhaps 1.0).

Cargo.toml
```diff
@@ -12,7 +12,7 @@ keywords = ["static", "no-heap"]
 license = "MIT OR Apache-2.0"
 name = "heapless"
 repository = "https://github.com/rust-embedded/heapless"
-version = "0.9.0"
+version = "0.9.1"
```
reitermarkus, line 15: Please move this into a separate PR.

sosthene-nitrokey:
> I think the idea was that 0.9.0 has waited already **very** long and any more breaking changes should go into the next breaking release (perhaps 1.0).

These breaking changes were not made part of the addition of the `View` types because my initial PRs were made with the goal of not including breaking changes to be included in a 0.8.1 release. These breaking changes are not very intrusive, and shouldn't need any code change for the majority of uses (and if it does it should be trivial). Adding them to the 0.9.0 release just makes the code simpler, both for heapless and for downstreams.

Cargo.toml
sosthene-nitrokey: Moved into #574

zeenix:
> > I think the idea was that 0.9.0 has waited already **very** long and any more breaking changes should go into the next breaking release (perhaps 1.0).
> 
> These breaking changes were not made part of the addition of the `View` types because my initial PRs were made with the goal of not including breaking changes to be included in a 0.8.1 release.

Ah ok. Thanks for explaining. :+1: 

> These breaking changes are not very intrusive, and shouldn't need any code change for the majority of uses (and if it does it should be trivial). Adding them to the 0.9.0 release just makes the code simpler, both for heapless and for downstreams.

Cool. I just hope that they're easy to review/merge and won't delay 0.9.1.

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 
 - Removed invalid `bytes::Buf` implementation.
+- Removed `DefaultLenType` struct
```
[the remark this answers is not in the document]
sosthene-nitrokey: Done!

src/vec/mod.rs
```diff
@@ -1806,7 +1805,7 @@ mod tests {
         {
             let v: Vec<Droppable, 2> = Vec::new();
             let v: Box<Vec<Droppable, 2>> = Box::new(v);
-            let mut v: Box<VecView<Droppable, u8>> = v;
+            let mut v: Box<VecView<Droppable, usize>> = v;
```
reitermarkus, line 1808: This is now redundant, right?

sosthene-nitrokey: Good catch!

reitermarkus (approved):

PR #571, opened 2025-04-29
De-monomorphize spsc consumer and producer

sosthene-nitrokey:
This needs a workaround for the `const` use case however


src/spsc.rs
```diff
@@ -464,8 +597,8 @@ impl<T, S: Storage> QueueInner<T, S> {
     ///     consumer.dequeue().unwrap();
     /// }
     /// ```
-    pub const fn split(&mut self) -> (ProducerInner<'_, T, S>, ConsumerInner<'_, T, S>) {
-        (ProducerInner { rb: self }, ConsumerInner { rb: self })
+    pub const fn split_const(&mut self) -> (Producer<'_, T>, Consumer<'_, T>) {
```
reitermarkus, line 600: The example here doesn't actually call this on `QueueView`. In fact, I don't think this makes sense on `QueueView`, since we first need a static `Queue` anyways, so there is no point in going through a `QueueView` to split it.

sosthene-nitrokey: It makes sense to have it available for `QueueView` if you have some `const` functions that does some initialization, which you use to build the statics.

I updated the doc to use the View.

src/spsc.rs
```diff
@@ -427,13 +450,123 @@ impl<T, S: Storage> QueueInner<T, S> {
...
+    ///
+    /// fn main() {
+    ///     let mut consumer = {
+    ///         let (p, c) = {
+    ///             static mut Q: Queue<(), 4> = Queue::new();
+    ///             // SAFETY: `Q` is only accessible in this scope
+    ///             // and `main` is only called once.
+    ///             #[allow(static_mut_refs)]
+    ///             unsafe {
+    ///                 Q.split()
```
reitermarkus, line 534: This should be on `split`, not `split_const`.

sosthene-nitrokey: Fixed

sosthene-nitrokey: Fixed

src/spsc.rs
zeenix: @reitermarkus Are you satisfied with @sosthene-nitrokey's response? If so, I believe we can get a 0.9 release now?

src/spsc.rs
```diff
@@ -363,8 +363,20 @@ impl<T, S: Storage> QueueInner<T, S> {
 
     /// Splits a queue into producer and consumer endpoints.
     ///
+    /// If you need this function in a `const` context, check out [`Queue::split_const`]() and [`Queueview::split_const`]()
```
reitermarkus:
```suggestion
    /// If you need this function in a `const` context, check out [`Queue::split_const`] and [`QueueView::split_const`].
```

sosthene-nitrokey: Done

zeenix (approved):

zeenix:
@reitermarkus even with my new super powers 🦸‍♂️, it'll still need your approval since you requested changes.

zeenix:
> @reitermarkus even with my new super powers 🦸‍♂️, it'll still need your approval since you requested changes.

@sosthene-nitrokey I think @reitermarkus has no time. Since you need to rebase this PR anyway, perhaps best you close this and create a new PR?

sosthene-nitrokey:
Re-opened as #590

PR #577, opened 2025-05-16
Add LinearMap entry() and retain()

### LINKED ISSUE #471: LinearMap entry API
Currently to insert in the `LinearMap` and get a mutable reference to the corresponding value (`map.entry(key).or_insert_with(...)`), it's necessary to `.insert()` then `.get().unwrap()`, because there is no entry API or assimilated.

This implies:
- Cloning the key, in order to be able to call `get()`
- Iterating on the map twice: once to realize the entry is not there, and once to get the inserted entry

There was already an issue for entry APIs for maps: #103, but that was closed after only implementing it for only one of the map types: #289 - I would tend to believe that this was not intentional.

mkj:
I've made `linear_map` public as a namespace for `linear_map::Entry`. There could be a bit of confusion with the `crate::Entry` re-export of `index_map::Entry`, not sure if there's any better option.

Closes #471

### DIFF
--- a/src/linear_map.rs
+++ b/src/linear_map.rs
@@ -509,6 +509,16 @@ where
             }),
         }
     }
+
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs `(k, v)` for which `f(&k, &mut v)` returns `false`.
+    pub fn retain<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&K, &mut V) -> bool,
+    {
+        self.buffer.retain_mut(|(k, v)| f(k, v));
+    }
 }
 
 impl<K, V, Q, S: LinearMapStorage<K, V> + ?Sized> ops::Index<&'_ Q> for LinearMapInner<K, V, S>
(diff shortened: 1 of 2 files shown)


src/linear_map.rs
```diff
@@ -635,11 +673,111 @@ where
...
+    /// The entry corresponding to the key `K` does not exist in the map
+    Vacant(VacantEntry<'a, K, V, S>),
+}
+
+/// An occupied entry which can be manipulated
+pub struct OccupiedEntry<'a, K, V, S: LinearMapStorage<K, V> + ?Sized> {
+    // SAFETY: `idx` must not be modified after construction, and
+    // the size of `map` must not be changed.
+    idx: usize,
+    map: &'a mut LinearMapInner<K, V, S>,
```
sgued: This can instead be a `LinearMapView<K, V>`. This removes one generic parameter and makes the code more flexible (so a function operating on entries does not need to be generic over the storage/capacity).

src/linear_map.rs
```diff
@@ -635,11 +673,111 @@ where
...
+    /// The entry corresponding to the key `K` exists in the map
+    Occupied(OccupiedEntry<'a, K, V, S>),
+    /// The entry corresponding to the key `K` does not exist in the map
+    Vacant(VacantEntry<'a, K, V, S>),
+}
+
+/// An occupied entry which can be manipulated
+pub struct OccupiedEntry<'a, K, V, S: LinearMapStorage<K, V> + ?Sized> {
+    // SAFETY: `idx` must not be modified after construction, and
+    // the size of `map` must not be changed.
```
sgued, line 701:
```suggestion
    // SAFETY: `idx` must point to an existing item of the buffer
```

I don't think that there is any safety issue as long as `idx < buffer.len()`, only  potential logic issues.

src/linear_map.rs
```diff
@@ -635,11 +673,111 @@ where
...
+        // SAFETY: Valid idx from OccupiedEntry construction
+        let (_k, v) = unsafe { self.map.buffer.get_unchecked_mut(self.idx) };
+        mem::replace(v, value)
+    }
+}
+
+/// A view into an empty slot in the underlying map
+pub struct VacantEntry<'a, K, V, S: LinearMapStorage<K, V> + ?Sized> {
+    key: K,
+    map: &'a mut LinearMapInner<K, V, S>,
```
sgued: Same as for the occupied entry. This can be a `LinearMapView<K, V>`.

sgued (changes requested):
Looking mostly good. Some quick nits and it should be fine.

src/linear_map.rs
mkj: It's using `get_unchecked()` so if `self.map.buffer` shrank it'd be UB. Maybe I've missed something though?

I wasn't sure whether it was worth using `_unchecked()` functions, was following the convention of other bits of heapless.

src/linear_map.rs
mkj: Thanks, that's nicer.

sgued (approved):

PR #580, opened 2025-06-10
Add missing `Debug` derive to `vec::IntoIter`

zeenix:


sosthene-nitrokey (commented):
This would lead to undefined behaviour.

The `vec` field in the `IntoIter` struct is not in a valid state after iteration has begun. All elements before `next` have been "removed" from the Vec through `unsafe` code even though it's still there. All methods on `IntoIter` must take care to not read from them.

If you want a `Debug` implementation, you will need not to go through `derive`, but instead write a manual implementation. You can take a look at the `Clone` implementation to see how `Debug` could be implemented manually.

sosthene-nitrokey:
This is where [Unsafe fields](https://github.com/rust-lang/rfcs/pull/3458) could come in  handy in the future.

zeenix:
@sosthene-nitrokey Thanks for the review. PTALA. I was first going for a more verbose option (`f.debug_struct("IntoIter").field("elements", &s).finish()`) but then decided against it. I can change to something like that, if you like.

sosthene-nitrokey (approved):

sgued (approved):

zeenix:
Rebased.

sgued (approved):

PR #584, opened 2025-07-14
String insertion

riley-williams:
Basically just copies the implementations of `insert` and `insert_str` from `std::String` with minor modification.

### DIFF
--- a/src/string/mod.rs
+++ b/src/string/mod.rs
@@ -716,26 +716,23 @@ impl<LenT: LenType, S: StringStorage + ?Sized> StringInner<LenT, S> {
 
         // SAFETY: Move the bytes starting from `idx` to their new location `ch_len`
         // bytes ahead. This is safe because we checked `len + ch_len` does not
-        // exceed the capacity and `idx` is a char boundary
+        // exceed the capacity and `idx` is a char boundary.
         unsafe {
             let ptr = self.vec.as_mut_ptr();
             core::ptr::copy(ptr.add(idx), ptr.add(idx + ch_len), len - idx);
         }
 
-        // SAFETY: Copy the encoded character into the vacated region if
-        // `idx != len`, or into the uninitialized spare capacity otherwise.
+        // SAFETY: Encode the character into the vacated region if `idx != len`,
+        // or into the uninitialized spare capacity otherwise. This is safe
+        // because `is_char_boundary` checks that `idx <= len`, and we checked that
+        // `(idx + ch_len)` does not exceed the capacity.
         unsafe {
-            // 4 bytes is the maximum length of a UTF-8 character
-            let mut buf = [0u8; 4];
-            let encoded = ch.encode_utf8(&mut buf);
-            core::ptr::copy_nonoverlapping(
-                encoded.as_ptr(),
-                self.vec.as_mut_ptr().add(idx),
-                ch_len,
-            );
+            let buf = core::slice::from_raw_parts_mut(self.vec.as_mut_ptr().add(idx), ch_len);
+            ch.encode_utf8(buf);
         }
 
-        // SAFETY: Update the length to include the newly added bytes.
+        // SAFETY: Update the length to include the newly added bytes. This is
+        // safe because we checked that `len + ch_len` does not exceed the capacity.
         unsafe {
             self.vec.set_len(len + ch_len);
         }


src/string/mod.rs
```diff
@@ -677,6 +677,138 @@ impl<LenT: LenType, S: StringStorage + ?Sized> StringInner<LenT, S> {
...
+        // `idx != len`, or into the uninitialized spare capacity otherwise.
+        unsafe {
+            // 4 bytes is the maximum length of a UTF-8 character
+            let mut buf = [0u8; 4];
+            let encoded = ch.encode_utf8(&mut buf);
+            core::ptr::copy_nonoverlapping(
+                encoded.as_ptr(),
+                self.vec.as_mut_ptr().add(idx),
+                ch_len,
+            );
```
riley-williams: This differs from the `std::String` implementation which uses an unstable method to encode directly into the destination buffer.

```rust
core::char::encode_utf8_raw_unchecked(ch as u32, self.vec.as_mut_ptr().add(idx));
```

There's probably a better way than what I've done here.

riley-williams:
Failing because of some existing clippy issues, opened a separate PR to fix them: #585

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Added `String::insert` and `String::insert_str`
```
zeenix: nit: Please follow the convention here:

```suggestion
- Added `String::insert` and `String::insert_str`.
```

src/string/mod.rs
zeenix: Can't we directly pass a mutable ref of `self.vec` to `char::encode_utf8` directly? Something like:

```rust
let buf = (&mut self.vec)[idx..idx+4];
ch.encode_utf8(buf);
```

src/string/mod.rs
```diff
@@ -677,6 +677,132 @@ impl<LenT: LenType, S: StringStorage + ?Sized> StringInner<LenT, S> {
...
+        // SAFETY: Copy the new string slice into the vacated region if `idx != len`,
+        // or into the uninitialized spare capacity otherwise. The borrow checker
+        // ensures that the source and destination do not overlap.
+        unsafe {
+            core::ptr::copy_nonoverlapping(
+                string.as_ptr(),
+                self.vec.as_mut_ptr().add(idx),
+                string_len,
+            );
+        }
```
zeenix, line 794: Do we really need unsafe operations to copy bytes into buffers? :thinking:

src/string/mod.rs
```diff
@@ -1240,4 +1366,103 @@ mod tests {
         let formatted = format!(2; "123");
         assert_eq!(formatted, Err(core::fmt::Error));
     }
+
+    #[test]
+    fn insert() {
```
zeenix, line 1368: Kudos for adding tests!

zeenix (changes requested):
LGTM overall. Some nits still though.

src/string/mod.rs
riley-williams: Probably not, I just assumed `std::String` had a good reason to do it this way so I kept it 🤷. The `copy_nonoverlapping` seems like a reasonable optimization hint, not sure if that's otherwise guaranteed to be inferred from mutable aliasing rules without the unsafe.

https://doc.rust-lang.org/src/alloc/string.rs.html#1777-1802

src/string/mod.rs
riley-williams: Yes...this seems obvious now that you've pointed it out 😅 

We'd have to update the length first to pass the slice bound check, because the underlying buffer length is separate from the string length, which is what the slice checks.

```rust
// SAFETY: Update the length to hold the number of bytes we're about to write.
unsafe {
    self.vec.set_len(len + ch_len);
}
// Encode the character into the vacated region if
// `idx != len`, or into the uninitialized spare capacity otherwise.
ch.encode_utf8(&mut self.vec[idx..idx + ch_len]);
```

Updating the length first just feels weird to me, but I can't think of any actual reason it wouldn't be safe.
I'm also not sure if the (infallible) slice bound check actually gets optimized away.

This is another alternative that does maintain the op order and doesn't add extra bounds checks:

```rust
// SAFETY: Encode the character into the vacated region if `idx != len`,
// or into the uninitialized spare capacity otherwise. This is safe
// because `is_char_boundary` checks that `idx <= len`, and we checked that
// `(idx + ch_len)` does not exceed the capacity.
unsafe {
    let buf = core::slice::from_raw_parts_mut(self.vec.as_mut_ptr().add(idx), ch_len);
    ch.encode_utf8(buf);
}

// SAFETY: Update the length to include the newly added bytes. This is
// safe because we checked that `len + ch_len` does not exceed the capacity.
unsafe {
    self.vec.set_len(len + ch_len);
}
```

riley-williams: I updated the PR with the latter, but I wasn't able to tell which was actually better in the godbolt output. Not because they're functionally identical, I'm just not good enough at reading it yet lol. I'll go with whatever you think makes more sense.

src/string/mod.rs
zeenix: Right. Ok, let's keep it this way then. :)

src/string/mod.rs
zeenix: Thanks so much for indulging me and trying out all possibilities here. :+1: 

My main worry was to avoid unsafe usage, if possible but if we have to make a call to set_len anyway, there's no much point to avoid the other call either, if we're certain that it's safe. It's good that we've MIRI setup in the CI so it can catch possible issues with unsafe code.

zeenix:
@sosthene-nitrokey When you've a moment, please take a look. I think the changes are good but it's good to have 2 set of eyes before merging. :)

PR #586, opened 2025-07-21
Add optional `embedded_io::Write` impl for `Vec`.

zyp:
This is analogous to how `std::vec::Vec` implements `std::io::Write`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
           components: miri
 
       - name: Run miri
-        run: MIRIFLAGS=-Zmiri-ignore-leaks cargo miri test --features="alloc,defmt,mpmc_large,portable-atomic-critical-section,serde,ufmt,bytes,zeroize"
+        run: MIRIFLAGS=-Zmiri-ignore-leaks cargo miri test --features="alloc,defmt,mpmc_large,portable-atomic-critical-section,serde,ufmt,bytes,zeroize,embedded-io-v0.7"
 
   # Run cargo test
   test:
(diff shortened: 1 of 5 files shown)


chrysn:
This'd be a very welcome addition. If embedded-io's stability is a concern, https://github.com/rust-embedded/embedded-hal/issues/566 ("is there anything to add before a 1.0?") would be a discussion that could use some implementators' input.

src/embedded_io.rs
```diff
@@ -0,0 +1,54 @@
+use crate::{
+    len_type::LenType,
+    vec::{VecInner, VecStorage},
+    CapacityError,
+};
+use embedded_io::{Error, ErrorKind, ErrorType, Write};
```
reitermarkus:
```suggestion
use embedded_io::{Error, ErrorKind, ErrorType, Write};

use crate::{
    len_type::LenType,
    vec::{VecInner, VecStorage},
    CapacityError,
};
```

reitermarkus: Prefer grouping imports by `core`, external dependencies, and `crate`, in that order.

.github/workflows/build.yml
```diff
@@ -176,6 +176,7 @@ jobs:
           cargo check --target="${target}" --features="portable-atomic-critical-section"
           cargo check --target="${target}" --features="serde"
           cargo check --target="${target}" --features="ufmt"
+          cargo check --target="${target}" --features="embedded-io"
```
sgued: Please add the embedded-io feature to the tests and to the miri test runs too

sgued (commented):
I think it we should implement it for `VecDeque` too.

.github/workflows/build.yml
zyp: How do you want this done? None of the other features appears to be tested in this manner, so I'm not sure what you expect.

sgued: Add the embedded-io feature line 51 and 87 of the `.github/workflows/build.yml` file.

sgued:
I rebased it on top of main. To address @chrysn 's concern we could rename the feature flag to `embedded-io-v0.6`. What do you think?

chrysn:
We're one rebase short of releasing embedded-io 0.7, so maybe wait a day and rename to that?

PR #588, opened 2025-08-13
Better names for method returning mutable references

zeenix:
I believe these names are better because they are more consistent with
the naming of their non-mutable siblings and also with the general
conventions in the Rust world.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Made `LenType` opt-in.
 - Minor fixes to `pool::boxed` docs.
 - Add missing `Debug` derive to `vec::IntoIter`.
+- Improved renaming of `VecSealedStorage`'s mutable getters.
```
sgued: This is a fully internal API, that hasn't been released yet. No need to put it in the changelog.

PR #590, opened 2025-08-18
De-monomorphize spsc consumer and producer

sgued:
Reopening #571 from my maintainer account.

### DIFF
--- a/src/spsc.rs
+++ b/src/spsc.rs
@@ -363,7 +363,8 @@ impl<T, S: Storage> QueueInner<T, S> {
 
     /// Splits a queue into producer and consumer endpoints.
     ///
-    /// If you need this function in a `const` context, check out [`Queue::split_const`] and [`QueueView::split_const`].
+    /// If you need this function in a `const` context,
+    /// check out [`Queue::split_const`] and [`QueueView::split_const`].
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


src/spsc.rs
```diff
@@ -363,8 +363,20 @@ impl<T, S: Storage> QueueInner<T, S> {
 
     /// Splits a queue into producer and consumer endpoints.
     ///
+    /// If you need this function in a `const` context, check out [`Queue::split_const`] and [`QueueView::split_const`].
```
zeenix: Nitpick: Can we keep all lines <= 100 chars?

src/spsc.rs
```diff
@@ -373,12 +385,11 @@ impl<T, S: Storage> QueueInner<T, S> {
     /// use critical_section::Mutex;
     /// use heapless::spsc::{Producer, Queue};
     ///
-    /// static PRODUCER: Mutex<RefCell<Option<Producer<'static, (), 4>>>> =
-    ///     Mutex::new(RefCell::new(None));
+    /// static PRODUCER: Mutex<RefCell<Option<Producer<'static, ()>>>> = Mutex::new(RefCell::new(None));
```
zeenix: Also here.

sgued: This is how `cargo +nightly fmt --check  -- --config format_code_in_doc_comments=true` formats this. Looking at how it renders I agree it's better to split it.

zeenix: ughh. the fmt check in CI failing now. I think we should add an fmt config like the one I use in [my projects](https://github.com/zeenix/zlink/blob/main/.rustfmt.toml) and modify CI to run nightly fmt. WDYT?

zeenix: > This is how `cargo +nightly fmt --check -- --config format_code_in_doc_comments=true` formats this.

We need `comment_width=100` as well.

sgued: CI is already using nightly with format_code_in_doc_comments.

We can add comment_width=100 and wrap_comments=true but that requires changing a lot of files. Let's do it in a separate PR.

Issue #522, opened 2024-12-13
Implement the `From` or `TryFrom` traits for `Deque`


hicklin:
It would be very useful to be able to create a `Deque` from a slice or a heapless Vec using the `from()` or `try_from()` methods.

For example:
```rust
let deq = Deque::<u8, 4>::try([1, 2, 3, 4]);
```

PR #524, opened 2024-12-16
Implemented TyrFrom for Deque from slice.

hicklin:
Fixes #522.

Ideally, this PR is updated after PR #521 is merged to simplify the tests.

This PR implements the `TryFrom` trait for creating a `Deque` from a slice.

Note the use of `unsafe` for copying all bytes from the slice to the `Deque` buffer after ensuring that the `Deque` buffer has enough space.

Also note the use of `ManuallyDrop` to ensure that any heap memory referred to by the element contents is not dropped at the end of this method since the elements in the `Deque` buffer will be pointing to it.

PR #591, opened 2025-08-18
 Implements TryFrom for Deque from array

hicklin:
Previous reviews and comments can be found in [pull request 524](https://github.com/rust-embedded/heapless/pull/524).

Fixes https://github.com/rust-embedded/heapless/issues/522.

This PR implements the TryFrom trait for creating a Deque from a slice.

Note the use of unsafe for copying all bytes from the slice to the Deque buffer after ensuring that the Deque buffer has enough space.

Also note the use of ManuallyDrop to ensure that any heap memory referred to by the element contents is not dropped at the end of this method since the elements in the Deque buffer will be pointing to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Implement `Default` for `CapacityError`.
 - Implement `defmt::Format` for `CapacityError`.
+- Implement `TryFrom` for `Deque` from array.
 
 ## [v0.9.1] - 2025-08-19
 
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -33,14 +33,14 @@
 //! }
 //! ```
 
+use crate::vec::{OwnedVecStorage, VecStorage, VecStorageInner, ViewVecStorage};
+use crate::CapacityError;
 use core::cmp::Ordering;
 use core::fmt;
 use core::iter::FusedIterator;
 use core::marker::PhantomData;
-use core::mem::MaybeUninit;
+use core::mem::{ManuallyDrop, MaybeUninit};
 use core::{ptr, slice};
-
-use crate::vec::{OwnedVecStorage, VecStorage, VecStorageInner, ViewVecStorage};
 
 /// Base struct for [`Deque`] and [`DequeView`], generic over the [`VecStorage`].
 ///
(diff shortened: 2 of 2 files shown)


zeenix (dismissed):
LGTM, thanks.

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Deque::{swap, swap_unchecked, swap_remove_front, swap_remove_back}`.
 - Make `String::from_utf8_unchecked` const.
 - Implemented `PartialEq` and `Eq` for `Deque`.
+- Implemented `TryFrom` for `Deque` from slice.
```
sgued:
```suggestion
- Implemented `TryFrom` for `Deque` from array.
```

Quick terminology to explain why:
- `[T; N]` is an array
- `[T]` is a slice
- `&[T]` (or the `mut` version`) is a (mutable) reference to a slice, often also called just a slice or a mutable slice.

.gitignore
```diff
@@ -1,4 +1,4 @@
 **/*.rs.bk
 .#*
 Cargo.lock
-target/
+target/
```
sgued: Can you pleas squash the commits and remove this change to the gitignore? I prefer to keep the git history clean.

src/deque.rs
```diff
@@ -999,11 +999,54 @@ impl<T: PartialEq, const N: usize> PartialEq for Deque<T, N> {
...
+
+    fn try_from(value: [T; NS]) -> Result<Self, Self::Error> {
+        if NS > ND {
+            return Err(CapacityError);
+        }
+
+        let mut deq = Self::default();
+        let value = ManuallyDrop::new(value);
+
+        if size_of::<T>() != 0 {
```
sgued: What happens if this check is removed? I don't understand why it's necessary. If the size of `T` is zero shouldn't `copy_nonoverlapping` still be fine?

src/deque.rs
```diff
@@ -1545,4 +1588,26 @@ mod tests {
...
+    #[test]
+    fn try_from_array() {
+        // Array is too big error.
+        assert!(matches!(
+            Deque::<u8, 3>::try_from([1, 2, 3, 4]),
+            Err(CapacityError)
+        ));
+
+        // Array is at limit.
+        assert!(Deque::<u8, 3>::try_from([1, 2, 3]).unwrap().is_full());
```
sgued: Please also test that the contents of the deque are the expected values, and that the contents are correctly dropped only once, after the deque is dropped, or if the deque is too small (you can use the `droppable` macro from the `test_helpers` for that).

src/deque.rs
```diff
@@ -999,11 +999,54 @@ impl<T: PartialEq, const N: usize> PartialEq for Deque<T, N> {
...
+    ///
+    /// let deq1 = Deque::<u8, 5>::try_from([1, 2, 3]).unwrap();
+    /// let mut deq2 = Deque::<u8, 5>::new();
+    /// deq2.push_back(1).unwrap();
+    /// deq2.push_back(2).unwrap();
+    /// deq2.push_back(3).unwrap();
+    ///
+    /// assert_eq!(deq1, deq2);
+    /// ```
+    type Error = CapacityError;
```
sgued: It might make sense to change it to `(CapacityError, [T; NS])` so that error recovery can still use the values from the array.

.gitignore
zeenix: > I prefer to keep the git history clean.

I do too and I was going to ask for the same but I didn't object cause we can always squash merge at the end anyway.

@hicklin if you're going to squash the commits, I'd suggest you do a rebase on master this time (instead of merging it).

src/deque.rs
hicklin: This was not needed. I removed it and added a test for it [here](https://github.com/rust-embedded/heapless/pull/591/commits/fbea745ae2f05bd819c0c3f47d66da41599d7b5e).

CHANGELOG.md
hicklin: Addressed [here](https://github.com/rust-embedded/heapless/pull/591/commits/fbea745ae2f05bd819c0c3f47d66da41599d7b5e).

src/deque.rs
hicklin: Added all of these tests [here](https://github.com/rust-embedded/heapless/pull/591/commits/9a401eaca57e1012f59a3e80a4b0255d4f4ba042).

src/deque.rs
hicklin: I like this. I wan considering `type Error = [T; NS]` but I think this its useful to know what the error was, even if in this case there is only one possible error. See addition [here](https://github.com/rust-embedded/heapless/pull/591/commits/6e104c37d3d361413723d9101a82826bd4310a3f).

.gitignore
hicklin: I'm going to leave this until all comments have been addressed and we are ready to merge. Also happy for you to squash merge the PR when its ready.

sgued (dismissed):
LGTM. Needs squashing before merging and removing the unnecessary change to `.gitignore`.

.gitignore
hicklin: @sgued, @zeenix Are we ready to merge this? Are you going to squash during the merge or are you waiting for me to squash it? I think it would be better to squash in github as this will preserve the commits and discussion on the PR page for posterity. This behaviour can be enforced in the repository settings page.

zeenix: Sorry, forgot. :)

zeenix: Oh, actually you never remove the unnecessary change to .gitignore that @sgued pointed out. Please do that.

zeenix: and since you're going to do that anyway, might as well just rebase (not merge) on main and squash, before force pushing to the PR branch.

hicklin: @zeenix changes to `.gitignore` have been reverted and all commits have been squashed.

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Deque::{swap, swap_unchecked, swap_remove_front, swap_remove_back}`.
 - Make `String::from_utf8_unchecked` const.
 - Implemented `PartialEq` and `Eq` for `Deque`.
+- Implemented `TryFrom` for `Deque` from array.
```
sgued: Sorry this also needs to be moved up to the "unreleased" section

hicklin: Yeah. I was doing that just now. Will be updated soon.

zeenix (approved):
In the future, please also add details to the git commit message body, not only the PR. That's a very good practice in general, so all details are always available offline.

PR #593, opened 2025-08-19
Fix failing testcase on VxWorks

### LINKED ISSUE #592: IndexMap testcase causes a stackoverflow on VxWorks
Hi,

I was running the testcases for this crate on VxWorks and I have noticed that the `almost_filled_map` testcase causes a stack overflow to occur on VxWorks. Reducing [MAP_SLOTS](https://github.com/rust-embedded/heapless/blob/main/src/index_map.rs#L1746) from `4096` to `1024` fixes this issue.

hax0kartik:
Reduce the number of slots that are created from 4096 to 1024 as too many slots can cause a stack overflow.

Closes #592

### DIFF
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -1742,8 +1742,11 @@ mod tests {
 
     // tests that use this constant take too long to run under miri, specially on CI, with a map of
     // this size so make the map smaller when using miri
-    #[cfg(not(miri))]
+    #[cfg(not(any(miri, target_os = "vxworks")))]
     const MAP_SLOTS: usize = 4096;
+    // Reduce number of slots as too many slots can cause a stack overflow on VxWorks.
+    #[cfg(target_os = "vxworks")]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - CI now uses flags specified in `Cargo.toml` for `rustdoc` tests.
 - Fixed clippy lints.
+- Fix index_map testcase causing a stack overflow on VxWorks.
```
zeenix: This doesn't concern users so we don't need it in the changelog IMO.

hax0kartik:
> LGTM otherwise.

Forced push to fix, thanks for the review!

E: A check is failing due to the missing change to Changelog, but I think that shouldn't be a problem.

zeenix (approved):

PR #597, opened 2025-08-23
Allow setting visibility of the pool struct in the `box_pool!` macro

zeenix:
While I think this is a good change, unfortunately it's a breaking change and hence keeping it in draft for now until the next API break (perhaps 1.0).


sgued:
Why is it a breaking change?

zeenix:
> Why is it a breaking change?

Because currently the struct is explicity declared `pub` by the macro. After this change, the default will not be `pub`. I thought about making the macro keep the existing behaviour if the caller doesn't specify any visibility but in Rust we don't have a keyword for explicitly marking any API private, so how would the caller ask for the struct to be private? :thinking:

PR #600, opened 2025-08-27
Remove duplicate documentation line

qbradley:
Arc was added twice, evidently a copy-paste bug.

It does not appear that anything else is missing, so I just removed the duplicate line.


reitermarkus:
Thanks!

PR #603, opened 2025-09-03
Derive defmt::Format for CapacityError

phycrax:


zeenix:
I would never understand this new trend of closing a PR, just to create a new one for exactly the same purpose. It is super annoying as the you end up with extra notifications. There is never a good reason for this. You can always `git push --force` if you really need to-rewrite history.

phycrax:
> I would never understand this new trend of closing a PR, just to create a new one for exactly the same purpose. It is super annoying as the you end up with extra notifications. There is never a good reason for this. You can always `git push --force` if you really need to-rewrite history.

It's not the exact thing. This does not compile because of the naming clash between defmt crate and defmt module. If yu are so bothered by it, you could add a contribution guide.

zeenix:
> It's not the exact thing.

The purpose is **exactly** the same. The intended change you wanted to make was **exactly** the same.

> This does not compile because of the naming clash between defmt crate and defmt module.

1. That's no reason to recreate the same PR. Just push the fix to the PR branch. There is no reason for spamming for fixing your PR.
1. You could have tried to compile locally first before submitting in the first place.

> If yu are so bothered by it, you could add a contribution guide.

Please don't make this about me. You spammed all the maintainers for absolutely no good reason. I'd assume basic stuff like "Don't spam" would be obvious and doesn't need to be explicitly mentioned.

If this served **any** useful purpose, then you might have a point but it absolutely doesn't.

PR #604, opened 2025-09-03
Implement defmt::Format for CapacityError

phycrax:


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Make MSRV of 1.87.0 explicit.
 
 - Implement `Default` for `CapacityError`
+- Implement `defmt::Format` for `CapacityError`
```
zeenix: Could you please add the missing `.` here and on the previous line (I know you didn't add that but let's fix it) while we're at it.

phycrax:
> LGTM otherwise but please in future, don't open multiple PRs for the same exact thing. 🙏

One is for deriving and one is for implementing the trait. For me it does not mean the same thing. I will take this into consideration for future contributions though.

zeenix:
> One is for deriving and one is for implementing the trait.

The effect is **exactly** the same, the *title* of the PR and Changelog entry is **exactly** the same. Just changing the approach slightly doesn't make it a different change.

> I will take this into consideration for future contributions though.

That's your right if instead of learning from your mistakes, you want to take it personally and punish the users of the crate by depriving them of your contributions.

Issue #605, opened 2025-09-05
Possible UB in the `pool::UnionNode<T>` and `pool::Node` types


Mortano:
As part of #562 I am working through the usage of `unsafe` code and I think I uncovered some UB in the `pool` module that affects `Arc<T>` and `Box<T>`. I would like it if someone could verify this as I am not entirely sure I understand the language documentation correctly. 

There are two places which IMO produce UB, both relating to the usage of the `UnionNode<T>` type:

## Obtaining an invalid reference to a union field in `UnionNode<T>`

First, [this line](https://github.com/rust-embedded/heapless/blob/f974fc129c35651283247b63361656e0b9ab564b/src/pool/treiber.rs#L56) right here in the implementation of the `Node` trait for `UnionNode<T>`:

```rust
impl<T> Node for UnionNode<T> {
    type Data = T;

    fn next(&self) -> &AtomicPtr<Self> {
        unsafe { &self.next } // UB if `self.next` is not initialized, but we can't know from within this method
    }

    //...
}
```

As per the Rustonomicon, a reference pointing to invalid data is instant UB (see [here](https://doc.rust-lang.org/nomicon/what-unsafe-does.html?utm_source=chatgpt.com)). To get a valid reference to the `self.next` field, it must be properly initialized. Looking at e.g. the `arc` module, the nodes come from the `ArcBlock<T>` type, which is initialized like this:

```rust
impl<T> ArcBlock<T> {
    /// Creates a new memory block
    pub const fn new() -> Self {
        Self {
            node: UnionNode {
                data: ManuallyDrop::new(MaybeUninit::uninit()), // TODO I think we want to initialize `next` instead of `data`!
            },
        }
    }
}
```

Initializing `data` does not guarantee that `next` is also properly initialized, so I think the fix (as I put in the comments) would be to initialize `self.next` instead in `ArcBlock::new()`. As I see, we never need `data` to be initialized as its purpose is to be pushed into the stack of the `ArcPoolImpl` type, and once we pull it out of the stack we make sure `data` is initialized through a `ptr::write`.

The same should be done for `BoxBlock::new()` in the `Box` module. IMO it would also require to make the `next()` and `next_mut()` methods of the `Node` trait `unsafe`, as we otherwise cannot guarantee that obtaining a reference to the `self.next` field is sound.

## Wrong assumptions on the memory layout of `UnionNode<T>`

[...]

PR #606, opened 2025-09-08
Fix memory layout of UnionNode<T> in pool module

Mortano:
This PR adds the missing `#[repr(C)]` for the `UnionNode<T>` type in the `pool` module, as mentioned in #605


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - CI now uses flags specified in `Cargo.toml` for `rustdoc` tests.
 - Fixed clippy lints.
+- The `UnionNode<T>` in `pool` now has the correct memory layout for reinterpreting it as a pointer to `T`
```
sgued: This is not a publicly visible change, readers of the changelog don't need to worry that it may be impacting their code.

```suggestion
- Fixed the memory layout of the internal `UnionNode<T>` type, fixing possible Undefined Behaviour.
```

sgued (commented):
Thanks for the PR. One nit on the changelog but other than that it's good!

CHANGELOG.md
Mortano: Makes sense, thanks for pointing that out

sgued (approved):

PR #609, opened 2025-09-17
feat: Arc strong_count and get_mut public functions

Hennzau:
hi, I don't really know why the public API did not provide those functions before so I added them.

Please tell me if I'm wrong and we can't do that like that

### DIFF
--- a/src/pool/arc.rs
+++ b/src/pool/arc.rs
@@ -218,6 +218,21 @@ where
         &mut *ptr::addr_of_mut!((*this.node_ptr.as_ptr().cast::<ArcInner<P::Data>>()).data)
     }
 
+    /// Returns the number of strong (`Arc`) pointers to this allocation
+    pub fn strong_count(this: &Self) -> usize {
+        this.inner().strong.load(Ordering::SeqCst)
+    }
+
+    /// Returns a mutable reference to the inner data if there are no other `Arc` pointers to the
+    pub fn get_mut(this: &mut Self) -> Option<&mut P::Data> {
+        if Self::strong_count(this) == 1 {
+            // SAFETY: we just checked that the strong count is 1
+            Some(unsafe { Self::get_mut_unchecked(this) })
+        } else {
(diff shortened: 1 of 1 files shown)


src/pool/arc.rs
```diff
@@ -218,6 +218,21 @@ where
         &mut *ptr::addr_of_mut!((*this.node_ptr.as_ptr().cast::<ArcInner<P::Data>>()).data)
     }
 
+    /// Returns the number of strong (`Arc`) pointers to this allocation
+    pub fn strong_count(this: &Self) -> usize {
+        this.inner().strong.load(Ordering::SeqCst)
```
sgued: Why `SeqCst` and not `Relaxed` ? This is a public API that has no use besides hinting.

src/pool/arc.rs
```diff
@@ -218,6 +218,21 @@ where
...
+
+    /// Returns a mutable reference to the inner data if there are no other `Arc` pointers to the
+    pub fn get_mut(this: &mut Self) -> Option<&mut P::Data> {
+        if Self::strong_count(this) == 1 {
+            // SAFETY: we just checked that the strong count is 1
+            Some(unsafe { Self::get_mut_unchecked(this) })
+        } else {
+            None
+        }
+    }
```
sgued: This needs to be loaded with `Acquire` like in the standard library (if `strong_count` is downgraded to `Relaxed`).

sgued (changes requested):
What's the purpose of having just this API?

I think it would need to also have the other methods that allow for manipulating the count manually such as `as_ptr` and `from_raw`.

Hennzau:
> What's the purpose of having just this API?
> 
> I think it would need to also have the other methods that allow for manipulating the count manually such as `as_ptr` and `from_raw`.

Yeah you're right I will add those methods then! (FYI, I needed these methods, this is why I forked, but indeed it's not complete)

PR #610, opened 2025-09-19
Remove implementing Deref<Target = [T]> for HistoryBuf

sirhcel:
There is already `as_slice()` which gives access to the underlying backing array where the elements are not in the order of writing.

With automatic dereferencing, functions like `iter()` or `windows()` are automatically available for `HistoryBuf` instances and are returning items in an order which does not reflect the write order and will likely surprise users which did not carefully read all the documentation for `HistoryBuf` like me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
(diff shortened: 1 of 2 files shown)


sirhcel:
There is more potential for streamlining the interface of `HistoryBuf` and #611 sets out to tackle these.

jamesmunns:
Probably worth noting this is a breaking change, as removal of the `Deref` impl is semver relevant.

sgued:
LGTM. This is a breaking change that should be included in the next breaking change release. Added as sub-issue: https://github.com/rust-embedded/heapless/issues/613

sirhcel:
Rebased onto current main for resolving merge conflicts.

PR #611, opened 2025-09-19
Streamline HistoryBuf interface

sirhcel:
An addendum to #610 for streamlining the interface of `HistoryBuf` even further. I'm opening this PR with two proposals for starting a discussion.

### DIFF
--- a/src/history_buf.rs
+++ b/src/history_buf.rs
@@ -886,6 +886,17 @@ mod tests {
     }
 
     #[test]
+    fn fmt() {
+        let mut buffer: HistoryBuf<u8, 4> = HistoryBuf::new();
+        buffer.extend([0, 0, 1, 2, 3, 4]);
+
+        // Check that the unordered slice representation is not ordered by accident but that the
+        // debug representation is ordered.
+        assert_ne!(buffer.as_unordered_slice(), [1, 2, 3, 4]);
+        assert_eq!(format!("{buffer:?}"), "[1, 2, 3, 4]");
+    }
+
+    #[test]
     fn ordered() {
         // test on an empty buffer
         let buffer: HistoryBuf<u8, 6> = HistoryBuf::new();
--- a/src/history_buf.rs
+++ b/src/history_buf.rs
@@ -618,7 +618,7 @@ where
     T: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        <[T] as fmt::Debug>::fmt(self.as_unordered_slice(), f)
+        f.debug_list().entries(self.oldest_ordered()).finish()
     }
 }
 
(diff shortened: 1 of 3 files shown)


sgued:
LGTM, but is a breaking change like https://github.com/rust-embedded/heapless/pull/610. Added to https://github.com/rust-embedded/heapless/issues/613

src/history_buf.rs
```diff
@@ -572,18 +591,10 @@ impl<T, S: HistoryBufStorage<T> + ?Sized> Drop for HistoryBufInner<T, S> {
...
-    }
-}
-
 impl<T, S: HistoryBufStorage<T> + ?Sized> AsRef<[T]> for HistoryBufInner<T, S> {
     #[inline]
     fn as_ref(&self) -> &[T] {
-        self
+        self.as_unordered_slice()
     }
 }
```
max-heller: `AsRef<[T]>` where the `[T]` is unordered also feels like a footgun, is this something that could be removed or is there an important reason to keep it?

src/history_buf.rs
```diff
@@ -592,7 +603,7 @@ where
     T: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        <[T] as fmt::Debug>::fmt(self, f)
+        <[T] as fmt::Debug>::fmt(self.as_unordered_slice(), f)
     }
 }
```
max-heller, line 616: Could we make `Debug` format the elements in order? Printing `HistoryBuf` as the underlying slice without including the write head is confusing

src/history_buf.rs
sirhcel: Yes, I see that this is not what I would expect from `AsRef<[T]>`. To my understanding, the result was already unordered before and I did not think about that when just getting the return type into the right shape.

Well, I am in favor of removing this trait implementation as I don't see a way in the sense of the low/neglect-able runtime costs `AsRef` suggests.

src/history_buf.rs
sirhcel: This sounds pretty reasonable to me. Added this with [9b2d27a](https://github.com/rust-embedded/heapless/pull/611/commits/9b2d27a35536549911904d559a58951abeec03ce).

sirhcel:
Rebased onto current main branch for conflict resolution for #610 and cleaned up renaming `write` to `push` for the latest additions to the tests.

sirhcel:
Fixed changelog entries messed up with my last rebase.

src/history_buf.rs
sirhcel: While at it, i removed implementing `AsRef<[T]>` with [b82c603](https://github.com/rust-embedded/heapless/pull/611/commits/b82c60372a256e29b8f618a7a5589c29008b9fe5).

Issue #562, opened 2025-04-26
Document all `unsafe` blocks.


reitermarkus:
All `unsafe` blocks should be documented and be scoped as closely as possible.

After all are documented, the `clippy::undocumented_unsafe_blocks` lint should be enabled.

Mortano:
I should be able to help with that, if the help is still wanted. Are there any contribution guidelines for first-time contributors?

zeenix:
> I should be able to help with that, if the help is still wanted.

I'm pretty sure it is. :) Thanks.

> Are there any contribution guidelines for first-time contributors?

None really. We do have a CoC (see `CODE_OF_CONDUCT.md` file) but that's pretty normal these days.

Mortano:
Ok I'll start working on this then :)

Mortano:
Do you have a preference as to how detailed the documentation should be? Unsafe Rust is a tricky beast and many of the unsafe functions that are used have quite a long list of safety requirements. The way I'm approaching this now is to go through these lists and verify that each requirement is met and document that accordingly, but that makes the `SAFETY` comments quite long in some cases. Here is one example in `DequeInner::make_contiguous`, which uses a lot of pointer arithmetics and raw memcpy:
```rust
...
} else if free >= back_len {
    // there is enough free space to copy the tail in one go,
    // this means that we first shift the head forwards, and then
    // copy the tail to the correct position.
    //
    // from: FGH....ABCDE
    // to:   ...ABCDEFGH.

    // SAFETY: Explaining it visually is easier:
    // step 1:   FGH....ABCDE
    //              |   |
    //              1   2
    //           |__|___|___|
    //  back_len---^  ^   ^---- front_len
    //              free
    // step 2:   FGHABCDE....
    //           |       |
    //           3       4
    // Pointers: 1: `buffer_ptr.add(self.back)`
    //           2: `buffer_ptr.add(self.front)`
    //           3: `buffer_ptr`
    //           4: `buffer_ptr.add(self.back + front_len)`
    // All pointers are clearly within the underlying buffer and are properly aligned, meeting the first safety
    // requirement for `ptr.add` and the `ptr::copy...` variants.
    // In step 1 above, we see that pointer 1 is always `<=` than pointer 2, so the copy stays within the bounds
    // of the buffer, but it might overlap, hence the usage of `ptr::copy`
    // In step 2 above, we see that the target memory region does not overlap with the source region, as pointer 4
[...]

PR #612, opened 2025-09-19
Fix UB in `pool` module when using `ArcBlock<T>` and `BoxBlock<T>`

Mortano:
This fixes UB when using the `ArcBlock<T>` and `BoxBlock<T>` types together with the `Stack` type, as it occurs in the `ArcPoolImpl` and `BoxPoolImpl` types. The underlying cause for the UB was a wrongly initialized `UnionNode`, as was discussed in #605 . This also required that the functions of the `Node` trait are `unsafe`, but this should not be a breaking API change as that trait is never exported from the `pool` module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `ExactSizeIterator` to `vec::IntoIter`, `deque::IntoIter`, `index_map::IntoIter` and `linear_map::IntoIter`.
 - Added `DoubleEndedIterator` to `vec::IntoIter` and `deque::IntoIter`.
 - Deprecate `mpmc` (see [#583](https://github.com/rust-embedded/heapless/issues/583#issuecomment-3469297720))
+- Fixed initialization of the `ArcBlock<T>` and `BoxBlock<T>` types, fixing possible Undefined Behavior
 
 ## [v0.9.1] - 2025-08-19
 


sgued (approved):
LGTM

Thanks a lot for the audit of the unsafe code. Already yielding results!

src/pool/treiber.rs
```diff
@@ -50,14 +64,27 @@ pub union UnionNode<T> {
...
+        }
+    }
+}
+
 impl<T> Node for UnionNode<T> {
     type Data = T;
 
-    fn next(&self) -> &AtomicPtr<Self> {
+    unsafe fn next(&self) -> &AtomicPtr<Self> {
+        // SAFETY: Caller ensures that `self.next` is properly initialized
```
zeenix, line 84: `next` is not public though. 🤔

src/pool/treiber.rs
```diff
@@ -38,10 +40,22 @@ where
...
     type Data;
 
-    fn next(&self) -> &AtomicPtr<Self>;
+    /// Returns a reference to the atomic pointer that stores the link to the next `Node`
+    ///
+    /// # Safety
+    ///
+    /// It must be valid to obtain a reference to the next link pointer, e.g. in the case of
+    /// `UnionNode`, the `next` field must be properly initialized when calling this function
+    unsafe fn next(&self) -> &AtomicPtr<Self>;
```
zeenix, line 50: I'm not too sure about making a trait method unsafe if there could be safe impls of it but (as I wrote in the other comment), I'm not sure even the `UnionNode::next` is itself unsafe.

src/pool/treiber.rs
```diff
@@ -50,14 +64,27 @@ pub union UnionNode<T> {
     pub data: ManuallyDrop<T>,
 }
 
+impl<T> UnionNode<T> {
+    /// Returns a new `UnionNode` that does not contain data and is not linked to any other nodes.
+    /// The return value of this function is guaranteed to have the `next` field properly initialized.
+    /// Use this function if you want to insert a new `UnionNode` into a linked list
```
zeenix: Right, so this ensure that `UnionNode::next` is always correctly initialzed (given that `next` field is private) and hence the `next` method is not unsafe then? 🤔

src/pool/treiber.rs
Mortano: I think it doesn't matter because it is a soundness issue. Line 83 does `unsafe { &self.next }`, which is `unsafe` for a reason (you must only take a reference to a properly initialized field). If I make the `next` function safe, I can't always guarantee that this unsafe operation is in fact safe. We could create a new `UnionNode` with the `data` field initialized somewhere and call `.next()` on it and instantly get UB, without ever explicitly calling any `unsafe` code. So the only way to ensure that we adhere to the safety requirements of `unsafe { &self.next }` is to escalate and make the calling function `unsafe` as well. Then the caller has to ensure that what they are doing is safe. 

Even if all our code today did the right thing (which it didn't, hence this PR), it would be easy to get things wrong in the future, so I think while this does require a bit more typing, it is much safer.

zeenix: That's not how `unsafe` is typically used in the Rust world. You only make the method that introduced unsafety/unsoundness as `unsafe` and not the rest of the API that becomes unsafe/unsound because of it. If the `Node` was initialized correctly, `next` impl would be completely safe and sound.

E.g [`std::str::from_utf8_unchecked`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.from_utf8_unchecked) gives you `&str` and it's marked as `unsafe` for a good reason but after that, all the API you use on the `&str`, is declared safe even though a lot of it cane be unsafe because of it. The same goes for `String::from_utf8_unchecked` and you'll find this a general pattern in std and many other Rust libraries.

Mortano: Ok I see, but isn't the case here different because `UnionNode` has two valid states (the `next` field initialized vs. the `data` field initialised)? With your `str` example, it is pretty clear that there is one way to break the API guarantees, but with `UnionNode` every change in the variant of the node will either cause the `Node` trait to be invalid to use or not. I am simply uncomfortable with the idea of "The Node trait is only *sometimes* valid/safe to use, depending on which variant it holds". If this is idiomatic Rust and you prefer this I am happy to change my PR to only include the proper initializing of the Node :)

sgued: The difference with `from_utf8_unchecked` is that constructing a `str` from invalid `utf-8` is already UB. Creating a `UnionNode` in any state is a valid state, but handling it can lead to UB.

Any operation that needs to mutate or read a `union` will need to be `unsafe`.

src/pool/treiber.rs
sgued: No, because if something is written to the `data` field, and `next` is then called on it, it makes it unsafe.

src/pool/treiber.rs
Mortano: To add to this: The following safe Rust code (prior to this PR) will lead to UB and that violates the guarantees of safe Rust:
```rust
let node = UnionNode {
    data: ManuallyDrop::new(42),
};
let next_ptr = node.next(); // UB, next() assumes that the node holds variant `next` but it holds `data`
```

To me the key point is that creating a union and writing union fields both are safe in Rust (see [here](https://doc.rust-lang.org/reference/items/unions.html#r-items.union.fields.write-safety)). Since we hide the unsafe *read* of the union field inside the safe `next` method of the `Node` trait, we now have an API that allows creating UB with 100% safe Rust. 

If you are unhappy with the `unsafe fn` on the trait the other option I see would be to hide all union writes as well as the creation of a new `UnionNode` behind `unsafe` functions and document the invariants of these functions. But I think this is less usable than simply stating that `Node::next` is an `unsafe` function. In the end that is a matter of personal preference and I am happy to adhere to whatever you envision for this crate.

zeenix: > The difference with `from_utf8_unchecked` is that constructing a `str` from invalid `utf-8` is already UB.

Not really. AFAICT nothing can go wrong if you don't do anything with the created `&str` after creating it, so there is no UB at `from_utf8_unchecked` time itself, yet it is the part that's declared unsafe. The important point here is that user needs to be informed about the unsafety where they introduce it (and preferably they can do something about it), not where it actually happens.

`unsafe` is generally a way to tell the caller to be careful when they make `unsafe` calls. Notifying them at `node.next()` time is useless since at that point, they've no way of ensuring that the call is safe. Hence, why I believe that any unsafe initialisation of the UnionNode should be declared unsafe but not the methods, that can be unsafe/unsound only because of such initialisation.

> To me the key point is that creating a union and writing union fields both are safe in Rust (see [here](https://doc.rust-lang.org/reference/items/unions.html#r-items.union.fields.write-safety)). Since we hide the unsafe _read_ of the union field inside the safe `next` method of the `Node` trait, we now have an API that allows creating UB with 100% safe Rust.

That's the same as the `from_utf8_unchecked`. You create a state which will blow-up later when you operate on it using the safe API.

Another "issue" I see is that the method is part of the `Node` trait (not the `UnionNode` alone) so we end up declaring it unsafe for all impls, which is not nice unless we're sure that it will always be (potentially) unsafe for all impls.

> In the end that is a matter of personal preference

Ok, we can agree on that. :)

> and I am happy to adhere to whatever you envision for this crate.

I wouldn't go as far as calling it a vision but yes, I think it'd be better if you remove the bits I disagree with so we can merge the actual fix quickly and release. Feel free to split the removed bits into a separate PR and we can discuss them there in a relaxed manner.

sgued: > Not really. AFAICT nothing can go wrong if you don't do anything with the created &str after creating it, so there is no UB at from_utf8_unchecked time itself,

The compiler's deny_by_default lints says the oposite: 
```rust
drop(core::str::from_utf8_unchecked(&[0xFF]));
```
Fails to compile with 
```
error: calls to `std::str::from_utf8_unchecked` with an invalid literal are undefined behavior
```

We can't make creating of a `UnionNode` `unsafe` because creating a union is a safe operation, and any mutation would need to be `unsafe` because it could leave the union in an "invalid state".

`UnionNode` being in a `data` state is also a valid in some cases. Requiring `next` to always be callable would mean it's not a valid state.

`push` is already unsafe, it is the one that expects its arguments to be in a certain state to work. It can require that calling `.next` is valid for the arguments passed to it.

I think having discussions about the meaning of unsafe don't make that much sense. It's all internal APIs and what matters here is only that usage of `unsafe` is documented and justified by a comment.

Mortano: @sgued I think we are in agreement then, as this is what I proposed with this PR? But @zeenix thinks differently? So how do we reach consensus here? 

I am also asking because I am still working on #562 and am mostly done with that, but I know that my approach to `unsafe` caused me to introduce some (internal) changes to what is marked as `unsafe` and what not, which could lead to more discussions like this one here. My time however is limited (as is yours probably) and I get the impression that our arguments don't convince either of us. We could of course discuss this on the PR for #562 once I open it, but I neither want to create unnecessary work for you due to discussions that lead nowhere nor do I want to compromise on my own standards and release things that IMO are incorrect. Please don't feel criticised by this, it is simply my effort to gauge if/how I can contribute.

zeenix: > The compiler's deny_by_default lints says the oposite:

Sorry, not the best example then. However, the point remains. You can cause undefined behaviour later, with safe API.

> But @zeenix thinks differently? So how do we reach consensus here?

As @sgued said, it's internal API anyway and it's very small. Besides I don't feel as strongly about it as it might seem. :laughing: So let's go with what you have.

sgued:
There was a rebase conflict that caused the CI to fail. Thanks merge queues for avoiding a broken main. @zeenix can you please re-review ? Since I forced push it doesn't want me to merge alone.

PR #617, opened 2025-10-01
Add truncate and retain(_mut) to Deque

nullstalgia:
Howdy!

As I was working on simplifying a crate's logic to fit in a `no_std` context, I found that `Deque` didn't have some of the methods I was using from `alloc`'s implementation.

So I took the time to port them over! Apologies if the comments in `truncate` are a bit much, it was mostly to help me wrap my head around how the data structure itself used its private variables. 

If you spot an error in my logic or if I missed something, please let me know!

### DIFF
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -811,6 +811,184 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
             inner: start.iter_mut().chain(end),
         }
     }
+
+    /// Shortens the deque, keeping the first `len` elements and dropping
+    /// the rest.
+    ///
+    /// If `len` is greater or equal to the deque's current length, this has
+    /// no effect.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::Deque;
+    ///
+    /// let mut buf: Deque<_, 5> = Deque::new();
+    /// buf.push_back(5);
+    /// buf.push_back(10);
+    /// buf.push_back(15);
+    /// buf.truncate(1);
+    /// assert_eq!(buf.make_contiguous(), [5]);
+    /// ```
+    pub fn truncate(&mut self, len: usize) {
+        /// Runs the destructor for all items in the slice when it gets dropped (gracefully or
+        /// during unwinding).
+        struct Dropper<'a, T>(&'a mut [T]);
+
+        impl<'a, T> Drop for Dropper<'a, T> {
+            fn drop(&mut self) {
+                unsafe {
+                    ptr::drop_in_place(self.0);
+                }
+            }
+        }
+
+        // Safety:
+        // * Any slice passed to `drop_in_place` is valid; the second case has
+        //   `len <= front.len()` and returning on `len > self.storage_len()` ensures
+        //   `begin <= back.len()` in the first case
+        // * Deque front/back cursors are moved before calling `drop_in_place`,
+        //   so no value is dropped twice if `drop_in_place` panics
+        unsafe {
+            // If new desired length is greater or equal, we don't need to act.
+            if len >= self.storage_len() {
+                return;
+            }
+
+            let (front, back) = self.as_mut_slices();
+
+            // If `len` desires to keep elements past front's entire length,
+            // then only back's contents will need to be dropped
+            // as the two slices combined should be more than `len`.
+            if len > front.len() {
+                let begin = len - front.len();
+                let drop_back = back.get_unchecked_mut(begin..) as *mut _;
+
+                // Self::to_physical_index returns the index `len` units _after_ the front cursor,
+                // meaning we can use it to find the decremented index for `back` for non-contiguous deques,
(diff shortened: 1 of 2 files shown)


src/deque.rs
```diff
@@ -811,6 +811,189 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+        F: FnMut(&mut T) -> bool,
+    {
+        let len = self.storage_len();
+        let mut idx = 0;
+        let mut cur = 0;
+
+        // Stage 1: Check if all values can be retained.
+        while cur < len {
+            // Safety: `cur` must be under the deque's total length as per the loop condition
+            if !f(unsafe { self.get_unchecked_mut(cur) }) {
```
sgued: Why deviate from `std`'s implementation and use unsafe here? I expect the bound check introduced in the safe path to be optimized out.

src/deque.rs
```diff
@@ -1703,4 +1886,68 @@ mod tests {
...
+        thread_local! {
+            static DROPS: Cell<u32> = const { Cell::new(0) };
+        }
+
+        #[derive(Debug)]
+        struct DropIncrementor;
+        impl Drop for DropIncrementor {
+            fn drop(&mut self) {
+                DROPS.set(DROPS.get() + 1);
+            }
```
sgued: There is a `droppable!` macro in `src/test_helpers.rs`, please use that.

src/deque.rs
```diff
@@ -1703,4 +1886,68 @@ mod tests {
...
+                } else {
+                    tester.push_back(index).unwrap();
+                }
+            }
+
+            tester.retain(|x| *x >= 10);
+
+            assert_eq!(tester.len(), 10);
+        }
+    }
```
sgued, line 2137: Please make it check for the drop count too.

src/deque.rs
```diff
@@ -1703,4 +1886,68 @@ mod tests {
         assert_eq!(deque.len(), 0);
         assert!(deque.is_empty());
     }
+
+    // Based on alloc's own VecDeque test
+    //
+    // https://github.com/rust-lang/rust/blob/fa3155a644dd62e865825087b403646be01d4cef/library/alloc/src/collections/vec_deque/tests.rs#L1086
+    //
+    // Tests that each element's destructor is called when being truncated.
```
sgued: Please test the following cases:

- Truncating an empty deque
- Truncating an continuous deque
  - leaving it unchanged
  - leaving it empty
  - leaving it half-empty
- Truncating a discontinuous deque
  - leaving it unchanged
  - leaving it discontinuous but dropping some elements
  - leaving it continuous (i.e. all the back is dropped)
    - reducing the front slice
    - not touching the front slice
  - leaving it empty

Essentially ensure that all branches of the `truncate` function are exercized.

src/deque.rs
```diff
@@ -811,6 +811,189 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+                    self.back = self.to_physical_index(len);
+                } else {
+                    // If the back cursor is "in front" of the front cursor,
+                    // we can just move it closer to truncate front's length.
+                    let back_trunc = front.len() - len;
+
+                    // This shouldn't underflow since `back` must be larger
+                    // than both `len` and `front` for this to even be reached.
+                    self.back -= back_trunc;
+                }
```
sgued: I don't understand why we can't use `to_physical_index` in both cases

sgued (changes requested):
Looking good.

I'd love to have tests covering all possible branches in the `truncate` method though.

src/deque.rs
nullstalgia: Ah, good catch. No reason why we can't, I just didn't reconsider that logic. Fixed!

src/deque.rs
nullstalgia: Fair! My reasoning came from the fact that Deque doesn't impl `ops::IndexMut` (and most things in heapless seem to not, I'm unsure how intentional of a choice that is still). I like to avoid `unwrap`s in `unsafe` blocks for a few reasons; `expect` is _fine_, but I also felt it was unnecessarily explicit? I mean, `std::VecDeque`'s `IndexMut` impl comes out to be the same as what I have now changed it to, a `get_mut()` with an `except()`.

So maybe another PR adding those `ops` impls would be in order? Unless this was an explicit choice to not allow indexing like that, I'm still new to heapless as a whole :sweat_smile:.

src/deque.rs
nullstalgia: Added more tests, thanks for the suggestions! I was surprised there wasn't more in the std, maybe I just didn't look deep enough. 

Let me know if any should be adjusted.

nullstalgia:
@sgued Let me know how you'd like me to proceed in regards to the outstanding conversations (about `IndexMut` and reviewing the new tests)!

sgued:
It looks fine sorry.
The miri issue is a nightly bug: https://github.com/rust-lang/rust/issues/147648, we can merge.

Issue #583, opened 2025-07-07
MpMcQueue can fail to push even when it's not full under contention


nodakai:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0a9dde3dac37be94314c5719726060c7

```rust
#[test]
fn test() {
    fn to_vec<T>(q: &QueueView<T>) -> Vec<T> {
        // inaccurate
        let mut ret = vec![];
        while let Some(v) = q.dequeue() {
            ret.push(v);
        }
        ret
    }
    const N: usize = 4;

    let q0 = Queue::<u8, N>::new();
    for i in 0..N {
        q0.enqueue(i as u8).expect("new enqueue");
    }
    eprintln!("start!");

    std::thread::scope(|sc| {
        for _ in 0..2 {
            sc.spawn(|| {
                for k in 0..1000_000 {
                    if let Some(v) = q0.dequeue() {
                        q0.enqueue(v).unwrap_or_else(|v| {
                            panic!("{}: q0 -> q0: {}, {:?}", k, v, to_vec(&q0))
                        });
                    }
                }
            });
        }
    });
}
```
Confirmed an implementation based on a mutex, and another based on [`ring-channel`](https://docs.rs/ring-channel/latest/ring_channel/) didn't panic

sosthene-nitrokey:
I think I see the case that can fail. Let's name the 2 threads: `A` and `B`, `B` is the one that will panic:

1. The queue is full, `enqueue_pos  = i + N` and `dequeue_pos = i`
1. `A` calls `dequeue`
2. `A` gets to the `compare_exchange` of the `dequeue` operation (reserving its own dequeuing to item `i`)
3. `B` calls `dequeue` dequeuing completely item `i + 1`.
4. `B` calls `enqueue`, tries to enqueue on the slot of `i + N`, which still contains item `i` which has not been dequeued yet (thread `A` has been parked by the scheduler and is still in the `dequeue` operation)

This is consistent with what I observe in the tests regarding the position of the `dequeue` operation that precedes the failure to `enqueue`.

The good news is that no invariant ever gets invalidated. The bad news is that I have no idea how to fix this without completely changing the implementation.
It might make sense to use an implementation that is better documented, like `crossbeam`'s `ArrayQueue`. Though it might not make the same performance tradeoffs (it certainly seems to be using a lot more code so will lead to a larger binary size).

This queue is designed for the use cases where items can be dropped when the consumer side is being too slow

[...]

PR #618, opened 2025-10-03
mpmc: add loom test for enqueue/dequeue contention

0xllx0:
Adds a `loom` permutation test for `mpmmc::Queue::enqueue` contention across threads.

Related: #583
Co-authored-by: NODA Kai @nodakai
Co-authored-by: Sosthène Guédon @sosthene-nitrokey


src/mpmc.rs
```diff
@@ -419,4 +428,81 @@ mod tests {
...
+            loom::thread::spawn(move || {
+                for k in 0..1000_000 {
+                    if let Some(v) = q1.dequeue() {
+                        q1.enqueue(v)
+                            .unwrap_or_else(|v| panic!("{k}: q0 -> q0: {v}, {:?}", to_vec(&*q1)));
+                    }
+                }
+            });
+        });
+    }
```
ghost, line 545: This [loom](https://docs.rs/loom) test passes, but is also not a direct replication of the failing test harnesses. There is currently no `loom::thread::scope` equivalent of `std::thread::scope`.

Edit: there was some work to add `loom::thread::scope` in [loom/!312](https://github.com/tokio-rs/loom/pull/312), but the author stopped working on it in May 2024.

ghost: Interestingly, adding a test using `loom::thread::scope` from the linked PR also passes.

sgued:
Thanks for working on fixing #583 

Please add `loom` first and change the implementation in a separate PR, this will make it much easier to make sur  that the tests are working as intended, and will make review much easier.

Regarding why the loom tests pass: you're misusing `loom`. You need to add  a `#[cfg(not(loom))]` to all `use core::sync::atomic*` and a `[cfg(loom)]` that instead imports the atomics from loom. You then need to pass a `--cfg loom` flag to the compiler when you run the `loom` tests (which should be `#[cfg(loom)]` too)

Here loom doesn't see any of the atomic operation so assumes all went fine.

Also please use stable `loom` . Using `Box::leak` or `Arc` is fine here.

ghost:
> Thanks for working on fixing #583

You're welcome <3

> Please add `loom` first and change the implementation in a separate PR, this will make it much easier to make sur that the tests are working as intended, and will make review much easier.

Sounds good, I just wanted to add the alternate implementation here first to make the comparison easier.

I'll create a separate PR for the alternate implementation.

I'll address the `loom` tests below.

> Regarding why the loom tests pass: you're misusing `loom`. You need to add a `#[cfg(not(loom))]` to all `use core::sync::atomic*` and a `[cfg(loom)]` that instead imports the atomics from loom. You then need to pass a `--cfg loom` flag to the compiler when you run the `loom` tests (which should be `#[cfg(loom)]` too)
> 
> Here loom doesn't see any of the atomic operation so assumes all went fine.
> 
> Also please use stable `loom` . Using `Box::leak` or `Arc` is fine here.

This definitely sounds like it requires a separate PR. I'm new to the `loom` library, so not surprising that I'm using it wrong.

Since the failing conditions can be reached by normal unit-tests, I'll probably wait to implement the `loom` tests until after we decide on the alternate implementation PR.

Edit: after a small experiment using a feature-gate for `use loom::sync::atomic` in `mpmc`, it will require some ugly feature-gates to make the queue constructors non-`const` for `loom` builds. May end up worthwhile to support permutation tests.

The `loom` changes for the current `mpmc::Queue` are not so bad, but running the tests give the error:

```
thread 'mpmc::tests::issue_583_enqueue_loom' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/loom-0.7.2/src/rt/path.rs:247:13:
Model exceeded maximum number of branches. This is often caused by an algorithm requiring the processor to make progress, e.g. spin locks.
```

This probably means the `loom` tests won't tell us anything, since both implementations rely on versions of spin-locks. I'm not going to go through the refactoring effort of making `loom` primitives work with the `crossbeam` implementation if it's throwing these kind of errors with the current `mpmc` implementation. The refactors for the `crossbeam` implementation are much more involved.

zeenix:
> > Also please use stable `loom` . Using `Box::leak` or `Arc` is fine here.
> 
> This definitely sounds like it requires a separate PR

Why? If it's just changes to unmerged changes, then it can (and should be) the same PR. Otherwise, it only creates unnecessary noise. Just modify your commits (if you have to) and force push to your branch.

ghost:
> > > Also please use stable `loom` . Using `Box::leak` or `Arc` is fine here.
> > 
> > 
> > This definitely sounds like it requires a separate PR
> 
> Why? If it's just changes to unmerged changes, then it can (and should be) the same PR. Otherwise, it only creates unnecessary noise. Just modify your commits (if you have to) and force push to your branch.

What I meant was that the `loom` stuff should be separate from the fix PR, which I've opened in #619. The `loom` tests fail for the existing implementation, and likely the port of the `crossbeam` implementation. They fail with an error about spin-locks:

```
thread 'mpmc::tests::issue_583_enqueue_loom' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/loom-0.7.2/src/rt/path.rs:247:13:
Model exceeded maximum number of branches. This is often caused by an algorithm requiring the processor to make progress, e.g. spin locks.
```

ghost:
Closing in favor of #620

PR #622, opened 2025-10-25
Pin nightly in CI

sgued:
Needs to be done until https://github.com/rust-lang/rust/issues/147648 is resolved.

Example of failed workflow: https://github.com/rust-embedded/heapless/actions/runs/18805787053/job/53659693297?pr=617


zeenix:
@sgued could you please move the issue URL to the description of the commit & PR, so it's clickable and the title isn't long?

zeenix:
LGTM. You forgot to update the commit but it's ok if we squash merge, I guess.

PR #624, opened 2025-10-30
Document #583 and deprecate `mpmc`

sgued:
This follows the wg meeting conclusions of 28/10/2025

### DIFF
--- a/src/mpmc.rs
+++ b/src/mpmc.rs
@@ -1,5 +1,34 @@
 //! A fixed capacity multiple-producer, multiple-consumer (MPMC) lock-free queue.
 //!
+//! # Deprecation
+//!
+//! <div class="warning">
+//! The current implementation of `mpmc` is marked as deprecated due to not being truly lock-free
+//! </div>
+//!
+//! If a thread is parked, or pre-empted for a long time by an higher-priority task
+//! during an `enqueue` or `dequeue` operation, it is possible that the queue ends-up
+//! in a state were no other task can successfully enqueue or dequeue items from it
+//! until the pre-empted task can finish its operation.
+//!
+//! In that case, [`enqueue`](QueueInner::dequeue) and [`dequeue`](QueueInner::enqueue)
+//! will return an error, but will not panic or reach undefined behaviour
+//!
+//! This makes `mpmc` unsuitable for some use cases such as using it as a pool of objects.
+//!
+//! ## When can this queue be used?
+//!
+//! This queue should be used for cross-task communication only when items sent over the queue
+//! can be dropped in case of concurrent operations, or when it is possible to retry
+//! the dequeue/enqueue operation after other tasks have had the opportunity to make progress.
+//!
+//! In that case you can safely ignore the warnings using `#[expect(deprecated)]`
+//! when `new` is called
+//!
+//! For more information, and possible alternative, please see
+//! <https://github.com/rust-embedded/heapless/issues/583>
+//!
+//!
 //! **Note:** This module requires atomic compare-and-swap (CAS) instructions. On
 //! targets where they're not natively available, they are emulated by the
 //! [`portable-atomic`](https://crates.io/crates/portable-atomic) crate.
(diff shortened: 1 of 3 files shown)


src/mpmc.rs
```diff
@@ -121,7 +148,36 @@ pub type Queue<T, const N: usize> = QueueInner<T, OwnedStorage<N>>;
...
+    /// <div class="warning">
+    /// The current implementation of `mpmc` is marked as deprecated due to not being truly lock-free
+    /// </div>
+    ///
+    /// If a thread is parked, or pre-empted for a long time by an higher-priority task
+    /// during an `enqueue` or `dequeue` operation, it is possible that the queue ends-up
+    /// in a state were no other task can successfully enqueue or dequeue items from it
+    /// until the pre-empted task can finish its operation.
+    ///
+    /// In that case, [`enqueue`](QueueInner::dequeue) and [`dequeue`](QueueInner::enqueue) will return an error, but will not panic or reach undefined behaviour
```
zeenix: Can we keep the comment lines also under 100 chars please?

It would be great to automate this with `comment_width = 100` in the `.rustfmt.toml` but unfortunately that's still nightly-only. In all my projects I have always used the nightly fmt for this (and 2 other nightly-only options) and we can do so in this project too if you'd be OK with that?

zeenix (changes requested):
Great stuff otherwise! :+1:

src/mpmc.rs
sgued: I fixed the lines that were touched by this PR that are longer than 100 characters.

I think we already talked about it. I don't really like it:

- it means that running just `cargo fmt` without nightly will fail
- I don't think that we should be forced to wrap comments when just one word is going over the 100 character limit
- my IDE wraps comment properly, keeping indentation and all so I never have an issue with readability
- It can force us to re-format code examples in a way that doesn't make sense for documentation.

But I'm fine with merging it if you make a PR for this

zeenix: > it means that running just cargo fmt without nightly will fail

It won't. It'll just give a warning. It'll only become an error if you explicitly ask for warnings to be treated as errors. We'll do that in the CI but we'll then be running nightly.

As I wrote I've been using this in many projects for many years without much of a problem. I'm talking about projects that get sufficient money of (new) contributions.

It's unfortunate that rustfmt is sort of unmaintained so these essential features remain to be unstable forever. 😓

---

Anyway, up to you. It's just annoying that I have to point out this inconsistency (lines in the same files should have the same max length) all the time to you. Please manage it manually if you're against automating this. 🙏

sgued: > It won't. It'll just give a warning. It'll only become an error if you explicitly ask for warnings to be treated as errors

Oh, right.

Then yeah let's go for it, just make a PR setting it up the way you usually do I'll approve it.

zeenix: > I don't think that we should be forced to wrap comments when just one word is going over the 100 character limit

Yeah, you should be. We want to keep it as consistent as possible. You're against automating this, you have to do this manually then. Sorry!

> my IDE wraps comment properly, keeping indentation and all so I never have an issue with readability

That's entirely between you and your IDE.

> It can force us to re-format code examples in a way that doesn't make sense for documentation.

It's code and it needs to follow the same rules as the rest of the code (even if the `/// ` takes a few chars away) anyway. I could be misremembering but I think there's also an unstable option for formatting the code inside doc comments. 🤔

We need to agree on a line limit and we all need to follow it to keep things consistent. The whole reason for rustfmt is that we don't need to have such debates. rustfmt has already chosen 100 chars limit for us, let's stick to that and keep things consistent.

zeenix: > Then yeah let's go for it, just make a PR setting it up the way you usually do I'll approve it.

Cool. Glad we come to an agreement here. :) I'll try to do that soon. I'm just a bit too busy with the new job right now but hopefully this weekend or next week..

zeenix (approved):

src/mpmc.rs
zeenix: Actually turns out that we're already using nightly fmt in the CI. Now I'll add the options in a PR...

PR #625, opened 2025-11-02
Add support for `into_raw` and `from_raw` for Box

Balint134:
Hi! This PR proposes adding `into_raw()` and `from_raw()` support for `Box` types, similar to what `std::boxed::Box` provides.

While implementing FFI bindings for a resource managed by a BoxPool (with create_resource() / destroy_resource() hooks), I ran into the need for APIs to obtain and reconstruct raw pointers.

I’ve been using this patch for a while in my own project, and it has worked well. It would be great if this functionality could be included upstream.

### DIFF
--- a/src/pool/boxed.rs
+++ b/src/pool/boxed.rs
@@ -85,8 +85,8 @@ use core::{
     hash::{Hash, Hasher},
     mem::{ManuallyDrop, MaybeUninit},
     ops, ptr,
+    ptr::{addr_of, addr_of_mut},
 };
-
 use stable_deref_trait::StableDeref;
 
 use super::treiber::{NonNullPtr, Stack, UnionNode};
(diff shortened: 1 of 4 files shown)


src/pool/boxed.rs
```diff
@@ -568,4 +603,57 @@ mod tests {
         assert!(once.is_ok());
         assert!(twice.is_ok());
     }
+
+    #[test]
+    fn test_into_raw_from_raw() {
```
zeenix: nit: `test_` prefix is redundant and inconsistent here.

src/pool/boxed.rs
```diff
@@ -568,4 +603,57 @@ mod tests {
...
+        pub struct MyStruct {
+            value: [u8; 64],
+        }
+
+        static DROP_WAS_CALLED: AtomicUsize = AtomicUsize::new(0);
```
zeenix: Why not AtomicBool? 🤔

src/pool/boxed.rs
```diff
@@ -180,6 +180,41 @@ where
...
+    /// Constructs a `Box<P>` from a raw pointer
+    ///
+    /// # Safety
+    ///
+    /// The `ptr` must have been allocated by the same `BoxPool` `P` that this `Box<P>` uses.
+    pub unsafe fn from_raw(ptr: *mut P::Data) -> Self {
+        // SAFETY: caller must guarantee that `ptr` is valid and was allocated by `P`
+        debug_assert!(!ptr.is_null(), "Pointer must be non-null");
+
+        // Calculate the offset of the `data` field within `UnionNode` (with the current layout tha data is at the beginning so offset is zero, but this may change in the future)
```
zeenix: Please keep comment lines under 100 chars too.

src/pool/treiber/llsc.rs
```diff
@@ -46,6 +46,14 @@ where
             inner: NonNull::from(ref_),
         }
     }
+
+    #[inline]
+    pub unsafe fn new_unchecked_from_ptr(ptr: *mut N) -> Self {
```
zeenix: `from_ptr_unchecked` would be a better name.

src/pool/treiber/cas.rs
```diff
@@ -125,6 +125,12 @@ where
         unsafe { Self::new_unchecked(initial_tag(), NonNull::from(reference)) }
     }
 
+    #[inline]
+    pub unsafe fn new_unchecked_from_ptr(ptr: *mut N) -> Self {
```
zeenix: `from_ptr_unchecked` would be a better name.

zeenix (changes requested):
meant to request changes.

src/pool/boxed.rs
Balint134: I wanted to assert the exact number of drops just to be on the safe side.

Balint134:
> Thanks! LGTM apart from the nitpicks.

Thanks @zeenix I addressed your comments and also updated the CHANGELOG.

src/pool/boxed.rs
zeenix: Ok, in that case the name should be changed so it doesn't sound like a boolean. :) `N_DROP_CALLS`?

Balint134: That's a fair point :) lemme update it real quick

Issue #626, opened 2025-11-02
FusedIterator, ExactSizeIterator, DoubleEndedIterator


therealbnut:
Hi,

Thanks for making this great library!

I came across this today, and it seems like it should be possible to implement `FusedIterator`, `ExactSizeIterator`, and `DoubleEndedIterator` for `Vec`'s `IntoIter` and probably others too without too much complexity.

I was thinking `ExactSizeIterator` at least would be useful when using these data types to extend heap allocated structures so they can allocate the correct size up front. The other traits likely have performance or functional benefits in other situations too.

What do you think? Are they intentionally not implemented yet?

zeenix:
> What do you think? Are they intentionally not implemented yet?

Sounds good to me. I think someone just haven't added these yet so please feel free to provide a PR for these.

therealbnut:
@zeenix I've thrown up a PR [#628](https://github.com/rust-embedded/heapless/pull/628)

PR #628, opened 2025-11-03
Add ExactSizeIterator, FusedIterator, and DoubleEndedIterator on IntoIter types

therealbnut:
Addresses https://github.com/rust-embedded/heapless/issues/626 by implementing missing traits on the `IntoIter` types. Specifically:
- Added `FusedIterator` to `vec::IntoIter`, `deque::IntoIter`, `index_map::IntoIter` and `linear_map::IntoIter`.
- Added `ExactSizeIterator` to `vec::IntoIter`, `deque::IntoIter`, `index_map::IntoIter` and `linear_map::IntoIter`.
- Added `DoubleEndedIterator` to `vec::IntoIter` and `deque::IntoIter`.

I didn't implement `DoubleEndedIterator` on the Map types, as their reference-based iterators didn't have it. Even if there was reference implementations, there wasn't a way to do it that would make sense without changing existing behaviour and/or the structures themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `retain_mut` to `Deque`
 - Added `into_raw` to `Box`
 - Added `from_raw` to `Box`
+- Added `FusedIterator` to `vec::IntoIter`, `deque::IntoIter`, `index_map::IntoIter` and `linear_map::IntoIter`.
+- Added `ExactSizeIterator` to `vec::IntoIter`, `deque::IntoIter`, `index_map::IntoIter` and `linear_map::IntoIter`.
+- Added `DoubleEndedIterator` to `vec::IntoIter` and `deque::IntoIter`.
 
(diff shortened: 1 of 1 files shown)


src/index_map.rs
```diff
@@ -1744,8 +1758,23 @@ mod tests {
         src.insert("k4", "v4").unwrap();
         let clone = src.clone();
         for (k, v) in clone.into_iter() {
-            assert_eq!(v, *src.get(k).unwrap());
+            assert_eq!(v, src.remove(k).unwrap());
         }
+        assert!(src.is_empty());
```
therealbnut, line 1763: Added this check as previously this test would pass if the iterator only returned `None`.

src/linear_map.rs
```diff
@@ -877,6 +893,20 @@ mod test {
         for (k, v) in clone.into_iter() {
             assert_eq!(v, src.remove(k).unwrap());
         }
+        assert!(src.is_empty());
```
therealbnut, line 896: Added this check as previously this test would pass if the iterator only returned `None`.

PR #631, opened 2025-11-07
Enable a few rustfmt options:

zeenix:
Let's enable a few rustfmt options:

* [`imports_granularity = "Crate"`][ig]
* [`comment_width = 100`][cw]
* [`wrap_comments = true`][wc]

These make a lot of sense since they keep the code consistent with respect to line width and keep
all imports grouped by the crate, keeps the imports more readable IMO.

Note that these are nightly-only options currently, however we already use the nightly fmt in our CI anyway and if you run non-nightly fmt against the code, you'll just get 3 warnings and that's it.

[ig]: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity
[cw]: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#comment_width
[wc]: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#wrap_comments

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,3 @@
+imports_granularity = "Crate"
+comment_width = 100
+wrap_comments = true
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -33,14 +33,18 @@
 //! }
 //! ```
 
-use crate::vec::{OwnedVecStorage, VecStorage, VecStorageInner, ViewVecStorage};
-use crate::CapacityError;
-use core::cmp::Ordering;
-use core::fmt;
-use core::iter::FusedIterator;
-use core::marker::PhantomData;
-use core::mem::{ManuallyDrop, MaybeUninit};
-use core::{ptr, slice};
+use crate::{
+    vec::{OwnedVecStorage, VecStorage, VecStorageInner, ViewVecStorage},
+    CapacityError,
+};
+use core::{
+    cmp::Ordering,
+    fmt,
+    iter::FusedIterator,
+    marker::PhantomData,
+    mem::{ManuallyDrop, MaybeUninit},
+    ptr, slice,
+};
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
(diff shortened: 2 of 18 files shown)


.rustfmt.toml
```diff
@@ -0,0 +1,3 @@
+imports_granularity = "Crate"
```
sgued, line 1: Not really a fan of this one. Sometimes I find it clearer to have multiple statements depending on what the imports are used for.

But I won't block on this, I rarely spend time looking at imports and just let rust-analyzer add them for me.

src/deque.rs
```diff
@@ -173,14 +177,16 @@ impl<T, const N: usize> Deque<T, N> {
...
+    /// [`storage_capacity`](DequeInner::storage_capacity) instead.
     pub const fn capacity(&self) -> usize {
         N
     }
 
     /// Returns the number of elements currently in the deque.
     ///
-    /// This method is not available on a `DequeView`, use [`storage_len`](DequeInner::storage_len) instead.
+    /// This method is not available on a `DequeView`, use [`storage_len`](DequeInner::storage_len)
+    /// instead.
```
sgued, line 189: This is why I don't like wrap-comments that much.

sgued (approved):

src/deque.rs
zeenix: This would be the same if this was a physical book. There has to be a limitation and it needs to be consistent. Sure, sometimes it ends up with more lines that you'd have wanted but I think that's really not something worth sacrificing consistency for, IMO.

.rustfmt.toml
zeenix: > Not really a fan of this one. Sometimes I find it clearer to have multiple statements depending on what the imports are used for.

I have been using this for years without questioning it so it's very much possible you have a point here. Any chance you could give an example? For me the important thing here is that it avoids a lot of repetition but I can be convinced to go for another kind of granularity. We just need to keep it consistent. 

> But I won't block on this, I rarely spend time looking at imports and just let rust-analyzer add them for me.

Same here when writing code but when when reading code, I do need to see them.

PR #633, opened 2025-11-23
implement a pendant to `Cstr::from_bytes_until_nul`

johann-cm:
I would have liked to build the implementation upon safe functions, but all my attempts needed to check the no-interior-nul condition twice: once for trimming the input bytes slice, and once for safely extending the String with it.
Of course, by trimming the input bytes to the first nul byte, we already know that there are no interior nul bytes.
With unsafe code, we can make use of that knowledge and don't need to check again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-- Added `from_bytes_until_nul` to `CString`
+- Added `from_bytes_truncating_at_nul` to `CString`
 
 ## [v0.9.2] 2025-11-12
 


sgued:
Thanks for the contribution.

Please add some tests (exercising all branches, not null terminated, null terminated, interior null, and with or without the capacity).

The behaviour here is net the same as `CStr::from_bytes_until_null`, which fails if there is no null (while here we add the null in that case). This could be confusing and I'm in favour of renaming it, or rejecting non-null terminated values. `from_bytes_truncating_at_null` is a bit verbose but clearer.
What do you think?

johann-cm:
I've added one test case. As far as I can tell, all other cases are covered now by the doc-tests, or am I missing something?

I've renamed the method, in order to avoid confusion. I think adopting the `CStr` behavior of throwing an error on non-nul data is unnecessary here, because we copy the data and can just push a nul byte anyway. This eliminates one source of error for free.
In the doc-comment, I've made the differences to `CStr::from_bytes_until_nul` more explicit.

sgued:
The current doc tests don't test the case where the null is not at the end.

johann-cm:
Okay, I see what you mean. I've tweaked one test to cover that case.

zeenix:
@sgued Thanks for taking care of reviewing and merging this PR. In the future, please squash merge if all commits after the first one are just fixups of the first one so we don't have unnecessary commits in git.

PR #635, opened 2025-12-14
Add `Vec::remove_insert()`

MoSal:
In patterns where an insertion directly follows a removal, this serves
 as being more efficient, since it only shifts/copies values as needed
 for the combined operation.

 This also comes with the added bonus of not needing to check for
 fullness, and thus, a `Result` is not needed as a return type.

### DIFF
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -960,6 +960,89 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
         value
     }
 
+    /// This is the same as calling [`Vec::remove`] with `remove_index`
+    /// followed by calling [`Vec::insert`] with `insert_index` and `element`.
+    ///
+    /// The returned value is the removed element.
+    ///
+    /// This is more efficient than removing then inserting since it only shifts
+    /// `remove_index.abs_diff(insert_index)` values.
+    ///
+    /// [`remove`]: Vec::remove
+    /// [`insert`]: Vec::insert
+    ///
+    /// # Panics
+    ///
+    /// Panics if `remove_index` or `insert_index` are out of bounds.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::Vec;
+    ///
+    /// let mut v: Vec<_, 8> = Vec::from_slice(&[0, 1, 2, 3]).unwrap();
+    /// assert_eq!(v.remove_insert(1, 2, 4), 1);
+    /// // only one element (2) is shifted back
+    /// assert_eq!(v, [0, 2, 4, 3]);
+    /// ```
+    pub fn remove_insert(&mut self, remove_index: usize, insert_index: usize, element: T) -> T {
+        let length = self.len();
+
+        assert!(remove_index < length);
+        assert!(insert_index < length);
+
+        // SAFETY: `remove_index < length` assertion guarantees the pointer is within bounds.
+        let to_remove = unsafe { ptr::read(self.as_ptr().add(remove_index)) };
+
+        match remove_index.cmp(&insert_index) {
+            Ordering::Equal => (),
+            Ordering::Less => {
+                // SAFETY: `remove_index < length` assertion guarantees the pointer is within
+                // bounds.
+                let remove_at = unsafe { self.as_mut_ptr().add(remove_index) };
+
+                // SAFETY:
+                // Copies from
+                //  (remove_index + 1)..(insert_index + 1)
+                // to
+                //  remove_index..insert_index
+                //
+                // remove_index < insert_index (this match)
+                // remove_index + 1 < insert_index + 1 (can't saturate because <= length)
+                // remove_index + 1 <= insert_index < length
+                // insert_index + 1 <= length
+                //
+                // If `remove_index == 0`, and `insert_index + 1 ==  length`, then we copy from
+                // 1..length to 0..(length-1).
+                unsafe { ptr::copy(remove_at.add(1), remove_at, insert_index - remove_index) };
+            }
+            Ordering::Greater => {
+                // SAFETY: `insert_index < length` assertion guarantees the pointer is within
+                // bounds.
+                let insert_at = unsafe { self.as_mut_ptr().add(insert_index) };
+
+                // SAFETY:
+                // Copies from
+                //  insert_index..remove_index
+                // to
+                //  (insert_index + 1)..(remove_index + 1)
+                //
+                // insert_index < remove_index (this match)
+                // insert_index + 1 < remove_index + 1 (can't saturate because <= length)
+                // insert_index + 1 <= remove_index < length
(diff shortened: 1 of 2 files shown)


src/vec/mod.rs
```diff
@@ -960,6 +960,56 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
...
+    /// // only one element (2) is shifted back
+    /// assert_eq!(v, [0, 2, 4, 3]);
+    /// ```
+    pub fn remove_insert(&mut self, remove_index: usize, insert_index: usize, element: T) -> T {
+        let length = self.len();
+
+        assert!(remove_index < length);
+        assert!(insert_index < length);
+
+        unsafe {
```
zeenix: * Can we have multiple unsafe blocks with each having as little scope as possible?
* Please adds `Safety: ` comments above each use of `unsafe`, briefly describing how the invariants are being held in the block.

MoSal: Pushed a commit that does this (unsquashed, in case I overdid the explanations).

sgued:
Do you know why this is not also available in the standard library's Vec? This would actually not even need to be implemented on `Vec` and could be implemented on `&mut [T]`, making it more generic.

src/vec/mod.rs
```diff
@@ -960,6 +960,89 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
...
+    ///
+    /// ```
+    /// use heapless::Vec;
+    ///
+    /// let mut v: Vec<_, 8> = Vec::from_slice(&[0, 1, 2, 3]).unwrap();
+    /// assert_eq!(v.remove_insert(1, 2, 4), 1);
+    /// // only one element (2) is shifted back
+    /// assert_eq!(v, [0, 2, 4, 3]);
+    /// ```
+    pub fn remove_insert(&mut self, remove_index: usize, insert_index: usize, element: T) -> T {
```
Darksonn, line 988: How about?
```Rust
fn remove_insert(&mut self, remove_index: usize, insert_index: usize, element: T) -> T {
    if remove_index < insert_index {
        self[remove_index..=insert_index].rotate_left(1);
    } else if insert_index < remove_index {
        self[insert_index..=remove_index].rotate_right(1);
    }
    mem::replace(&mut self[insert_index], element)
}
```

sgued: A benchmark would be welcome to compare both implementations. Given the complexity of the implementation of `rotate_left` and `rotate_right`, I'm convinced it would be slower and larger in code size.

MoSal:
@sgued
I asked myself that actually.

Anyway, I relayed this to [zulip](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/.60remove_insert.28.29.20for.20.60Vec.60.2C.20or.20more.20generally.20for.20.60.26mut.20.5BT.5D.60/with/564238856). And `slice` methods `rotate_left()`/`rotate_right()` make this much simpler (commit incoming).

src/vec/mod.rs
MoSal: I wrote the comment below before seeing this conversation.

MoSal: > A benchmark would be welcome to compare both implementations. Given the complexity of the implementation of rotate_left and rotate_right, I'm convinced it would be slower and larger in code size.

Yep. It's even slower that `romove()`+`insert()`.

https://github.com/MoSal/bench-insert-remove

```
remove_insert_rotate u16    time:   [42.533 ms 42.662 ms 42.813 ms]
remove and insert u16       time:   [31.312 ms 31.379 ms 31.456 ms]
remove_insert_unsafe u16    time:   [16.316 ms 16.383 ms 16.474 ms]

remove_insert_rotate u8     time:   [55.221 ms 55.435 ms 55.695 ms]
remove and insert u8        time:   [46.525 ms 46.745 ms 47.026 ms]
remove_insert_unsafe u8     time:   [36.480 ms 36.582 ms 36.701 ms]
```

The test applies the Move-To-Front transform then the reverse on random u16 and u8 buffers. The u8 benches only differ in input buffer being larger, and `memchr::memchr()` being used instead of `.iter().position()`.

MoSal:
I added this as a local trait implemented for `AsMut<[T]>` types in the project where I needed this and moved on.

Feel free to do with it whatever you want. The implementation using rotates is slower as noted a few days ago. So, only the original impl is actually useful performance wise.

zeenix:
> Feel free to do with it whatever you want. The implementation using rotates is slower as noted a few days ago. So, only the original impl is actually useful performance wise.

I understand that you yourself no longer need this but please consider the time reviewers put into this as well. It seems to me like all that's left now is reverting to the original implementation (`git reflog` to find the commit, `git reset --hard THE_COMMIT_HASH` to revert to that and then `git push --force`)?

MoSal:
@zeenix 
Squashed.

zeenix:
I'm a bit baffled by the CI failure. I can't reproduce the warning locally. :thinking:

sgued:
This is a bug in 1.92 I think, you may not have the latest compiler.

sgued:
Given that this is a feature that can be implemented on slices, I'm not fully certain it makes sense to merge it here? I'd rather try to get it implemented upstream in `core`. As https://github.com/rust-embedded/heapless/pull/635#issuecomment-3678209939 points it can be implemented downstream too, so I'm not sure it should be merged as-is.

sgued:
Hmm, no this does not reproduce for me either

sgued:
This might be related: https://github.com/rust-lang/rust/issues/147648

scottmcm:
(Hi, I'm from the peanut gallery, having seen this in [#t-libs > &#96;remove_insert() for &#96;Vec&#96;, or more generally for &#96;&amp;mut &#91;T&#93;&#96; @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/.60remove_insert.28.29.20for.20.60Vec.60.2C.20or.20more.20generally.20for.20.60.26mut.20.5BT.5D.60/near/565715426))

> In patterns where an insertion directly follows a removal

Note that if it doesn't change the *length* then it'd be better as something on slice instead of on `Vec`.

That would also simplify the interface, because the slice could be sub-sliced `[remove_index..=insert_index]` and then the "move everything sideways one, filling in the hole with the argument and returning the thing that goes off the end" doesn't need to mention indexes at all.

(Some people might still define a local helper with the `Ord::cmp` to dispatch between them properly, but I bet often you know the relative positions)

jannic:
The suggested methods on slice have been implemented. (Currently as an unstable feature, see https://github.com/rust-lang/rust/issues/151772 for the tracking issue.) 

Can this merge request be closed?

sgued:
Yes, the API is not exactly the same but it can do the same thing with the same performance. I'll close this. Thank you for the update!

PR #639, opened 2026-01-04
Cstring into Vec or String

johann-cm:
I've implemented the methods `into_bytes`, `into_bytes_with_nul`, and `into_string`, which are present in the `std::ffi::CString`.

I've also added `thiserror` as a dependency to neatly derive the `Error` trait on the `IntoStringError` struct.
If you prefer to keep dependencies to an absolute minimum, I can also remove `thiserror` and implement the `Error` trait by hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 - Added `from_bytes_truncating_at_nul` to `CString`
-- Added `into_bytes` to `CString`
-- Added `into_bytes_with_nul` to `CString`
-- Added `into_string` to `CString`
+- Added `CString::{into_bytes, into_bytes_with_nul, into_string}`
 
 ## [v0.9.2] 2025-11-12
 


Cargo.toml
```diff
@@ -71,6 +71,7 @@ ufmt-write = { version = "0.1", optional = true }
 defmt = { version = "1.0.1", optional = true }
 zeroize = { version = "1.8", optional = true, default-features = false, features = ["derive"] }
 embedded-io = { version = "0.7", optional = true }
+thiserror = {version = "2.0.12", default-features = false}
```
zeenix: I'm not sure it's worth adding a dep, just for one additional error type. It might be better to split this into a separate (follow-up) PR, where the advantage of adding this dep will be more prominent and you could make a better case for it separately. Just my 2 cents.

zeenix (changes requested):
Nice!

Cargo.toml
johann-cm: Okay, I've removed the dep from this PR and will be opening another one :+1:

sgued: I don't think I would approve adding `thiserror`. Currently we don't depend on `syn` even transitively by default (we depend on it only through the `defmt`, `ufmt`, and `zeroize` features). I don't think the "nice to have" practicality of `thiserror`, which is easily replaced by just a couple of lines of code justifies slowing down downstream builds that don't depend on `syn`.

src/c_string.rs
```diff
@@ -415,6 +491,64 @@ impl<const N: usize, LenT: LenType> fmt::Debug for CString<N, LenT> {
...
+    #[inline]
+    /// Access the underlying UTF-8 error that was the cause of this error.
+    pub fn utf8_error(&self) -> Utf8Error {
+        self.error
+    }
+}
+
+impl<const N: usize, LenT: LenType> Display for IntoStringError<N, LenT> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "CString contained non-utf8 bytes")
```
sgued:
```suggestion
        "CString contained non-utf8 bytes".fmt(f)
```

Why not follow `std`'s implementation here? One less macro and one less layer of indirection.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 - Added `from_bytes_truncating_at_nul` to `CString`
+- Added `into_bytes` to `CString`
+- Added `into_bytes_with_nul` to `CString`
+- Added `into_string` to `CString`
```
zeenix: Oh, just realized that this could just be one bullet:


```suggestion
- Added `CString::{into_bytes, into_bytes_with_nul, into_string}`
```

zeenix (changes requested):
I promise, last nicpick. :)

zeenix (approved):
LGTM.

PR #640, opened 2026-01-10
add FromUtf8Error in sync with std

johann-cm:
I'm proposing changing the error type of `String::from_utf8` form `core::str::Utf8Error` to `heapless::string::FromUtf8Error`. This would mirror the error type of `std::String::from_utf8` (which is `std::string::FromUtf8Error`). This is a breaking change.

`FromUtf8Error` has the advantage that you can regain ownership over the data that you want to convert to UTF-8 in case of an error, which you currently lose in that case.
The change would have made it possible for me in #639 to implement `into_string` without unsafe code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 - Added `from_bytes_truncating_at_nul` to `CString`
+- Added `FromUtf8Error` to `heapless::string`
+- [breaking-change] Changed error type of `String::from_utf8` from `core::str::Utf8Error` to `heapless::string::FromUtf8Error`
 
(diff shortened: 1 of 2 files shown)


sgued:
Thank you for the PR!

I'm not convinced it is that useful

The ergonomics of having the Error type have a const-generic and not being `Copy` don't seem justified. Unlike in `std`, we never allocate, so cloning the `Vec` is somewhat cheap and can be done before calling the `from_utf8` function without a significant performance cost. I suspect LLVM is very capable of avoiding the copy when doing this. If a benchmark shows this is justified, I would merge this, but otherwise I think keeping it the way it currently is is best from a downstream ergonomics standpoint.

therealprof:
I agree with @sgued here and I'm also not a big fan of the breaking change, even if it shouldn't really be too bad.

Issue #442, opened 2024-01-15
API alignment


jordens:
In the light of aiming for a stabilized API of v1.0 (#410) but having just done a significant API changeset in 0.8, I wanted to bring something up that has been bugging me several times.

Currently the [`heapless::Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html) API is significantly different from both [`std::Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html) and alternative no_std/no-alloc `Vec` implementations like [`tinyvec::ArrayVec`](https://docs.rs/tinyvec/latest/tinyvec/struct.ArrayVec.html) or [`arrayvec::ArrayVec`](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html). This is especially true for the important case where `std::Vec` would reallocate but its fixed size alternatives must either panic or become fallible: e.g. `Vec::push()`.

`heapless` deviates from the others in its API and makes these fallible. The others are panicing and **additionally**/**alternatively** tend to offer a fallible API ([`ArrayVec::try_push()`](https://docs.rs/tinyvec/latest/tinyvec/struct.ArrayVec.html#method.try_push)).

IMO it would be nice if we could overall

* align the API with the `std::Vec` API and `tinyvec`
* offer a fallible API (`try_push()` etc).

That would make porting a bit easier. And it could allow `heapless::Vec` to be a drop-in replacement for `tinyvec::ArrayVec`. The only (AFAICT) remaining difference would be under the hood: the use of `MaybeUninit`.

The API difference comes up from time tom time e.g. #440.
There is demand for distinct operations that are both infallible and non-panicing: #341

There may also be other places in `heapless` where the API choice diverges from the `std` analogue.

newAM:
I agree with this.  It's going to be an annoying breaking change, but `std::vec::Vec::try_push` will exist at some point for the Linux kernel and we should align with that.

There are definitely other places where the API diverges, `String::from_utf8` for example, and not having `String::from` because `from` traits cannot panic.

In general we're going to have the same problems the Linux kernel has with `std::` for fallible vs infallible allocations.  This crate is going to be used for embedded systems and we should make the difference between panicky and non-panicky functions clear somehow.

Dirbaio:
I'm not sure if I agree with this.

[...]

PR #642, opened 2026-01-22
Don't panic in VecInner::extend and return a Result

sirhcel:
Other methods for extending VecInner already return results instead of panicking. In case of VecInner::extend, avoiding a panic by checking that the actual amount of elements to extend with fits beforehand is not generally possible.

So attempt to add the elements from the iterator and restore the original length in case of an error.

This is a semver-breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `from_bytes_truncating_at_nul` to `CString`
 - Added `CString::{into_bytes, into_bytes_with_nul, into_string}`
 - Added `pop_front_if` and `pop_back_if` to `Deque`
+- Replace panicking in `VecInner::extend` with returning a `Result`.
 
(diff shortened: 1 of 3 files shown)


sgued:
This goes hand in hand with the panicking `FromIterator` implementations. 

There was some discussion of the topic in https://github.com/rust-embedded/heapless/issues/442.

My opinion is that for the next breaking release we should aim to remove implicitly panicking APIs.

sirhcel:
> This goes hand in hand with the panicking `FromIterator` implementations.
> 
> There was some discussion of the topic in #442.

Thank you for the references! The latter is a great read on the pros and cons. Do you mean the former in the sense that there should be a fallible conversion like `TryFromIterator` instead?


> My opinion is that for the next breaking release we should aim to remove implicitly panicking APIs.

I would prefer this too. Especially for operations involving iterators where getting to know the actual length/number of items to return look non-trivial to me and opting in to panicking with unwrapping the result seems way less annoying to me than writing all the pre-checks for ensuring that an operation won't panic.

PR #652, opened 2026-04-02
Remove the modulo operations in spsc

sgued:
These modulo operations used to be well optimized when N was a power of 2 However, the consumer and producer now use view-types that make N runtime dependant, preventing the compiler from optimizing these modulo operations even when N is always a power of 2.

This patch leverages the fact that `head` and `tail` are always kept lower than N to replace the modulo operations with a simple if, which gets optimized pretty well by the compiler and no branch is left.

Closes https://github.com/rust-embedded/heapless/issues/650

### DIFF
--- a/src/spsc.rs
+++ b/src/spsc.rs
@@ -176,7 +176,10 @@ impl<T, S: Storage> QueueInner<T, S> {
 
     #[inline]
     fn increment(&self, val: usize) -> usize {
-        let val = val + 1;
+        // We know that self.n() <= usize::MAX
+        // So this can only overflow if N == usize::MAX
+        // and in this case the overflow will be equivalent to the modulo N operation
+        let val = val.wrapping_add(1);
         if val >= self.n() {
             val - self.n()
         } else {
@@ -628,8 +631,11 @@ impl<'a, T> Iterator for Iter<'a, T> {
         if self.index < self.len {
             let head = self.rb.head.load(Ordering::Relaxed);
 
-            let i = head + self.index;
-            let i = if i >= self.rb.n() { i - self.rb.n() } else { i };
+            let i = match head.checked_add(self.index) {
+                Some(i) if i >= self.rb.n() => i - self.rb.n(),
+                Some(i) => i,
+                None => head.wrapping_add(self.index).wrapping_sub(self.rb.n()),
+            };
             self.index += 1;
 
             Some(unsafe { &*(self.rb.buffer.borrow().get_unchecked(i).get() as *const T) })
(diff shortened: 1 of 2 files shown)


sgued:
I wish we had some benchmarking infra in place. This would help me make sure this actually solves the problem and would have helped detect it in the first place.

sgued:
Clippy lint fixes are already included in https://github.com/rust-embedded/heapless/pull/644

sgued:
@823984418 does this fix your performance issues?

src/spsc.rs
```diff
@@ -626,7 +628,8 @@ impl<'a, T> Iterator for Iter<'a, T> {
         if self.index < self.len {
             let head = self.rb.head.load(Ordering::Relaxed);
 
-            let i = (head + self.index) % self.rb.n();
+            let i = head + self.index;
```
jannic: Could this (theoretically) overflow? If n is larger than half of the usize::MAX and the queue already wrapped around so head is close to the end of the underlying array, and index is large?

(Very unlikely in practice, and I only had this idea because I wondered why QueueInner::len uses wrapping_add/wrapping_sub).

sgued: Good catch.

This would be a bug even with the original implementation if `n` is not a divisor of `usize::MAX`.

I don't think this is the same as for `len`. in `len` we're doing wrapping operations because we "know" it's going to go negative (thus wrap) since `current_head > current_tail`. We could in theory remove the wrapping operation by changing the order (but then it's sensible to the same bug, where it could in theory overflow if N is too close to `usize::MAX`).

I'll take a look at fixing this.

jannic: > We could in theory remove the wrapping operation by changing the order (but then it's sensible to the same bug, where it could in theory overflow if N is too close to usize::MAX)

Yes, sorry, my comment about len was a bit short. Of course, as it's written currently, the wrapping_sub/add calls are required. I was thinking about changing that by changing the order. But then:
a) as you noticed, it would have the same potential overflow issue
b) wrapping operations might be more efficient, because they don't require any overflow checks (if they are enabled, e.g. in debug mode or by setting `overflow-checks = true` in `[profile.release]`)

So I decided to not suggest removing the wrapping operations.

sgued: I added a fix for this in 73261bd14ecd04b33bdbbd007b618ba87f5de664

sgued: Crap, looks like miri doesn't like the `[(); usize::MAX]` zero-size arrays...

jannic (approved):
LGTM. Nice idea to use zero-sized types for the overflow checks!

zeenix (approved):

PR #654, opened 2026-04-14
Backport SPSC fixes to 0.7 branch

sgued:
This is being backported because it's a fix to an easy to reach UB if you're ever using N > usize::MAX/2, which could happen on some platforms where usize is 16 bits.

To minimise the changes between the two releases once this is merged I'll release 0.7.18 from commit [420a49d](https://github.com/rust-embedded/heapless/pull/654/commits/420a49dfe35cbfb6f94e293c05a18cbd82458d88)

https://github.com/rust-embedded/heapless/pull/656 should be reviewed and merged first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- spsc: Fix integer overflow in iterators when N > usize::MAX/2 and the queue loops.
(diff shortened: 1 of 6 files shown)


reitermarkus (approved):

jannic:
The pipeline could be fixed by backporting https://github.com/rust-embedded/heapless/pull/622.

sgued:
There have been many more fixes that would need to be backported if we consider backporting to the 0.8 and 0.7 branches.
I don't really have the motivation to do so especially given they would require more work to properly backport and test so I will close this and I think we can consider these branches unmaintained until someone decides to take the work of doing the backports.

PR #655, opened 2026-04-14
Backport SPSC fixes to 0.8 branch

sgued:
This is being backported because it's a fix to an easy to reach UB if you're ever using N > usize::MAX/2, which could happen on some platforms where usize is 16 bits.

To minimise the differences with 0.8.0 once this is merged I'll release 0.8.1 from commit [0ed6913](https://github.com/rust-embedded/heapless/pull/655/commits/0ed6913214f08c802155885e690c172500de817f)

https://github.com/rust-embedded/heapless/pull/656 should be reviewed and merged first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- spsc: Fix integer overflow in iterators when N > usize::MAX/2 and the queue loops.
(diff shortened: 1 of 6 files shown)


reitermarkus (approved):

jannic:
I assume this MR should not be merged into main, but into a (not yet existing) `v0.8.x` branch.

sgued:
Thanks @jannic, fixed the target branch

sgued:
There have been many more fixes that would need to be backported if we consider backporting to the 0.8 and 0.7 branches.
I don't really have the motivation to do so especially given they would require more work to properly backport and test so I will close this and I think we can consider these branches unmaintained until someone decides to take the work of doing the backports.

PR #657, opened 2026-04-21
Remove release date.

reitermarkus:
Maybe we should just remove the date, since this always requires the release PR to be up-to-date when merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [v0.9.3] 2025-04-15
+## [v0.9.3]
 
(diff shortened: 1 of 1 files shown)


sgued (approved):

sgued (changes requested):
Wait the release was not published to crates.io

reitermarkus:
Ah, sorry, I didn't realize we don't have a publishing workflow.

reitermarkus:
Okay, it is now published.

Also removed the date here now since it doesn't really matter anyways when you can trivially find the commit.

sgued:
If we remove the release date for 0.9.3 we should remove it for every release no ?

zeenix:
>  we remove the release date for 0.9.3 we should remove it for every release no ?

Fwiw, I agree. I also don't think there's anything wrong with removing a tag and then retagging. A tag represents a release and if it didn't really happen, the tag in question loses all its meaning.

PR #658, opened 2026-04-23
Add SortedLinearMap

extremeandy:
A fixed-capacity map that keeps entries sorted by key and performs lookups via binary search - the `BTreeMap` analogue of `LinearMap`.

There might be a better name for this - I don't really like `SortedLinearMap` but couldn't decide what else to call it.

This is slower for inserts (O(n)) but faster for lookups (O(log(n)) than LinearMap. Works well for data with low frequency writes and high frequency reads.

Includes BTreeMap-style APIs: first_key_value/last_key_value, pop_first/pop_last, range/range_mut, plus the usual entry API, retain, Serialize/Deserialize, and Zeroize support.

This was the prompt I used to kick off the work:

> i want to add a SortedLinearMap which is the BTreeMap equivalent that HashMap is to LinearMap.
> inserts and reads should use a binary search to find the entry. the API should otherwise be quite similar to LinearMap.

I reviewed changes, made some minor tweaks, and had Claude review it as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


ambihelical:
I know C++ isn't the best source of good naming (see std::vector, but that ship has sailed), but it uses flat_map for this kind of container.

PR #660, opened 2026-04-27
Fix unwind vulnerability

### LINKED ISSUE #659: [SECURITY] Panic-safety vulnerabilities in heapless v0.9.2 with demonstrated arbitrary code execution
# [SECURITY] Panic-safety vulnerabilities in heapless v0.9.2 with demonstrated arbitrary code execution

## Summary

I have identified four panic-safety vulnerabilities in heapless v0.9.2 affecting core container operations. Through security research, I have developed proof-of-concept demonstrations showing that two of these vulnerabilities can be exploited for arbitrary code execution.

**Technical Assessment:**
- **Affected APIs**: 4 container operations
- **Severity**: High to Critical (CVSS 7.8-8.5 range)
- **Attack Vector**: Local execution via safe Rust code
- **Demonstrated Impact**: Arbitrary code execution capability

## Affected APIs

1. `HistoryBuf::write()` - Single metadata update vulnerability
2. `Vec::IntoIter::drop()` - Single metadata update vulnerability
3. `HistoryBufInner::clear()` - Multiple metadata update vulnerability **(ACE demonstrated)**
4. `DequeInner::clear()` - Multiple metadata update vulnerability **(ACE demonstrated)**

## Disclosure Background

I am disclosing this issue publicly as I received guidance to proceed with a public issue.

### Vulnerability Pattern

All four issues share a common root cause where container metadata updates occur after element destruction rather than before:

```rust
// Vulnerable pattern
unsafe { ptr::drop_in_place(...) }; // Panic here leaves container inconsistent
self.metadata = safe_value;         // Never reached
```

When an element's `Drop` implementation panics, containers remain in inconsistent states, leading to:
- Use-after-free conditions during subsequent operations
- Double-free scenarios with heap-allocated element types
- Memory corruption exploitable through controlled heap manipulation

### Arbitrary Code Execution Capability

I have successfully developed working demonstrations for the `clear()` method vulnerabilities that achieve arbitrary code execution through vtable hijacking.

**Exploitation technique:**
- Storing trait objects with controlled vtable pointers
- Triggering panic conditions during element destruction
- Using heap spraying to reclaim freed memory locations
- Hijacking vtable calls to redirect execution flow

**Working proof-of-concept code is available for private review if a security advisory process is established.**

Security Advisory Request
I would respectfully request consideration of creating a GitHub Security Advisory (GHSA) for these findings.
The most significant aspect of these discoveries is that actual arbitrary code execution capability has been demonstrated. Beyond simple memory safety issues, I have confirmed the reliability of vtable hijacking techniques through proof-of-concept that achieves shell execution. Such practical attack scenarios could pose serious security risks, particularly in embedded and systems programming environments.
The Rust ecosystem has established precedent for handling panic-safety vulnerabilities with demonstrated exploitation potential through coordinated security advisory processes. Most notably, CVE-2026-6654 (thin-vec) addressed the identical panic-safety → ACE pattern, confirming that this class of vulnerability warrants formal security advisory treatment when arbitrary code execution is demonstrated.
Given the demonstrated arbitrary code execution capability and the established precedent for handling similar panic-safety issues, I believe these findings would benefit from the same coordinated security process to ensure proper vulnerability tracking and ecosystem notification.
Thank you for maintaining heapless and for considering appropriate security processes for these findings.

sgued:
Fixes #659

Prior to this, a panic in a downstream drop implementation could lead to an inconsistent state during unwinding that could lead to undefined behaviour.

0.9.3 has not been released to crates.io for some reason, so I suggest we add https://github.com/rust-embedded/heapless/pull/652 and this PR to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
+- Fixed unsoundness in `Vec::IntoIter::drop` and `HistoryBuf::write`in the context of panicking drop implementations.
 
 ## [v0.9.3] 2025-04-15
 
-- Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear` and `IndexMap::clear` in the context
-of panicking drop implementations.
+- Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear` and `IndexMap::clear` in the context of panicking drop implementations.
 - Added `from_bytes_truncating_at_nul` to `CString`
 - Added `CString::{into_bytes, into_bytes_with_nul, into_string}`
 - Added `pop_front_if` and `pop_back_if` to `Deque`
--- a/src/history_buf.rs
+++ b/src/history_buf.rs
@@ -385,9 +385,12 @@ impl<T, S: HistoryBufStorage<T> + ?Sized> HistoryBufInner<T, S> {
 
     /// Writes an element to the buffer, overwriting the oldest value.
     pub fn write(&mut self, t: T) {
+        let _tmp;
         if self.filled {
-            // Drop the old before we overwrite it.
-            unsafe { ptr::drop_in_place(self.data.borrow_mut()[self.write_at].as_mut_ptr()) }
+            // Copy the old so that it is dropped at the end
+            // We don't drop it now so that a panic in its destructor doesn't
+            // lead to an invalid state
+            _tmp = unsafe { ptr::read(self.data.borrow_mut()[self.write_at].as_mut_ptr()) };
         }
         self.data.borrow_mut()[self.write_at] = MaybeUninit::new(t);
 
@@ -994,10 +997,11 @@ mod tests {
         }
     }
 
+    // Tests that a panic in in a downstream drop implementation does not lead to an
+    // inconsistent state during unwinding that could lead to undefined behaviour.
+    // See https://github.com/rust-embedded/heapless/issues/646
     #[test]
     fn test_use_after_free_clear() {
-        // See https://github.com/rust-embedded/heapless/issues/646
-
         static COUNT: AtomicI32 = AtomicI32::new(0);
 
         #[derive(Debug)]
@@ -1029,6 +1033,45 @@ mod tests {
         catch_unwind(move || unwind_safe_history_buf.clear()).unwrap_err();
         assert_eq!(history_buf.len(), 0);
         history_buf.clear();
+        assert_eq!(Dropper::count(), 0);
+    }
+
+    // Tests that a panic in in a downstream drop implementation does not lead to an
+    // inconsistent state during unwinding that could lead to undefined behaviour.
+    // See https://github.com/rust-embedded/heapless/issues/659
+    #[test]
+    fn test_use_after_free_write() {
+        static COUNT: AtomicI32 = AtomicI32::new(0);
+
+        #[derive(Debug)]
+        struct Dropper(bool);
+
+        impl Dropper {
+            fn new(should_panic: bool) -> Self {
+                COUNT.fetch_add(1, Ordering::Relaxed);
+                Self(should_panic)
+            }
+            fn count() -> i32 {
+                COUNT.load(Ordering::Relaxed)
+            }
+        }
+        impl Drop for Dropper {
+            fn drop(&mut self) {
+                COUNT.fetch_sub(1, Ordering::Relaxed);
+                assert!(!self.0, "Testing  panicking");
+            }
+        }
+
+        let mut histbuf = HistoryBuf::<Dropper, 5>::new();
+        histbuf.write(Dropper::new(true));
+        histbuf.write(Dropper::new(false));
+        histbuf.write(Dropper::new(false));
+        histbuf.write(Dropper::new(false));
+        histbuf.write(Dropper::new(false));
+        let mut unwind_safe = AssertUnwindSafe(&mut histbuf);
+
+        catch_unwind(move || unwind_safe.write(Dropper::new(false))).unwrap_err();
+        drop(histbuf);
         assert_eq!(Dropper::count(), 0);
     }
 
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -745,7 +745,7 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
         unsafe {
             // Note: It's intentional that this is `>` and not `>=`.
             //       Changing it to `>=` has negative performance
-            //       implications in some cases. See rust-lang/rust#78884 for more.
+            //       implications in some cases. See https://github.com/rust-lang/rust/issues/78884 for more.
             if len > self.len() {
                 return;
             }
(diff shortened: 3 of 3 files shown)


src/vec/mod.rs
```diff
@@ -1571,11 +1571,18 @@ where
 
 impl<T, LenT: LenType, const N: usize> Drop for IntoIter<T, N, LenT> {
     fn drop(&mut self) {
+        // Ensures that everything is removed even if the drop implementation panics
+        struct Guard<'a, T, LenT: LenType, const N: usize>(&'a mut IntoIter<T, N, LenT>);
+        impl<'a, T, LenT: LenType, const N: usize> Drop for Guard<'a, T, LenT, N> {
+            fn drop(&mut self) {
+                // Prevent dropping of other elements
+                self.0.vec.len = LenT::ZERO;
```
Dirbaio: why not set length to zero *before* calling drop_in_place? Seems simpler than writing a guard.

sgued: Because it would break the use of `as_mut_slice`, but you're right I fixed it.

src/history_buf.rs
```diff
@@ -1032,6 +1035,44 @@ mod tests {
         assert_eq!(Dropper::count(), 0);
     }
 
+    #[test]
+    fn test_use_after_free_write() {
+        // See https://github.com/rust-embedded/heapless/issues/659
```
zeenix: It'd be good to add a bried explanation here for offline viewing/coding.

zeenix (changes requested):
Please also add an explanation to the PR description (and commit message too, if we don't do a squash merge).

src/history_buf.rs
sgued: Added

zeenix: @sgued Great. This wasn't addressed though:

Please also add an explanation to the PR description (and commit message too, if we don't do a squash merge).

sgued: Done

zeenix (dismissed):

reitermarkus:
> 0.9.3 has not been released to crates.io for some reason, so I suggest we add https://github.com/rust-embedded/heapless/pull/652 and this PR to it.

It is now, but anyways, the release tag already existed, and re-tagging is a very bad practice.

src/vec/mod.rs
```diff
@@ -2385,6 +2393,42 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_use_after_free_intoiter_drop() {
+        // This tests that a panic in a downstream drop implementation does not lead to an
+        // inconsistent state during unwinding that could lead to undefined behaviour.
+        // For more info see https://github.com/rust-embedded/heapless/issues/659
```
reitermarkus: Maybe move the comment to the test function instead.

src/vec/mod.rs
```diff
@@ -1571,11 +1571,15 @@ where
...
             // Drop all the elements that have not been moved out of vec
-            ptr::drop_in_place(&mut self.vec.as_mut_slice()[self.next.into_usize()..]);
-            // Prevent dropping of other elements
-            self.vec.len = LenT::ZERO;
+            let remaining = slice::from_raw_parts_mut(
+                self.vec.as_mut_ptr().add(self.next.into_usize()),
+                (len - self.next).into_usize(),
+            );
+            ptr::drop_in_place(remaining);
         }
```
reitermarkus, line 1641: Isn't this basically the same logic as `VecInner::truncate`? Does `VecInner::truncate` have the same vulnerability?

src/vec/mod.rs
```diff
@@ -1571,11 +1571,15 @@ where
 
 impl<T, LenT: LenType, const N: usize> Drop for IntoIter<T, N, LenT> {
     fn drop(&mut self) {
+        let len = self.vec.len;
+        self.vec.len = LenT::ZERO;
         unsafe {
```
reitermarkus, line 1635:
```suggestion
        // SAFETY: `self.vec` contains initialized elements in the range `self.next..len` 
        //         which have not been consumed yet, so it is safe to create
        //         a mutable slice to them and drop them.
        //
        //         `self.vec.len` is set to `0` before dropping the elements so that
        //         a panicking `Drop` implementation does not result in inconsistent 
        //         state during unwinding that could lead to undefined behaviour.
        unsafe {
```

src/vec/mod.rs
reitermarkus: Ah, `VecInner::truncate` sets `self.len` before `ptr::drop_in_place`, so no, but would be good to clean up the `unsafe` block there as well to be consistent with this one, and add the same test for it.

reitermarkus:
Added the same test for `VecInner::clear` and made the implementations consistent across `vec::IntoIter::drop` and `VecInner::truncate`.

reitermarkus (dismissed):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [v0.9.3] 2025-04-15
 
 - Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear` and `IndexMap::clear` in the context
+- Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear`, `IndexMap::clear`, `Vec::IntoIter::drop` and `HistoryBuf::write` in the context
```
reitermarkus:
```suggestion
## [Unreleased]

- Fixed unsoundness in `Vec::IntoIter::drop` and `HistoryBuf::write` in the context of panicking drop implementations.

## [v0.9.3] 2025-04-15

- Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear` and `IndexMap::clear` in the context
- Fixed unsoundness in `Deque::clear`, `HistoryBuf::clear` and `IndexMap::clear`  in the context
```

src/vec/mod.rs
sgued: It's not the same as truncate because we don't drop the beginning of the vec if it has already been iterated over.

CHANGELOG.md
sgued: Done

src/vec/mod.rs
sgued: Good idea, I added a safety comment.

src/vec/mod.rs
sgued: Done, I did not make it into a doc comment because then clippy complains about markdown formatting and this will never be shown in docs anyway

src/vec/mod.rs
```diff
@@ -1900,7 +1914,7 @@ mod tests {
     }
 
     #[test]
-    fn drop() {
+    fn test_drop() {
```
zeenix: Why? `test_` prefix in test function names is redundant & only adds noise to `cargo test` output.

src/vec/mod.rs
```diff
@@ -2517,6 +2531,41 @@ mod tests {
         }
     }
 
+    // Tests that a panic in a downstream drop implementation does not lead to an
+    // inconsistent state during unwinding that could lead to undefined behaviour.
+    // For more info see https://github.com/rust-embedded/heapless/issues/659
+    #[test]
+    fn test_use_after_free_intoiter_drop() {
```
zeenix: I'd instead remove the `test_` prefix from this new one.

zeenix (changes requested):
LGTM otherwise.

src/vec/mod.rs
sgued: I renamed it to `drop_impl`, since just `drop` leads to conflicts with `core::mem::drop` in the test bellow.

src/vec/mod.rs
sgued: Done

zeenix (approved):

PR #661, opened 2026-05-18
Add `Vec::resize_with`

jschillem:
Closes https://github.com/rust-embedded/heapless/issues/648.

Adds `Vec::resize_with`, the closure-based counterpart to the existing `resize` method, in line with the `std` library `resize_with` method.

This method is valuable when you wish to resize the `Vec` and require some computation for each new element.

### Changes
- `resize_with` method added to `Vec` with documentation specifying when to use each of the `resize` functions.
- Test coverage matching the `resize` method's coverage, as well as coverage for per-slot closure invocation, to test the invariant mentioned in the documentation.

### DIFF
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -814,7 +814,7 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
     ///
     /// # Examples
     ///
-    ///```
+    /// ```
     /// use heapless::Vec;
     ///
     /// let mut vec: Vec<usize, 8> = Vec::new();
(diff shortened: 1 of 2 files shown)


src/vec/mod.rs
```diff
@@ -796,6 +796,58 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
         self.resize(new_len, T::default())
     }
 
+    ///  Resizes the `Vec` in-place to that `len` is equal to `new_len`.
```
zeenix: typo: to -> so.

zeenix (changes requested):
LGTM otherwise.

zeenix (approved):

PR #663, opened 2026-06-06
Add retain_back (aka truncate_front) method to Deque

nullstalgia:
Howdy again!

I originally intended to do this after championing for the stabilization of `alloc`'s [`truncate_front`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.truncate_front) method, but it appears someone got to it before me!

Feeling inspired, I jumped right to the `heapless` implementation like I had done earlier last year for `Deque::truncate`, using the new name decided for the method in the [stabilization PR](https://github.com/rust-lang/rust/pull/151379#issuecomment-4432179760).

Miri seems happy, and running through the logic on pen and paper it seems sound, but please do let me know if I missed anything or if some more exhaustive tests would be needed!

### DIFF
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -2562,17 +2562,11 @@ mod tests {
 
             assert_eq!(Droppable::count(), LEN as i32);
 
-            let (front, back) = tester.as_slices();
-
             tester.retain_back(TRUNC);
             assert_eq!(tester.len(), TRUNC);
             assert_eq!(Droppable::count(), TRUNC as i32);
 
-            let (front, back) = tester.as_slices();
-
             tester.retain_back(0);
-
-            let (front, back) = tester.as_slices();
 
             assert_eq!(tester.len(), 0);
             assert_eq!(Droppable::count(), 0);
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -963,6 +963,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
         }
     }
 
+    /// Shortens the deque, keeping the last `len` elements and dropping
+    /// the rest.
+    ///
+    /// If `len` is greater or equal to the deque's current length, this has
+    /// no effect.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use heapless::Deque;
+    ///
+    /// let mut buf: Deque<_, 5> = Deque::new();
+    /// buf.push_back(5);
+    /// buf.push_back(10);
+    /// buf.push_back(15);
+    /// buf.retain_back(1);
+    /// assert_eq!(buf.make_contiguous(), [15]);
+    /// ```
+    #[doc(alias = "truncate_front")]
+    pub fn retain_back(&mut self, len: usize) {
+        /// Runs the destructor for all items in the slice when it gets dropped (gracefully or
+        /// during unwinding).
+        struct Dropper<'a, T>(&'a mut [T]);
+
+        impl<'a, T> Drop for Dropper<'a, T> {
+            fn drop(&mut self) {
+                unsafe {
+                    ptr::drop_in_place(self.0);
+                }
+            }
+        }
+
+        // Safety:
+        // * Any slice passed to `drop_in_place` is valid; the second case has `len <= back.len()`
+        //   and returning on `len > self.storage_len()` ensures `end <= front.len()` in the first
+        //   case
+        // * Deque front/back cursors are moved before calling `drop_in_place`, so no value is
+        //   dropped twice if `drop_in_place` panics
+        unsafe {
+            // If new desired length is greater or equal, we don't need to act.
+            if len >= self.storage_len() {
+                return;
+            }
+
+            let (front, back) = self.as_mut_slices();
+
+            if len > back.len() {
+                let end = front.len() - (len - back.len());
+                let drop_front = core::ptr::from_mut(front.get_unchecked_mut(..end));
+
+                self.front = self.to_physical_index(end);
+                self.full = false;
+
+                ptr::drop_in_place(drop_front);
+            } else {
+                let drop_front = core::ptr::from_mut(front);
+                // 'end' is non-negative by the condition above
+                let end = back.len() - len;
+                let drop_back = core::ptr::from_mut(back.get_unchecked_mut(..end));
+
+                self.front = self.to_physical_index(self.storage_len() - len);
+                self.full = false;
+
+                // If `drop_front` causes a panic, the Dropper will still be called to drop it's
+                // slice during unwinding. In either case, front will always be
+                // dropped before back.
+                let _back_dropper = Dropper(&mut *drop_back);
+                ptr::drop_in_place(drop_front);
+            }
+        }
+    }
+
     /// Retains only the elements specified by the predicate.
     ///
     /// In other words, remove all elements `e` for which `f(&e)` returns false.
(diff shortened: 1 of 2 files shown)


src/deque.rs
```diff
@@ -954,6 +954,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+    /// ```
+    #[doc(alias = "truncate_front")]
+    pub fn retain_back(&mut self, len: usize) {
+        /// Runs the destructor for all items in the slice when it gets dropped (gracefully or
+        /// during unwinding).
+        struct Dropper<'a, T>(&'a mut [T]);
+
+        impl<'a, T> Drop for Dropper<'a, T> {
+            fn drop(&mut self) {
+                unsafe {
```
zeenix: safety comment please.

src/deque.rs
```diff
@@ -954,6 +954,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+            }
+        }
+
+        // Safety:
+        // * Any slice passed to `drop_in_place` is valid; the second case has `len <= back.len()`
+        //   and returning on `len > self.storage_len()` ensures `end <= front.len()` in the first
+        //   case
+        // * Deque front/back cursors are moved before calling `drop_in_place`, so no value is
+        //   dropped twice if `drop_in_place` panics
+        unsafe {
```
zeenix: Would be good to minimize the scope of `unsafe` here. This would mean multiple `unsafe` blocks but that's good. I believe there is also a clippy lint for this.

src/deque.rs
```diff
@@ -954,6 +954,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+                }
+            }
+        }
+
+        // Safety:
+        // * Any slice passed to `drop_in_place` is valid; the second case has `len <= back.len()`
+        //   and returning on `len > self.storage_len()` ensures `end <= front.len()` in the first
+        //   case
+        // * Deque front/back cursors are moved before calling `drop_in_place`, so no value is
+        //   dropped twice if `drop_in_place` panics
```
zeenix: nitpick: Missing `.` at the end of each sentence. Same goes for multiple other comments here.

src/deque.rs
```diff
@@ -954,6 +954,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+        unsafe {
+            // If new desired length is greater or equal, we don't need to act.
+            if len >= self.storage_len() {
+                return;
+            }
+
+            let (front, back) = self.as_mut_slices();
+
+            if len > back.len() {
+                let end = front.len() - (len - back.len());
```
zeenix: A comment under both the `if` and `else` blocks at the start, to briefly describe what needs to be done, would be very helpful for readability.

src/deque.rs
```diff
@@ -954,6 +954,78 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
...
+    /// buf.push_back(10);
+    /// buf.push_back(15);
+    /// buf.retain_back(1);
+    /// assert_eq!(buf.make_contiguous(), [15]);
+    /// ```
+    #[doc(alias = "truncate_front")]
+    pub fn retain_back(&mut self, len: usize) {
+        /// Runs the destructor for all items in the slice when it gets dropped (gracefully or
+        /// during unwinding).
+        struct Dropper<'a, T>(&'a mut [T]);
```
zeenix: Since this is only used in the `else` block below, I think it should be defined here. Feel free to split the impl of both branches into separate methods, if you like.

zeenix (changes requested):
Thanks! As usual, I only have nitpicks to offer. :)

nullstalgia:
@zeenix Since this PR is essentially a continuation of the implementation I did for `truncate`, am I to apply these same suggested changes to that method as well?

The Safety comments (and overall implementation) I had there and here are essentially copies from the ones from the stdlib's implementation of [`VecDeque::truncate`](https://doc.rust-lang.org/1.96.0/src/alloc/collections/vec_deque/mod.rs.html#1370) and ~~`truncate_front`~~ [`VecDeque::retain_back`](https://doc.rust-lang.org/1.96.0/src/alloc/collections/vec_deque/mod.rs.html#1436). Otherwise there would be unexplained differences in the implementations despite how logically similar they are both between each other here and the stdlib's. Would that fall into the scope of a different PR?

src/deque.rs
nullstalgia: > I believe there is also a clippy lint for this.

As in, there is a clippy lint being thrown? Or there is an optional lint that could be enabled, since `cargo clippy` doesn't throw anything for me at the moment.

zeenix: > As in, there is a clippy lint being thrown? Or there is an optional lint that could be enabled,

If it's not trigering then it's optional. :)

zeenix:
> @zeenix Since this PR is essentially a continuation of the implementation I did for `truncate`, am I to apply these same suggested changes to that method as well?

Yes please. It's good to be always consistent but please in a separate PR than this.

> The Safety comments (and overall implementation) I had there and here are essentially copies from the ones from the stdlib's implementation of [`VecDeque::truncate`](https://doc.rust-lang.org/1.96.0/src/alloc/collections/vec_deque/mod.rs.html#1370) and ~`truncate_front`~ [`VecDeque::retain_back`](https://doc.rust-lang.org/1.96.0/src/alloc/collections/vec_deque/mod.rs.html#1436). Otherwise there would be unexplained differences in the implementations despite how logically similar they are both between each other here and the stdlib's.

The important bit is the logic and actual code. As long as they remain as similar as possible, I don't think deviating for improvement, is a problem.

> Would that fall into the scope of a different PR?

This would be fixup of this umerged work, so I don't think so.

**EDIT:** I might have misunderstood you here. I mean, the fixups/changes to the unmerged changes in this PR, should go in this PR. Changes to existing code (already merged) should go into a separate one. I hope it's clear now. :)

nullstalgia:
Thanks for clarifying, I'll push the requested changes + submit an accompanying PR soon.

src/deque.rs
```diff
@@ -2243,6 +2331,248 @@ mod tests {
...
+        const LEN: usize = 20;
+        const TRUNC: usize = 3;
+        for push_front_amt in 0..=LEN {
+            let mut tester: Deque<_, LEN> = Deque::new();
+            for index in 0..LEN {
+                if index < push_front_amt {
+                    assert!(
+                        tester.push_front(Droppable::new()).is_ok(),
+                        "deque must have room for all {LEN} entries"
+                    );
```
zeenix, line 2554: Why not simply:

```rust
tester
    .push_front(Droppable::new())
    .expect("deque must have room for all {LEN} entries");
```

src/deque.rs
```diff
@@ -2243,6 +2331,248 @@ mod tests {
...
+                        tester.push_back(Droppable::new()).is_ok(),
+                        "deque must have room for all {LEN} entries"
+                    );
+                }
+            }
+
+            assert_eq!(Droppable::count(), LEN as i32);
+
+            let (front, back) = tester.as_slices();
+            println!("A: {:?} {:?}\n", front, back);
```
zeenix: I don't think we should have prints in tests.

zeenix (commented):
lgtm otherwise.

zeenix (changes requested):

src/deque.rs
nullstalgia: `.expect()` does not support formatting strings as it is not a macro.

Would you prefer that I use `.expect(format!())`, or simply not mention the `LEN` at all?

zeenix: I thought so too, that's why I checked before commenting: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f9cecd4084d473ebc5a41dab7b8b1c21

nullstalgia: I might be missing something, but your example isn't very clear/is conflicting.

It's `expect()`-ing the `Ok(())` variant (so no error message gets printed due to a failed unwrap), and the `#![allow(unused)]` hides the warning that `LEN` is never read.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=06510eeb3a4c30ff16264f72f5f914f2

zeenix: Right, nm then. :)

zeenix (changes requested):
ok, time for merge then but before that, please rebase on master/main (instead of merging that) to keep the history easier to follow (and avoiding unrelated commits in the PR).

nullstalgia:
Alrighty, should be good to go! Thanks for helping see this PR through, hopefully they (`t-libs`) don't change the method name on us again! ;P

zeenix (approved):

PR #664, opened 2026-06-09
Modify Deque::truncate to closer match the new Deque::retain_back implementation 

nullstalgia:
As requested by @zeenix in #663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 - Added `resize_with` to `Vec`
-- Modify `Deque::truncate` to closer match the new `Deque::retain_back` implementation
 
 ## [v0.9.3] 2025-04-15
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 - Added `resize_with` to `Vec`
+- Modify `Deque::truncate` to closer match the new `Deque::retain_back` implementation
```
zeenix: This is internal details that doesn't need to be in the release notes.

zeenix (changes requested):
Thanks so much. LGTM otherwise.

CHANGELOG.md
nullstalgia: I don't disagree, but I assumed the changelog CI step would've complained like it has for me in the past. But I guess that's on me for assuming. I'll remove it and see if it throws an error.

Edit: Ah, I see you added the `skip-changelog` label, noted.

zeenix (approved):

PR #665, opened 2026-06-09
vec: add `extract_if()`

max-heller:
Adapts `alloc::vec::Vec::extract_if()` to implement `heapless::Vec::extract_if()`

I noticed that `vec::Drain` stores `LenT` instead of `usize` for its internal indices. Should we do the same here to save a few bytes? I wasn't sure if it's worth it relative to the (also small) cost of casting back and forth between `LenT` and `usize`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
+- Added `Vec::extract_if`.
 
 ## [v0.9.3] 2025-04-15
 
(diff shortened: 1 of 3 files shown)


hiSandog:
Worth using `LenT` here for consistency — the existing `vec::Drain` in the crate already stores its indices as `LenT`, and since `heapless::Vec`'s own length is bounded by `LenT` anyway, the retained/removed range indices can never exceed `LenT::MAX`, so the narrowing is always safe. The `LenT as usize` casts on indexing aren't a perf concern: `extract_if`'s indices drive slice bounds and `drop_in_place` bookkeeping, not a hot arithmetic loop, and on most targets the cast compiles to nothing. Main thing to watch is keeping the kept-range start/end and the read/write cursors all on `LenT` (not mixed), so you don't end up casting in two directions. Keeping the two drain-like types (`Drain` and `ExtractIf`) on the same index representation is a nice consistency win too.

Issue #666, opened 2026-07-07
[Feature Request] Add `BinaryHeap::retain`


Conaclos:
While I was using `heapless::BinaryHeap`, I found that it lacked a `retain` method similar to [`std::BinaryHeap::retain`](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.retain).
Would you be open to the addition of such a method?

zeenix:
there is already a PR for this: https://github.com/rust-embedded/heapless/pull/345

PR #345, opened 2023-02-23
feat: support retain operation over binary heap

pwnorbitals:
Might be useful :)

PR #668, opened 2026-07-08
Add `BinaryHeap::retain`

Conaclos:
Implement https://github.com/rust-embedded/heapless/issues/666

This PR adds the `BinaryHeap::retain` method.
I didn't base my work on https://github.com/rust-embedded/heapless/pull/345 because the implementation seems wrong to me.

I added an internal method `BinaryHeap::remove` that is now used by both `BinaryHeap::retain` and `BinaryHeap::pop_unchecked`.
In the process I improved its implementation.

One alternative implementation could be to traverse the internal vector in reverse order. This could reduce the number of shifted elements if `retain` removes multiple elements. However, this would make the implementation much more complex.


zeenix:
> I didn't base my work on #345 because the implementation seems wrong to me.

How so?

Conaclos:
> How so?

The implementation doesn't preserve the order invariant of the binary heap.
For example, if you take the binary max-heap `[8, 2, 5]` and ask to retain only the numbers that are different of `8`, then you end up with `[2, 5]`, which is incorrect.

zeenix:
> For example, if you take the binary max-heap `[8, 2, 5]` and ask to retain only the numbers that are different of `8`, then you end up with `[2, 5]`, which is incorrect.

Huh? Is it? :thinking: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=98dfdde25b2a102d7146761f6d5294db

Conaclos:
I will add some tests for `sift_down_to_bottom` because the function seems untested with arguments (`pos`) greater than `0`.

Conaclos:
> Huh? Is it? 🤔 https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=98dfdde25b2a102d7146761f6d5294db

Your example is wrong because you called `into_sorted_vec()` in the assertion. It will always return `[2, 5]`.
Instead, try: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dccba979ca3526f18a2a2dea5f93130f

zeenix:
> Your example is wrong because you called `into_sorted_vec()` in the assertion. It will always return `[2, 5]`.
> Instead, try: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dccba979ca3526f18a2a2dea5f93130f

Ah right. That was from the std doc example I started off from.

Conaclos:
Ready for review!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
+- Added `retain` to `BinaryHeap`
```
zeenix: missing `.`. I know the previous entries did the same but let's not perpetuate this. :) Feel free to add to them too.

src/binary_heap.rs
```diff
@@ -471,9 +502,28 @@ where
     }
 
     /* Private API */
+
+    /// Removes and returns the element at position `index` within the inner vec.
+    /// The elements are shifted to preserve the invariants of the binary heap.
```
zeenix, line 508: Let's follow the Rust convention: title + empty line and then explanation etc in the next paragraph:

```suggestion
    ///
    /// The elements are shifted to preserve the invariants of the binary heap.
```

src/binary_heap.rs
```diff
@@ -415,13 +415,44 @@ where
...
-        let mut item = self.data.pop_unchecked();
+        // SAFETY: the binary heap is not empty, thus `0` is smaller than `self.len()`.
+        unsafe { self.remove_unchecked(0) }
+    }
 
-        if !self.is_empty() {
-            mem::swap(&mut item, self.data.as_mut_slice().get_unchecked_mut(0));
-            self.sift_down_to_bottom(0);
+    /// Retains only the elements specified by the predicate.
+    /// The elements are visited in arbitrary order.
```
zeenix, line 421:
```suggestion
    ///
    /// The elements are visited in arbitrary order.
```

zeenix (changes requested):
Only tiny nits and then it's good from my side.

zeenix (dismissed):

Conaclos:
@zeenix Is there any concerns preventing merging this feature?

zeenix:
> @zeenix Is there any concerns preventing merging this feature?

Not from my side but I was hoping for another review from someone else.

src/binary_heap.rs
```diff
@@ -471,9 +503,29 @@ where
...
+        }
+        item
+    }
+
+    /// Moves the element from `pos` down through the heap until it becomes a leaf,
+    /// then sift up the element to meet the order invariant.
     fn sift_down_to_bottom(&mut self, mut pos: usize) {
         let end = self.len();
         let start = pos;
+        // Moves the element down to a leaf.
```
sgued: Please make this a `// SAFETY:` comment, that way we can progressively work towards enabling `#![deny(clippy::undocumented_unsafe_blocks)]`

src/binary_heap.rs
```diff
@@ -890,6 +945,192 @@ mod tests {
...
+        assert_eq!(heap.pop(), Some(17));
+        assert_eq!(heap.pop(), Some(7));
+        assert_eq!(heap.pop(), Some(3));
+        assert_eq!(heap.pop(), Some(1));
+        assert_eq!(heap.pop(), None);
+    }
+
+    #[test]
+    fn remove_unchecked() {
+        // This test depends on implementation details.
```
sgued:
```suggestion
        // This test depends on private APIs.
```

I spent some times figuring out which implementation detail regarding order or something was impactful, I think "private APIs" is clearer and is already used in another comment.

sgued (commented):
LGTM besides two small nits.

Thank you!

sgued (changes requested):

Conaclos:
@sgued Fixed.

I took the opportunity of making unsafe the relevant functions, adding missing unsafe blocks and SAFETY comments. This improves compatibility with `clippy::undocumented_unsafe_blocks` and `clippy::multiple_unsafe_ops_per_block`.

I added `PeekMutInner::new` that allows checking the type invariant.
Ideally, this type should be moved to a dedicated module to protect its instantiation.
I renamed `PeekMutInner::sift` into `PeekMutInner::removed` (inverting the boo logic) to make clearer its purpose.

Last, I turned some conditionals into early returns to make them cold paths.

sgued (dismissed):

sgued (approved):

sgued:
I rebased it and force pushed a typo fix and CI fixes

zeenix (approved):

PR #669, opened 2026-07-08
Support rustup

Conaclos:
Add `rust-toolchain.toml` that allows users to install the required Rust version using `rustup`.


zeenix:
don't think it's a good idea.

zeenix:
we don't require a specific Rust version. It can be any Rust version after our MSRV and MSRV is declared in the Cargo.toml.

PR #670, opened 2026-07-16
Add Vec::push_mut

mdrssv:
Mirrors: [std::vec::Vec::push_mut](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push_mut) into heapless

### DIFF
--- a/src/vec/mod.rs
+++ b/src/vec/mod.rs
@@ -687,13 +687,24 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
         }
     }
 
-    /// Appends an `item` to the back of the collection
+    /// Appends an `item` to the back of the collection.
     ///
     /// Returns back the `item` if the vector is full.
     pub fn push(&mut self, item: T) -> Result<(), T> {
-        if self.len() < self.capacity() {
-            unsafe { self.push_unchecked(item) }
-            Ok(())
+        self.push_mut(item).map(|_| ())
+    }
+
+    /// Appends an `item` to the back of the collection.
+    ///
+    /// Returns back an mutable reference to the `item` or the `item` if the vector is full.
+    pub fn push_mut(&mut self, item: T) -> Result<&mut T, T> {
+        let len = self.len();
+        if len < self.capacity() {
+            // Safety: len has been checked to be less than capacity, enough for one more item.
+            unsafe { self.push_unchecked(item) };
+            // Safety: len which was obtained before insertion now corresponds to self.len() - 1
+            // representing the index for the just added item.
+            Ok(unsafe { self.get_unchecked_mut(len) })
         } else {
             Err(item)
         }
@@ -2571,5 +2582,14 @@ mod tests {
     }
     fn _test_variance_view<'a: 'b, 'b, 'c>(x: &'c VecView<&'a ()>) -> &'c VecView<&'b ()> {
         x
+    }
+
+    #[test]
+    fn push_mut() {
+        let mut vec = Vec::<u32, 16>::new();
+        *vec.push_mut(3).unwrap() += 2;
+        vec.push_mut(4).unwrap();
+        *vec.push_mut(5).unwrap() += 10;
+        assert_eq!(&vec[..], &[5, 4, 15]);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
 - Fixed unsoundness in `Vec::IntoIter::drop` and `HistoryBuf::write`in the context of panicking drop implementations.
+- Added `push_mut` to `Vec`.
 
 ## [v0.9.3] 2025-04-15
 


src/vec/mod.rs
```diff
@@ -691,9 +691,22 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
...
-            unsafe { self.push_unchecked(item) }
-            Ok(())
+            unsafe {
+                self.push_unchecked(item);
+                Ok(self
+                    .buffer
+                    .borrow_mut()
+                    .get_unchecked_mut(self.len.into_usize() - 1)
+                    .assume_init_mut())
+            }
```
sgued: I think this is a bit more readable:

```suggestion
    pub fn push_mut(&mut self, item: T) -> Result<&mut T, T> {
        let len = self.len();
        if len < self.capacity() {
            unsafe {
                self.push_unchecked(item);
                Ok(self.get_unchecked_mut(len))
            }
        } else {
            Err(item)
        }
    }
```

src/vec/mod.rs
```diff
@@ -2572,4 +2585,15 @@ mod tests {
...
     }
+
+    #[test]
+    fn push_mut() {
+        let mut vec = Vec::<u32, 16>::new();
+        vec.push_mut(3).unwrap();
+        vec.push_mut(4).unwrap();
+        let item_ref = vec.push_mut(5).unwrap();
+        dbg!(*item_ref);
+        *item_ref = *item_ref + 10;
```
sgued:
```suggestion
        *item_ref += 10;
```

sgued (changes requested):
Thanks for the contribution! This looks useful.

Some of the clippy lints failure are caused by existing code and the update to 1.97 and will be fixed in https://github.com/rust-embedded/heapless/pull/671, but there are still a couple of them that are in your code.

sgued (changes requested):
Thanks for the contribution! This looks useful.

Some of the clippy lints failure are caused by existing code and the update to 1.97 and have been fixed in https://github.com/rust-embedded/heapless/pull/671 so please rebase on top of master. There are still a couple of lints that affect your code though.

src/vec/mod.rs
zeenix: @sgued That's good suggestion, thanks. @mdrssv Also it would be great if you could:

* split the `unsafe` blocks to keep their scopes as small as possible. So `unsafe { self.push_unchecked(item); }` and then `Ok(unsafe { self.get_unchecked_mut(len) })` (still on separate lines, of course).
* Add `Safety` comments on each `unsafe` blocks, describing how each invariant is holding.

zeenix (changes requested):
I left a comment with 2 changes requested.

mdrssv:
Thanks for the feedback

src/vec/mod.rs
```diff
@@ -691,9 +691,20 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
     ///
     /// Returns back the `item` if the vector is full.
     pub fn push(&mut self, item: T) -> Result<(), T> {
-        if self.len() < self.capacity() {
-            unsafe { self.push_unchecked(item) }
-            Ok(())
+        self.push_mut(item).map(|_| ())
+    }
+
+    /// Appends an `item` to the back of the collection
```
zeenix: nitpick:


```suggestion
    /// Appends an `item` to the back of the collection.
```

Same for all comments and the changelong entry. I know some folks have been doing a sloppy job before you but let's not perpetuate human slop. :)

src/vec/mod.rs
```diff
@@ -687,13 +687,24 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
...
+    /// Appends an `item` to the back of the collection.
+    ///
+    /// Returns back an mutable reference to the `item` or the `item` if the vector is full.
+    pub fn push_mut(&mut self, item: T) -> Result<&mut T, T> {
+        let len = self.len();
+        if len < self.capacity() {
+            // Safety: len has been checked to be less then capacity, enough for one more item
+            unsafe { self.push_unchecked(item) };
+            // Safety: len which was obtained before insertion now corresponds to self.len() - 1
+            // representing the index for the just added item
```
zeenix: Comments here are still missing their `.`. :)

zeenix: A final thouch and we can merge it, thanks!

src/vec/mod.rs
```diff
@@ -687,13 +687,24 @@ impl<T, LenT: LenType, S: VecStorage<T> + ?Sized> VecInner<T, LenT, S> {
...
+        self.push_mut(item).map(|_| ())
+    }
+
+    /// Appends an `item` to the back of the collection.
+    ///
+    /// Returns back an mutable reference to the `item` or the `item` if the vector is full.
+    pub fn push_mut(&mut self, item: T) -> Result<&mut T, T> {
+        let len = self.len();
+        if len < self.capacity() {
+            // Safety: len has been checked to be less then capacity, enough for one more item
```
sgued: A little typo
```suggestion
            // Safety: len has been checked to be less than capacity, enough for one more item
```

sgued (approved):

zeenix (approved):

PR #671, opened 2026-07-16
Fix clippy lints updates from 1.97

sgued:


zeenix (changes requested):
Could you please add a bit of details to the commit log **and** PR description (they could be the same in this case). Just copy&pasting the errors/warnings from clippy would more than suffice.

In general, could you please get in the habbit of doing this. I keep asking for each PR separately. :) Besides other reasons, it makes reviewing much easier.

sgued:
I updated the commit message and PR description

zeenix (approved):
Thanks!

PR #675, opened 2026-07-24
IndexMap: rebuild indices after truncate

### LINKED ISSUE #674: IndexMap::truncate leaves the map in an inconsistent state.
Hi,

I am on the Cargo Team and was recently hired by the Rust Foundation to help triage AI discovered bugs. This bug was found with AI, namely [scrutineer](https://github.com/alpha-omega-security/scrutineer) and Fable. Except where specifically labeled this report was written by me.

This is similar to #559 but can be reproduced after #560.

```rust
    let mut map: FnvIndexMap<u16, u32, 4> = FnvIndexMap::new();
    map.insert(4, 444).unwrap();
    map.insert(2, 222).unwrap();
    map.insert(0, 0).unwrap();
    map.insert(7, 7).unwrap();

    assert_eq!(map.get(&2), Some(&222)); // ok before map.truncate(2);
    map.truncate(2);
    assert_eq!(map.iter().collect::<Vec<_>>(), vec![(&4, &444), (&2, &222)]); // ok
    assert_eq!(map.get(&4), Some(&444)); // ok
    assert_eq!(map.get(&2), Some(&222)); // <-- key present in iter() but unreachable
```
Panics on the last line.

The AI explains what happens as:

> truncate() calls entries.truncate(len) then iterates indices and sets to None only those Pos whose index()>=len. Surviving entries (index<len) keep their Pos, but any probe chain that passed THROUGH a now-nulled slot is broken: open-addressing find() stops at the first None slot (index_map.rs:190 let pos = self.indices[probe]?;). A surviving key whose desired_pos precedes a removed key's slot, and which was displaced past it, becomes unreachable. No backward-shift / rehash is performed, unlike remove_found and retain_in_order which correctly rebuild.

sgued:
This missing operation after truncate left the map in an inconsistent state The logic is kept the same as for `retain`
Closes https://github.com/rust-embedded/heapless/issues/674

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Fixed `IndexMap::truncate` leading to an inconsistent state.
 - Added `resize_with` to `Vec`
 - Added `retain_back` (aka `truncate_front`) to `Deque`
 - Fixed unsoundness in `Vec::IntoIter::drop` and `HistoryBuf::write`in the context of panicking drop implementations.
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -311,11 +311,14 @@ where
     where
         F: FnMut(&mut K, &mut V) -> bool,
     {
-        const INIT: Option<Pos> = None;
-
         self.entries
             .retain_mut(|entry| keep(&mut entry.key, &mut entry.value));
 
+        self.reinsert_all();
+    }
+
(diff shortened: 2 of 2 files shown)


sgued:
Note that this will conflict with #673

zeenix:
> Note that this will conflict with #673

Oops, sorry I had missed that one. Conflict as it, one of these will need to be rebased once one for them is merged?

sgued:
Yes I mean just git conflict. It should be straightforward to fix but will require a re-review after that.

zeenix (approved):
LGTM.

PR #677, opened 2026-07-24
Deque::make_contiguous: don't let `back` become equal to `N`

### LINKED ISSUE #676: `make_contiguous` on a full `Deque` can lead to an out-of-bounds write
The following program aborts on a debug build and writes out of bounds on a release build:

```rust
use heapless::Deque;

let mut q: Deque<i32, 4> = Deque::new();
for i in 0..4 { q.push_back(i).unwrap(); }
for i in 4..7 { q.pop_front(); q.push_back(i).unwrap(); }
q.make_contiguous();
q.pop_front();
q.push_back(99).unwrap();
```

On debug it aborts with:

```
thread 'main' panicked at src/deque.rs:672:35:
unsafe precondition(s) violated: slice::get_unchecked_mut requires that the index is within the slice
thread caused non-unwinding panic. aborting.
```

On release there is no error; the final `push_back` writes past the end of the backing array.

The deque is full (4/4) and wrapped when `make_contiguous` is called. After that call, `is_full()` returns `false` even though the deque still holds 4 elements:

```rust
use heapless::Deque;

let mut q: Deque<i32, 4> = Deque::new();
for i in 0..4 { q.push_back(i).unwrap(); }
for i in 4..7 { q.pop_front(); q.push_back(i).unwrap(); }
q.make_contiguous();
q.pop_front();
q.push_front(3).unwrap();
assert_eq!(q.len(), 4);
assert!(q.is_full()); // fails: returns false
```

The `push_back` in the first example goes ahead only because `is_full()` wrongly reports `false`.

Tested on current `main` and on 0.9.3.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

sgued:
In one branch of `make_contiguous`, `back` was given the value of `len`, which is incorrect and should be `0` when the `Deque` is full

Closes #676

### DIFF
--- a/src/deque.rs
+++ b/src/deque.rs
@@ -504,7 +504,7 @@ impl<T, S: VecStorage<T> + ?Sized> DequeInner<T, S> {
                     // the used part of the buffer now is `0..self.len`, so set
                     // `head` to the beginning of that range.
                     self.front = 0;
-                    self.back = len;
+                    self.back = if free == 0 { 0 } else { len };
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/deque.rs
```diff
@@ -1807,6 +1807,20 @@ mod tests {
 
         // Deque contains: 5, 6, 7, 8
         assert_eq!(q.as_slices(), ([5, 6, 7, 8].as_slice(), [].as_slice()));
+
+        // Make contiguous failing to reset back index properly.
```
zeenix: I read this many times and the sentence didn't make sense to me and then I realized, you meant:


```suggestion
        // Test `make_contiguous` failing to reset back index properly.
```

zeenix (changes requested):
lgtm otherwise.

zeenix:
> Closes #676

Please also add this to the commit message. :pray:

zeenix (approved):
Nice!

PR #680, opened 2026-07-28
IndexMap and IndexSet: mark `remove` as deprecated

### LINKED ISSUE #679: IndexMap/IndexSet: deprecate `remove()` in favour of `swap_remove`
`.remove` does not convey the current behaviour that it changes the insertion order of the values in the map/set. [indexmap](https://docs.rs/indexmap) deprecated the method in favour of the more explicit `swap_remove`.

We should add `swap_remove` where it's missing and mark `remove` as deprecated, following the `IndexMap` API.

Since I anticipate there could be discussion on whether this is the best path forward, I did not include this change in https://github.com/rust-embedded/heapless/pull/673.

sgued:
`remove()` changed the ordering of the map and this was not made clear in either the documentation or the name, this patch follows the original indexmap crate in using renaming it `swap_remove` instead

Closes https://github.com/rust-embedded/heapless/issues/679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 - Added `swap_remove()` to `IndexMap` and `IndexSet`.
+- Deprecated `.remove()` in `IndexMap` and `IndexSet` in favour of `.swap_remove()`.
 - Fixed `IndexMap::truncate` leading to an inconsistent state.
 - Fixed `Deque::make_contigous` leading to an inconsistent state.
 - Added `resize_with` to `Vec`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Added `swap_remove()` to `IndexMap` and `IndexSet`.
 - Fixed `IndexMap::truncate` leading to an inconsistent state.
 - Fixed `Deque::make_contigous` leading to an inconsistent state.
 - Added `resize_with` to `Vec`
--- a/src/index_map.rs
+++ b/src/index_map.rs
@@ -693,8 +693,12 @@ where
 
     /// Removes this entry from the map and yields its corresponding key and value.
     ///
-    /// **NOTE**: This is equivalent to `.swap_remove_entry()`, replacing this entry’s position
-    /// with the last element.
+    /// **NOTE**: This is equivalent to [`.swap_remove_entry()`](OccupiedEntry::swap_remove_entry),
+    /// replacing this entry’s position with the last element, and it is deprecated in favor of
+    /// calling that explicitly.
+    #[deprecated(
+        note = "`remove_entry` disrupts the map order -- use `swap_remove_entry` for explicit behavior."
+    )]
     pub fn remove_entry(self) -> (K, V) {
         self.swap_remove_entry()
     }
@@ -744,8 +748,12 @@ where
 
     /// Removes this entry from the map and yields its value.
     ///
-    /// **NOTE**: This is equivalent to `.swap_remove()`, replacing this entry’s position
-    /// with the last element.
+    /// **NOTE**: This is equivalent to [`.swap_remove()`](OccupiedEntry::swap_remove), replacing
+    /// this entry’s position with the last element, and it is deprecated in favor of calling
+    /// that explicitly.
+    #[deprecated(
+        note = "`remove` disrupts the map order -- use `swap_remove` for explicit behavior."
+    )]
     pub fn remove(self) -> V {
         self.swap_remove()
     }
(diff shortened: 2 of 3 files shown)


src/index_set.rs
```diff
@@ -1,4 +1,6 @@
+#![deny(clippy::undocumented_unsafe_blocks)]
```
zeenix, line 1: This is good but why don't we just have it on the crate-level? I'm guessing it's because it'll be a lot of work adding comments on all the unsafe blocks?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Deprecated `.remove()` in `IndexMap` and `IndexSet` in favour of `.swap_remove()`.
```
zeenix, line 11: I'd split the method additons and deprecation points here. Otherwise, it sounds like you're only deprecating in favor of existing methods.

src/index_map.rs
```diff
@@ -732,8 +746,22 @@ where
     }
 
     /// Removes this entry from the map and yields its value
+    ///
+    /// **NOTE**: This is equivalent to `.swap_remove()`, replacing this entry’s position
+    /// with the last element, and it is deprecated in favor of calling that explicitly.
+    #[deprecated(
+        note = "`remove` disrupts the map order -- use `swap_remove` for explicit behavior."
+    )]
```
zeenix, line 756: Could you please split the deprecation into a separate commit? If we decideded to undeprecate for any reason (could happen), it'll be an easier revert.

zeenix (changes requested):
LGTM otherwise.

src/index_set.rs
sgued: Making it crate-wide would need a total of 86 more comments. I don't think this should be part of this PR. 

I think this should be our end-goal though, so I think making sure that the parts of the crate that have well-documented `unsafe` don't start introducing some undocumented unsafe is good.

I've also started documenting all unsafe in a module whenever I do a significant refactor of said module, like in #673, when it's less effort to do so because I have the logic of the module in mind.

CHANGELOG.md
sgued: Good point, done

src/index_map.rs
sgued: Done

zeenix: I see 4 commits and your push wasn't forced so I don't think you ended up doing what I asked. :) Splitting commits requires an interactive rebase and force push.

src/index_map.rs
```diff
@@ -692,7 +692,21 @@ where
     }
 
     /// Removes this entry from the map and yields its corresponding key and value
+    ///
+    /// **NOTE**: This is equivalent to `.swap_remove_entry()`, replacing this entry’s position
+    /// with the last element, and it is deprecated in favor of calling that explicitly.
+    #[deprecated(
+        note = "`remove` disrupts the map order -- use `swap_remove` for explicit behavior."
```
reitermarkus:
```suggestion
        note = "`remove_entry` disrupts the map order -- use `swap_remove_entry` for explicit behavior."
```

src/index_map.rs
```diff
@@ -1335,6 +1363,12 @@ where
     /// assert_eq!(map.remove(&1), Some("a"));
     /// assert_eq!(map.remove(&1), None);
```
reitermarkus, line 1366: Examples should be moved to the non-deprecated methods.

reitermarkus: Ah, they are already there. I'd still remove them from the deprecated methods though.

sgued: Done

src/index_map.rs
sgued: Done

src/index_map.rs
```diff
@@ -692,7 +692,18 @@ where
     }
 
     /// Removes this entry from the map and yields its corresponding key and value
+    ///
+    /// **NOTE**: This is equivalent to `.swap_remove_entry()`, replacing this entry’s position
+    /// with the last element
```
zeenix: I noticed the lack of `.` at the end of many sentences in the comments and commit log's detail section (No need for them in the title, as they're titles). Could you please remember to do this. Otherwise, it looks AI-generated slop. 😆

zeenix (changes requested):
All good otherwise IMO. 👍

src/index_map.rs
sgued: > Otherwise, it looks AI-generated slop. 😆

That is both wrong and not funny.

I honestly don't care about `.` at the end of sentences of comments. `rg '// .*[a-zA-Z]$'` shows they're missing for a lot of comments in the crate. I feel like this is just wasting time generating back-and-forth in almost every PR (not just mine).

zeenix: > That is both wrong and not funny.

I'm deciding to find it funny because I don't want be annoyed at the fact that I have to keep asking you to do this very basic thing all the time while OTOH I had the exact some problem with LLM but I only had to tell those things once not to do this and they stopped.

> I honestly don't care about `.`

This is not your personal project so that's not at all argument to be sloppy here. 
I'm not asking you to do something that's **just** my personal preference. I'm asking you to follow the well-established English grammer rules where they apply in this project. I believe comments, git logs and PR descriptions etc fall under this category.

zeenix: > I feel like this is just wasting time generating back-and-forth in almost every PR (not just mine).

This is about quality control. I know people love to be sloppy but I don't see why I should be linient about it in my reviews.

zeenix (approved):

PR #686, opened 2026-08-24
IndexMap: add insert_full and Index<usize>

### LINKED ISSUE #325: Add usize index methods to IndexMap
For my usecase, I need to be able to refer to `IndexMap` keys by `usize` index as well as by their keys. Specifically, I would like access to the [`.insert_full(k, v)`](https://docs.rs/indexmap/1.9.2/indexmap/map/struct.IndexMap.html#method.insert_full) method and the [`Index<usize>`](https://docs.rs/indexmap/1.9.2/indexmap/map/struct.IndexMap.html#impl-Index%3Cusize%3E-for-IndexMap%3CK%2C%20V%2C%20S%3E) trait implemented from the upstream non-heapless `IndexMap`.  Otherwise I'm better off using a plain `Vec` and doing linear searches, so that I can use both keys and fast `usize` indexing.

One alternative to the above would be `Entry`'s [`.index()`](https://docs.rs/indexmap/1.9.2/indexmap/map/enum.Entry.html#method.index) and `IndexMap`'s [`.get_index(i)`](https://docs.rs/indexmap/1.9.2/indexmap/map/struct.IndexMap.html#method.get_index) methods, which would also allow me to save the index of a key when I insert then read from it later.

ChrisJr404:
This ports the index-oriented accessors from the upstream indexmap crate, as suggested in #325. `insert_full` behaves like `insert` but also hands back the entry's index, and `Index<usize>`/`IndexMut<usize>` let you read and write values by position with `map[i]`. Both round out the existing `get_index` family so you can refer to entries by position as well as by key.

Closes #325


sgued (commented):
Thanks, this looks pretty good.

Can you please also add a `get_full(&self, key: &Q) -> Option<(usize, &K, &V)>` similar in spirit?

PR #689, opened 2026-09-14
IndexMap::retain: Fix unsoundness in the context of panicking predicate

### LINKED ISSUE #688: `IndexMap::retain` is not exception safe, leading to read of uninit values
Hello, we are security researchers targeting Rust security. By running our tool in this repository, we found a soundness issue. **This report is written by 100% human**. We promise that all you read will never be generated by LLM except otherwise stated.

`IndexMap::retain` calls `CoreMap::retain_in_order`, which further calls `VecInner::retain_mut`:

https://github.com/rust-embedded/heapless/blob/f008da8b34aa14098e0a14a9078c57c9a7e288f6/src/index_map.rs#L406-L414

The `keep` function is a user-supplied closure, which may panic. However, when it panics, the `BackshiftOnDrop` drop guard will make the `VecInner` exception safe:

https://github.com/rust-embedded/heapless/blob/f008da8b34aa14098e0a14a9078c57c9a7e288f6/src/vec/mod.rs#L1243-L1281

This would make `self.entries` shrink to be valid by `set_len`. However, the `indices` field of `CoreMap` is not guarded, in which some elements' indices will then be greater than `self.entries.len()`. Then if we catch the panic using `std::panic::catch_unwind`, and we call `map.get()`, the `get_unchecked` will then lead to read of uninit values:

https://github.com/rust-embedded/heapless/blob/f008da8b34aa14098e0a14a9078c57c9a7e288f6/src/index_map.rs#L716-L721

To fix this problem, I think adding a drop guard similar to what `VecInner::retain_mut` does to rebuild the indices could help.

## PoC

Here is a PoC for this issue. **THIS PoC IS GENRATED BY LLM AND VERIFIED BY HUMAN.**

```rust
// PoC for a potential exception-safety violation (ESV) in heapless::IndexMap::retain.
//
// If verified with Miri, this triggers use of out-of-bounds (stale) indices after a
// panic inside the user-supplied `retain` predicate leaves the `indices` table
// pointing beyond the (shrunk) `entries` vector.

use heapless::index_map::FnvIndexMap;
use std::panic::{catch_unwind, AssertUnwindSafe};

fn main() {
    // Capacity must be a power of two > 1.
    let mut map: FnvIndexMap<u64, u8, 8> = FnvIndexMap::new();
    let keys = [10u64, 20, 30, 40, 50, 60];
    for (i, k) in keys.iter().enumerate() {
        map.insert(*k, i as u8).unwrap();
    }
    assert_eq!(map.len(), keys.len());

    // The predicate removes early elements (returning `false`) and then panics on a
    // later element. `Vec::retain_mut` runs a drop guard that backshifts and shrinks
    // `entries`, but `self.indices` is only rebuilt *after* `retain_mut` returns, so
    // on panic the indices table is left stale (pointing beyond the new length).
    let res = catch_unwind(AssertUnwindSafe(|| {
        map.retain(|k, _| {
            if *k == 10u64 {
                true
            } else if *k == 20u64 {
                false
            } else {
                panic!("user predicate panic");
            }
        });
    }));
    assert!(res.is_err(), "expected the predicate to panic");

    assert!(map.len() < keys.len(), "entries should have shrunk");

    // Querying removed keys may probe stale `indices` slots whose stored entry index
    // is now out of bounds w.r.t. the shrunk `entries`, leading to OOB get_unchecked.
    for k in keys.iter() {
        let _ = map.get(k);
    }

    println!("survived; map.len={}", map.len());
}
```

You can use miri to detect UB in this code.

sgued:
Closes https://github.com/rust-embedded/heapless/issues/688


src/index_map.rs
```diff
@@ -407,7 +407,15 @@ where
...
+            fn drop(&mut self) {
+                self.0.reinsert_all();
+            }
+        }
+        DropGuard(self)
+            .0
+            .entries
             .retain_mut(|entry| keep(&mut entry.key, &mut entry.value));
 
         self.reinsert_all();
```
zeenix: This should now be removed? 🤔

zeenix (changes requested):
LGTM otherwise.

src/index_map.rs
sgued: Right, removed it and force-pushed.

zeenix (approved):

